PR #3, opened 2019-07-28
Extract artichoke core

lopopolo:
This is an experimental branch to split apart `mruby` crate into a core and several backend implementations.

I do not intend to merge this branch, but will pull learnings from it back into `master`.

Changes so far in this branch:

- `Value` is a trait.
- `class:Spec` and `module::Spec` unified under a `ClassLike` trait object.
- Simplified `Method` definition.
- Converters are generic over interpreter.
- Added `Nil` trait so interpreters can provide a `nil` `Value`.
- Removed `Mrb` prefix from all trait definitions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,6 +40,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "artichoke-core"
+version = "0.1.0"
+dependencies = [
+ "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "artichoke-frontend"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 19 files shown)


lopopolo:
This latest commit extracts all of the behavior-defining traits and converter
traits from artichoke-backend (currently implemented with mruby) to a
base crate.

The intent is to move `extn` into artichoke-core such that core and
stdlib are generically implemented for any Artichoke interpreter that
implements the core traits.

This commit is a start, but there are still some outstanding items:

- RustBackedValue is not yet pulled into core.
- Args extraction and argspecs are not pulled into core.
- Need to figure out how to handle arg parsing for native Rust
  extensions
- Need to figure out how to handle gluing native Rust extensions to
  mruby interpreter with C function pointers.

Issue #50, opened 2019-07-30
Make the playground example require something


lopopolo:
Make the example a little more interesting. This will likely require `Regexp` (see https://github.com/artichoke/artichoke/issues/48).

```ruby
require 'json'

class A
  def bar(x)
    { bar: x }.to_json
  end
end

10.times do |i|
    puts A.new.bar(i)
end
```

PR #51, opened 2019-07-30
Make the playground example more interesting

lopopolo:
Fixes GH-50.

Make a wrapper around Set using Forwardable. Requires two packages from
stdlib.

### DIFF
--- a/artichoke-wasm/src/playground.js
+++ b/artichoke-wasm/src/playground.js
@@ -6,21 +6,38 @@ import "artichoke-wasm/artichoke_wasm.wasm";
 import "artichoke-wasm/deps/artichoke-wasm.js";
 
 const sample = `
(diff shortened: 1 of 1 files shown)


lopopolo:
```ruby
# https://ruby-doc.org/stdlib-2.5.1/libdoc/forwardable/rdoc/Forwardable.html
require 'forwardable'
# https://ruby-doc.org/stdlib-2.6.3/libdoc/set/rdoc/Set.html
require 'set'

class Registry
  extend Forwardable
  def_delegators :@records, :add, :to_a

  def initialize
    @records = Set.new
  end
end

registry = Registry.new

3.times do
  10.times do |record|
    registry.add(record)
  end
end

puts registry.to_a
registry.to_a
```

PR #59, opened 2019-07-30
Limit total memory in Wasm build to 5MiB

lopopolo:
Fix a crash in Mobile Safari caused by an out of memory error.


lopopolo:
This didn't fix the crash. Mobile Safari still crashes.

lopopolo:
https://github.com/artichoke/artichoke/issues/60

PR #61, opened 2019-07-31
Vendor and fork path_abs::PathAbs

lopopolo:
path_abs does not compile under wasm32-unknown-unknown because it
includes conditional compilation on OS vendor targeting unix and
windows. Artichoke does not make use of this functionality.

path_abs canonicalizes relative paths using env::current_dir. Artichoke
uses a virtual filesystem so this behavior is never what we want.

This commit removes the dependency on path_abs and replaces it with a
vendored copy of the PathAbs implementation. This fork modifies the
public API of PathAbs::new to take the "canonicalized" current working
directory to use for absolutizing relative paths.

Tests from upstream are included.

This commit forks path_abs from commit https://github.com/vitiral/path_abs/tree/8370838b6110786c2ba7cf7fd984a4783f37701c.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,7 +35,6 @@ dependencies = [
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "mruby-sys 2.0.1-12",
  "onig 4.3.2 (git+https://github.com/artichoke/rust-onig?branch=wasm)",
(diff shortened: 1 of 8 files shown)


lopopolo:
This commit also moves path absolutifying out of the kernel require implementation and into the `fs` module in artichoke-backend that wraps the VFS.

Issue #48, opened 2019-07-30
Enable Regexp in Wasm build


lopopolo:
Currently disabled with conditional compilation because of a link error under emscripten.

PR #68, opened 2019-08-01
Enable Regexp in Wasm build

lopopolo:
Works around a link failure under emscripten.

Updates the prepopulated example to use JSON and Regexp.

Fixes GH-48.


lopopolo:
Upstream bug report: https://github.com/emscripten-core/emscripten/issues/9135

Issue #145, opened 2019-08-07
Modify mruby-sys build to only generate sources and compile with cc crate.


lopopolo:
mruby-sys depends on the mruby build for codegen _and_ compilation, and linking. This requires a fairly massive dependency on a clang toolchain. We can avoid this by modifying our build configuration to only depend on mruby build for source generation and use the `cc` crate to build the generated sources.

Compiling with the `cc` crate instead of using `minirake` to compile gives several advantages:

- avoids dependency on clang installation
- gives cross builds for free
- gives `wasm32-unknown-unknown` target support for free.

We can do this by replacing the `conf.cc.command`, `conf.linker.command`, and `conf.ar.command` with `/usr/bin/true`.

See GH-137.

lopopolo:
I have a branch that experiments with this idea here: https://github.com/artichoke/artichoke/tree/mruby-sys-build

Able to build, but the playground fails to load and `cargo test` fails to compile mruby-sys.

Issue #137, opened 2019-08-06
Fix build warnings for mruby-sys


lopopolo:
```console
$ cargo clippy --all-targets --all-features
warning: couldn't execute `llvm-config --prefix` (error: No such file or directory (os error 2))
warning: set the LLVM_CONFIG_PATH environment variable to a valid `llvm-config` executable
    Finished dev [unoptimized + debuginfo] target(s) in 0.50s
```

I think this was caused by switching the mruby build toolchain to clang.

https://github.com/artichoke/artichoke/blob/ff426131ca32ead78a46684a6b50b17f4e2c26ac/mruby-sys/debug_build_config.rb#L10

I'm getting these warnings on macOS 10.14.6 with this clang

```console
$ clang --version
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

abelsromero:
Als `cargo build`, `cargo test`, etc. fail on Linux if clang is not installed.
This should probably be explained in the toolchain guide: https://github.com/artichoke/artichoke/blob/master/CONTRIBUTING.md#rust-toolchain

lopopolo:
Building `mruby-sys` currently depends on a `clang` toolchain being installed + `bintools` + `yacc` + `gperf` + `ruby`.

https://github.com/artichoke/artichoke/blob/6ad780b1a9899155bc7fe9a71a5b8eb3d16acc2f/.circleci/config.yml#L48-L61

I have ideas on how we can modify the `mruby-sys` build to depend on the `cc` crate. This would still depend on `yacc` for generating the parser C source and maybe `gperf`.

See GH-145.

lopopolo:
See https://github.com/artichoke/artichoke-ci/issues/4

lopopolo:
These warnings come from `clang-sys` which is pulled in by `bindgen`. These warnings will go away once `bindgen` releases a version that depends on `clang-sys` 1.0.0.

https://github.com/rust-lang/rust-bindgen/pull/1826#issue-448653594

lopopolo:
Fixed by https://github.com/artichoke/artichoke/pull/804

Issue #32, opened 2019-07-28
Remove dependency on mruby-*-ext mrbgems


lopopolo:
These implementations of Ruby core methods have been mostly buggy and not spec-compliant.

Use them as a starting point to implement them natively in Rust/evaled Ruby with the goal of passing ruby/spec.

- [ ] `mruby-compar-ext`
- [ ] `mruby-enum-ext`
- [ ] `mruby-numeric-ext`
- [ ] `mruby-array-ext`
- [ ] `mruby-hash-ext`
- [ ] `mruby-range-ext`
- [ ] `mruby-proc-ext`
- [ ] `mruby-symbol-ext`
- [ ] `mruby-object-ext`
- [ ] `mruby-kernel-ext`
- [ ] `mruby-class-ext`

lopopolo:
This change will allow eliminating dependency on `mrbc` which means:

- we can null out the `host` build for `mruby-sys`.
- no more dependency on `gcc`.

tkbky:
I'm working on for `mruby-enum-ext`

lopopolo:
Sounds good thanks @tkbky

tkbky:
picking up `mruby-numeric-ext`

lopopolo:
Thank you @tkbky

lopopolo:
Updated the checklist with the great work @tkbky has been doing. 💎 💪

lopopolo:
@tkbky has been making awesome progress on this ticket 🎉

lopopolo:
We can't yet remove `mruby-proc-ext` from `mruby-sys/build.rs` yet because `mruby-method` depends on it.

_Originally posted by @tkbky in https://github.com/artichoke/artichoke/pull/217#issuecomment-526615816_

PR #147, opened 2019-08-08
Use cc crate to build mruby-sys static library

lopopolo:
Modify mruby-sys build to use the `cc` crate to compile one static library called `libmrubysys.a`.

Compiling with `cc` crate makes it easier to compile a wasm32-unknown-unknown build. We still can't do that because of `onig`.

To accomplish this, the `host` build now compiles the bare minimum of mruby to bootstrap the `sys` build. Until GH-32 is complete, mruby-sys depends on gems with Ruby sources. These Ruby sources get translated to bytecode embedded in C sources using the `mrbc` binary. The `host` build produces this binary.

The `sys` build only generates C sources using `minirake` by replacing compiler commands with `true`. `build.rs` the collects all of the headers and sources and compiles them with `cc` crate.

Adds docs for GH-137. `cc` is required to build. By default, that is expected to be `gcc`.
Fixes GH-145.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ jobs:
           # https://github.com/mruby/mruby/blob/master/doc/guides/compile.md#prerequisites
           name: Install mruby Build Dependencies
           command: |
-            sudo apt-get install -y binutils bison gperf ruby-full
+            sudo apt-get install -y bison gperf ruby-full
             bison --version
             gperf --version
             ruby --version
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,6 +81,28 @@ cargo fmt -- --check
 cargo clippy --all-targets --all-features
 ```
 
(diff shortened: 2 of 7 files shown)


lopopolo:
This build is a bit fragile right now, but removing dependencies on all mrbgems except the compiler will make things easier to maintain.

lopopolo:
This PR also changes all of the path handling in `mruby-sys/build.rs` to use `PathBuf::join` instead of string formatting which should hopefully make the build work on platforms other than the ones I've tested.

lopopolo:
This change caused the build to become [flaky](https://circleci.com/gh/artichoke/artichoke/1135?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

Hopefully will be fixed once we no longer need to build `mrbc`. See GH-32.

lopopolo:
See GH-169.

Issue #113, opened 2019-08-04
Implement ENV object in Rust


lopopolo:
[`ENV` is currently a stub](https://github.com/artichoke/artichoke/blob/60fe1be8d92660185909ea26a7ce633af5844224/artichoke-backend/src/extn/core/env.rb).

To support both `wasm32` and embedded builds, we should implement ENV with multiple backends:

- `std::env`
- `HashMap`

For an example of a similar component implemented in Rust with support for multiple backends, see [`Regexp`](https://github.com/artichoke/artichoke/blob/60fe1be8d92660185909ea26a7ce633af5844224/artichoke-backend/src/extn/core/regexp/mod.rs#L129-L132).

Expose methods on `State` to toggle the `ENV` backend.

cdeler:
May I take this one?

lopopolo:
Absolutely, it's all yours @cdeler. 💎

lopopolo:
To support the in-memory backend, we my want to expose manipulating the interpreter environment as a trait on the interpreter, similar to `Warn`.

lopopolo:
@cdeler I just realized the `ENV` API is fairly large. I don't think it is necessary to complete the [entire `ENV` API](https://ruby-doc.org/core-2.5.0/ENV.html) in one PR. Maybe focus on `ENV::[]` and `ENV::[]=` first since that is the part the glues together Rust and Ruby.

I had a peek at your branch and I think it is looking pretty good. 🙏 

Looking at the API, I think the only pieces we must implement in Rust are `ENV::[]`, `ENV::[]=`, `ENV::delete` and `ENV::to_h`. These four APIs would allow the rest to be implemented in Ruby.

cdeler:
@lopopolo I've done with `ENV::[]` and `ENV::[]=`

I think `ENV::delete` could be emulated through `ENV::[]=nil`

I should add `ENV::to_h` or create PR?

cdeler:
@lopopolo The diff for `ENV::[]` and `ENV::[]=` is [here](https://github.com/artichoke/artichoke/compare/master...cdeler:%23113_ENV_implementation?expand=1).

lopopolo:
@cdeler we can add the `HashMap` backend as an add on in a subsequent PR. We'll just keep this ticket open until the `ENV` implementation is complete.

I probably won't have time to review your PR until the weekend. If you'd like to add `ENV::to_h` before then, that sounds good. If you'd prefer to push what you have, that works too. 😄 💎

[...]

PR #148, opened 2019-08-09
#113 env implementation

cdeler:


cdeler:
I'll check the errors tomorrow morning

nirvdrum:
Apologies if this is the wrong place to be asking, but what's the motivation for the `HashMap`-based implementation? If I'm not mistaken, that would fall out of sync if any native extension updated the environment.

As for the Ruby part of the API, you may want to take a look at [how we did it in TruffleRuby](https://github.com/oracle/truffleruby/blob/master/src/main/ruby/truffleruby/core/env.rb), which is an extension of how Rubinius did it. The `Truffle::POSIX` calls are FFI calls and should be easy to adapt to your approach (perhaps as "primatives"). The `Truffle::System` methods are implemented in the host language and should also be easy to adapt.

lopopolo:
Hi @nirvdrum, the approach Truffle has taken looks to be similar to how I was imagining things. Only `ENV::[]`, `ENV::[]=`, and `ENV::to_h` will be implemented in Rust and the rest of the API will be implemented in Ruby using these native interfaces.

You’re right that a `HashMap`-based `ENV` will not stay in sync with the system environment. One of Artichoke’s goals is to be embedded in potentially untrusted environments, like as a scripting engine in a game or a templating engine in a web server. In these environments, allowing Ruby to manipulate the environment of the host process is undesirable. The `HashMap` implementation is an API compatible stub.

artichoke-backend/src/extn/core/env/env_object.rs
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        Ok(())
+    }
+    fn set_internal(key: &str, value: Option<&String>) {
+        T::set_value(key, value);
+    }
+
+    fn string_to_value(interp: &Artichoke, key: &str) -> Value {
+        let gc_was_enabled = interp.disable_gc();
```
lopopolo: I think this GC disabling was found in the `String#scan` implementation. These explicit disables were to work around a bug https://github.com/artichoke/ferrocarril/pull/141 and are no longer required. I've filed GH-151 to clean up this bad pattern.

artichoke-backend/src/extn/core/env/env_object.rs
```diff
@@ -0,0 +1,295 @@
...
+            return Err(EnvError::InvalidSetValue);
+        }
+
+        Ok(())
+    }
+    fn set_internal(key: &str, value: Option<&String>) {
+        T::set_value(key, value);
+    }
+
+    fn string_to_value(interp: &Artichoke, key: &str) -> Value {
```
lopopolo: this method seems unnecessary since all we're doing is calling the converter.

artichoke-backend/src/extn/core/env/env_object.rs
```diff
@@ -0,0 +1,295 @@
...
+        let result = Value::convert(interp, key);
+
+        if gc_was_enabled {
+            interp.enable_gc();
+        }
+
+        result
+    }
+
+    fn hash_map_to_value(interp: &Artichoke, data: HashMap<String, String>) -> Value {
```
lopopolo: Same as `string_to_value`, the GC management is not needed and all we're doing is calling the converter. This function is not needed.

artichoke-backend/src/extn/core/env/env.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+ENV = EnvClass.new()
+
```
lopopolo: You can fix the lint errors here by running `yarn lint`.

artichoke-backend/src/extn/core/env/env_object.rs
```diff
@@ -0,0 +1,295 @@
...
+            other.as_mut_ptr(),
+        );
+        let other = other.assume_init();
+
+        let arg_value = Value::new(interp, other);
+
+        Some(arg_value.to_s())
+    }
+
+    unsafe fn get_internal(interp: &Artichoke, arg_name: &str) -> sys::mrb_value {
```
lopopolo: this function should not be necessary since we can just call `T::get_value` and use a converter.

```rust
Value::convert(interp, T::get_value(arg_name)).inner()
```

artichoke-backend/src/extn/core/env/env_object.rs
```diff
@@ -0,0 +1,295 @@
...
+        let value = T::get_value(arg_name);
+
+        match value {
+            Some(variable_value) => Self::string_to_value(interp, &variable_value).inner(),
+            None => sys::mrb_sys_nil_value(),
+        }
+    }
+
+    unsafe fn env_to_h_internal(interp: &Artichoke) -> sys::mrb_value {
```
lopopolo: ditto `get_internal` above.

artichoke-backend/src/extn/core/env/env_object.rs
```diff
@@ -0,0 +1,295 @@
...
+        let (key, value) = Self::extract_two_string_args(&interp);
+
+        match Self::validate_set_args(&key, &value) {
+            Ok(_res) => {
+                Self::set_internal(&key, value.as_ref());
+                Self::get_internal(&interp, &key)
+            }
+            Err(error) => {
+                match error {
+                    EnvError::InvalidSetKey => {
```
lopopolo: this nested match can be avoided by making the match arms in the outer match:

```
Err(EnvError::InvalidSetKey)
Env(EnvError:: InvalidSetValue)
```

artichoke-backend/src/extn/core/env/env_object.rs
```diff
@@ -0,0 +1,295 @@
...
+                        ArgumentError::raise(interp, "Invalid key for ENV set")
+                    }
+                    EnvError::InvalidSetValue => {
+                        ArgumentError::raise(interp, "Invalid value for ENV set")
+                    }
+                };
+                sys::mrb_sys_nil_value()
```
lopopolo: `raise` returns nil, so this explicit sys nil is not required.

cdeler:
@lopopolo I'm not sure, but I think I've done with all issues (except `&[u8]` data types for all back-ends)

lopopolo:
@cdeler this looks great thank you! 💪 

I've filed some follow up tickets to address some of the outstanding items and complete the implementation of the whole `ENV` API.

Thank you for your contribution to Artichoke 💎

lopopolo:
💪 💎 

```console
$ env ARTICHOKE="💎" cargo run --bin airb
mruby 2.0 [2.0.1-12]
>>> ENV["USER"]
=> "lopopolo"
>>> ENV["ARTICHOKE"]
=> "💎"
```

PR #157, opened 2019-08-11
Env followup

lopopolo:
ENV object: Follow up from GH-148.

- Modify backend trait to return Result with Env Error.
- Make env module Error into an Error type.
- Handle null byte in env::var_os (fixes segfaulting test introduced in
  previous commit).
- Return Error::Fatal instead of unwrapping, e.g. HashMap backend with
  RwLock.
- Move arg validation into backend implementation.
- Move calls to raise into C API to work prevent memory leaks with
  longjmp.
- get set and to_h internal return Value instead of sys::mrb_value.

```
mruby 2.0 [2.0.1-12]
>>> ENV["as\0"]
Backtrace:
    (airb):1: bad environment variable name: contains null byte (ArgumentError)
    (airb):1
>>> ENV["as\0"] = "a"
Backtrace:
    (airb):2: bad environment variable name: contains null byte (ArgumentError)
    (airb):2
>>> ENV["as"] = "a\0"
Backtrace:
    (airb):3: bad environment variable value: contains null byte (ArgumentError)
    (airb):3
```

### DIFF
--- a/artichoke-backend/src/extn/core/env/backends.rs
+++ b/artichoke-backend/src/extn/core/env/backends.rs
@@ -1,12 +1,13 @@
 use std::collections::HashMap;
 use std::env;
-use std::ffi::OsString;
 use std::sync::RwLock;
+
+use crate::extn::core::env::Error;
 
 pub trait EnvBackend {
     fn new() -> Self;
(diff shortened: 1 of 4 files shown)


lopopolo:
cc @cdeler

Issue #86, opened 2019-08-03
'delegate', 'monitor', and 'uri' are auto-required on boot of MRI


lopopolo:
Modify `uri.rs` in `extn::stdlib` to eval `require 'uri'` on package init.

lopopolo:
Thanks for uncovering this @johnwahba.

lopopolo:
```console
$ ruby -e 'puts ''$''LOADED_FEATURES.sort'
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/delegate.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/monitor.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/basic_specification.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/bundler_version_finder.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/compatibility.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_gem.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_warn.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/defaults.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/dependency.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/deprecate.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/errors.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/exceptions.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/path_support.rb
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/platform.rb
[...]

PR #117, opened 2019-08-04
[wip] Wahba run autoimport in build

johnwahba:
Run the auto import for benchmark in the build step

Issue #12, opened 2019-05-15
Implement the Ruby 2.6 standard library tracking ticket


lopopolo:
Implement the [Ruby 2.6.3 Standard Library](https://ruby-doc.org/stdlib-2.6.3/).

This ticket tracks the scheduling and completion of the [Standard Library Milestone]. Checkmarks mean a ticket to track package implementation has been created.

- [x] abbrev - GH-74
- [x] base64 - GH-75
- [x] benchmark - GH-76
- [ ] bigdecimal
- [ ] cgi
- [x] cmath
- [ ] coverage
- [ ] csv
- [ ] date
- [ ] dbm
- [ ] debug
- [x] delegate - completed artichoke/ferrocarril#110
- [ ] digest
- [ ] drb
- [ ] e2mmap
- [ ] English
- [ ] erb
- [ ] etc
- [ ] expect
- [ ] extmk
- [ ] fcntl
- [ ] fiddle
- [ ] fileutils
- [ ] find
- [x] forwardable - completed artichoke/ferrocarril#108
- [ ] gdbm
- [ ] getoptlong
- [ ] io/console
- [ ] io/nonblock
- [ ] io/wait
- [ ] ipaddr
- [ ] irb
- [ ] json - pure Ruby implementation completed artichoke/ferrocarril#126
- [ ] logger
- [ ] mathn
- [ ] matrix
- [ ] mkmf
- [ ] monitor - completed for single-threaded implementation artichoke/ferrocarril#89
- [ ] mutex_m
- [ ] net/ftp
- [ ] net/http
- [ ] net/imap
- [ ] net/pop
- [ ] net/smtp
- [ ] net/telnet
- [ ] nkf
- [ ] objspace
- [ ] observer
- [ ] open-uri
- [ ] open3
- [ ] openssl
- [ ] optparse
- [x] ostruct - completed artichoke/ferrocarril#112
- [ ] pathname
- [ ] prettyprint
- [ ] prime
- [ ] profile
- [ ] profiler
- [ ] pstore
- [ ] psych
- [ ] pty
- [ ] racc
- [ ] racc/parser
- [ ] rake
- [ ] rdoc
- [ ] readline
- [ ] resolv
- [ ] resolv-replace
- [ ] rexml
- [ ] rinda
- [ ] ripper
- [ ] rss
- [ ] rubygems
- [ ] scanf
- [ ] sdbm
- [x] securerandom - GH-34
- [x] set - completed artichoke/ferrocarril#111
[...]

PR #158, opened 2019-08-11
Generate uri package sources

lopopolo:
Generate the URI package using @johnwahba's package generator scripts. the mruby backend remains spec-compliant for URI.

This PR required some tweaks to the package generator (and to fix some bugs in the generator that I introduced 😅)

- Drain the load path to remove all paths before adding vendored lib root
- Pass flags to Ruby invocation to disable autoloading of default gems (thanks @eregon for pointing this out in GH-86)
- Fix some templating errors. Try to format the Rust code generated by the template consistently even though we don't lint this code.
- Track the class of each loaded constant (`Class` vs `Module`, vs other).
- Define class or module in the interpreter appropriately for each discovered constant.

This PR modifies the `artichoke-backend` `build.rs` to use `PathBuf` everywhere instead of String concatenation to bring it in line with the mruby-sys build. I hope this helps with portability.

This PR tracks the two patches required to pass spec for URI on the mruby backend.
This PR removes the artisanal patched and RuboCopped URI sources from `extn`.


lopopolo:
This PR validates the approach introduced in GH-85 and GH-117 in support of GH-12.

lopopolo:
Thank you @johnwahba 🙏

PR #164, opened 2019-08-12
Remove dependency on mruby-array-ext and mruby-hash-ext

lopopolo:
Working toward eliminating build dependency on `mrbc`. See GH-32.


lopopolo:
There are some methods in mruby-array-ext and mruby-hash-ext that are implemented in C. The only one I've implemented in Artichoke is `Array#slice!` since it is required for the `uri` specs to pass.

Unimplemented methods:

- `Array#assoc`
- `Array#at`
- `Array#rassoc`
- `Array#values_at`
- `Hash#values_at`
- `Hash#slice`

These methods must be implemented for spec compliance, but are non-blocking now.

PR #169, opened 2019-08-12
Fix mruby-sys build flakiness when compiling mruby-compiler mrbgem

lopopolo:
I think the build flakiness started when we switched back to gcc for the bootstrap
build. Maybe something to do with the buster upgrade?

Either way, cc crate already requires clang so it makes sense to unify things.


lopopolo:
It appears the culprit was gperf 3.1

lopopolo:
Filed an issue upstream about build failures with gperf 3.1: https://github.com/mruby/mruby/issues/4628

Issue #149, opened 2019-08-10
Building mruby-sys and rust-onig without emscripten requires trunk clang


lopopolo:
mruby-sys and rust-onig use computed gotos.

Building with clang instead of emscripten for the wasm32 targets requires a clang with this fix: https://bugs.llvm.org/show_bug.cgi?id=42498

lopopolo:
Something like this will install trunk clang in CircleCI:

```yaml
      - run:
          # needed for cc crate in build.rs
          name: Install LLVM toolchain
          command: |
            sudo apt-get install software-properties-common
            wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
            sudo apt-add-repository 'deb http://apt.llvm.org/buster/ llvm-toolchain-buster main'
            sudo apt-get update
            sudo apt-get install -y clang lld
            clang --version
```

lopopolo:
clang 10 snapshots are not available via the apt repository.

lopopolo:
The Dockerfile for the base image in https://github.com/artichoke/artichoke-ci bundles snapshot clang-9.

singpolyma-shopify:
If you use:

```diff
diff --git a/src/regexec.c b/src/regexec.c
index fa61839..94aa465 100644
--- a/src/regexec.c
+++ b/src/regexec.c
@@ -2338,9 +2338,11 @@ typedef struct {
 } posix_regmatch_t;
 
 
+#if 0
 #ifdef __GNUC__
 #define USE_THREADED_CODE
 #endif
+#endif
 
 #ifdef USE_THREADED_CODE
 ```

Then the computed gotos are not needed anymore and it will build with any llvm-8 for wasm32-unknown-unknown

singpolyma-shopify:
[build.sh.txt](https://github.com/artichoke/artichoke/files/3493997/build.sh.txt)

That should reproduce my current state, including the mis-linked airb.wasm file

lopopolo:
Thanks for the build steps @singpolyma-shopify. I've made some progress on this: https://github.com/artichoke/artichoke/pull/172.

Still failing at link type, but new and exciting linker errors. If you could help me sort these out I'd be 🙏

lopopolo:
[...]

PR #172, opened 2019-08-13
[WIP] build under wasm32-unknown-unknown target

lopopolo:
This branch makes much progress to build artichoke under the wasm32-unknown-unknown target.

This branch includes a new `wasm` crate which exports a funtion that initializes an interpreter and evals a single `Integer` expression and returns the result. This code can be built with:

```console
$ cargo build --target wasm32-unknown-unknown -p wasm
```

The build currently fails at the linker stage:

```
  = note: rust-lld: error: function signature mismatch: mrb_protect
          >>> defined as (i32, i32, i32, i64, i32, i32, i32) -> void in /home/artichoke/target/wasm32-unknown-unknown/debug/deps/artichoke_backend.1ja1zfqnm5ty63ed.rcgu.o
          >>> defined as (i32, i32, i32, i32, i32) -> void in /home/artichoke/target/wasm32-unknown-unknown/debug/deps/libmruby_sys.rlib(exception.o)

          rust-lld: error: function signature mismatch: mrb_gc_protect
          >>> defined as (i32, i64, i32, i32) -> void in /home/artichoke/target/wasm32-unknown-unknown/debug/deps/artichoke_backend.7pvpltx2v0gmmy3.rcgu.o
          >>> defined as (i32, i32) -> void in /home/artichoke/target/wasm32-unknown-unknown/debug/deps/libmruby_sys.rlib(exception.o)
```

In order to get to this point, this branch makes the following changes:

- Update [rust-onig fork](https://github.com/artichoke/rust-onig) to the latest 5.0.0 master.
- Update the oniguruma dependency to a [fork](https://github.com/artichoke/oniguruma/tree/disable-direct-threading) that contains these changes:
  - Introduce a check for a macro that, when defined, disables direct threading in the regex engine. Direct threading is only enabled in snapshot clang. See GH-149. Defining this macro enables building oniguruma for wasm32-unknown-unknown on clang-8.
  - Explicitly `#include <stdint.h>` in two sources. This header defines the `uintptr_t ` type which is part of C99, but optional. The emscripten wasm headers we vendor do not automatically define this type and require an explicit include.
- Vendor emscipten latest-upstream headers in onig_sys in rust-onig.
- Include these headers for wasm targets.
- Alter onig_sys `build.rs` to handle `CARGO_CFG_TARGET_FAMILY` being unset.
- Define `ONIG_EXTERN` in wasm builds to set export visibility.
- Define `MRB_DISABLE_DIRECT_THREADING` for wasm builds of mruby-sys.
- Update rust-onig dep in artichoke-backend to new fork branch.
- Update `Regexp` implementation in artichoke-backend to compile with API changes in onig 5.0.

To build for wasm, I've been using this docker container: https://github.com/artichoke/artichoke-ci

cc @singpolyma-shopify

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git
+node_modules
+target
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -803,6 +803,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "wasm"
+version = "0.1.0"
+dependencies = [
+ "artichoke-backend 0.1.0",
+]
+
+[[package]]
 name = "which"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,7 +43,7 @@ dependencies = [
  "libc 0.2.60 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "mruby-sys 2.0.1-14",
- "onig 5.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "onig 5.0.0 (git+https://github.com/artichoke/rust-onig?rev=bb6482017f6e8a5b7e75a10d5151cee0191bd6b8)",
  "quickcheck 0.8.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "quickcheck_macros 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -370,18 +370,18 @@ dependencies = [
 [[package]]
 name = "onig"
 version = "5.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
+source = "git+https://github.com/artichoke/rust-onig?rev=bb6482017f6e8a5b7e75a10d5151cee0191bd6b8#bb6482017f6e8a5b7e75a10d5151cee0191bd6b8"
 dependencies = [
  "bitflags 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.60 (registry+https://github.com/rust-lang/crates.io-index)",
- "onig_sys 69.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "onig_sys 69.2.0 (git+https://github.com/artichoke/rust-onig?rev=bb6482017f6e8a5b7e75a10d5151cee0191bd6b8)",
 ]
 
 [[package]]
 name = "onig_sys"
 version = "69.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
+source = "git+https://github.com/artichoke/rust-onig?rev=bb6482017f6e8a5b7e75a10d5151cee0191bd6b8#bb6482017f6e8a5b7e75a10d5151cee0191bd6b8"
 dependencies = [
  "bindgen 0.50.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "cc 1.0.38 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -947,8 +947,8 @@ dependencies = [
 "checksum nix 0.13.1 (registry+https://github.com/rust-lang/crates.io-index)" = "4dbdc256eaac2e3bd236d93ad999d3479ef775c863dbda3068c4006a92eec51b"
 "checksum nodrop 0.1.13 (registry+https://github.com/rust-lang/crates.io-index)" = "2f9667ddcc6cc8a43afc9b7917599d7216aa09c463919ea32c59ed6cac8bc945"
 "checksum nom 4.2.3 (registry+https://github.com/rust-lang/crates.io-index)" = "2ad2a91a8e869eeb30b9cb3119ae87773a8f4ae617f41b1eb9c154b2905f7bd6"
-"checksum onig 5.0.0 (registry+https://github.com/rust-lang/crates.io-index)" = "e4e723fc996fff1aeab8f62205f3e8528bf498bdd5eadb2784d2d31f30077947"
-"checksum onig_sys 69.2.0 (registry+https://github.com/rust-lang/crates.io-index)" = "0a8d4efbf5f59cece01f539305191485b651acb3785b9d5eef05749f0496514e"
+"checksum onig 5.0.0 (git+https://github.com/artichoke/rust-onig?rev=bb6482017f6e8a5b7e75a10d5151cee0191bd6b8)" = "<none>"
+"checksum onig_sys 69.2.0 (git+https://github.com/artichoke/rust-onig?rev=bb6482017f6e8a5b7e75a10d5151cee0191bd6b8)" = "<none>"
 "checksum peeking_take_while 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)" = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
 "checksum pkg-config 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)" = "a7c1d2cfa5a714db3b5f24f0915e74fcdf91d09d496ba61329705dda7774d2af"
 "checksum proc-macro2 0.4.30 (registry+https://github.com/rust-lang/crates.io-index)" = "cf3d2011ab5c909338f7887f4fc896d35932e29146c12c8d01da6b22a80ba759"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -162,6 +162,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "2.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "byteorder"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -807,7 +812,56 @@ name = "wasm"
 version = "0.1.0"
 dependencies = [
  "artichoke-backend 0.1.0",
+ "wasm-bindgen 0.2.48 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
+
+[[package]]
+name = "wasm-bindgen"
+version = "0.2.48"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "wasm-bindgen-macro 0.2.48 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "wasm-bindgen-backend"
+version = "0.2.48"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bumpalo 2.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 1.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "proc-macro2 0.4.30 (registry+https://github.com/rust-lang/crates.io-index)",
+ "quote 0.6.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "syn 0.15.42 (registry+https://github.com/rust-lang/crates.io-index)",
+ "wasm-bindgen-shared 0.2.48 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "wasm-bindgen-macro"
+version = "0.2.48"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "quote 0.6.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "wasm-bindgen-macro-support 0.2.48 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "wasm-bindgen-macro-support"
+version = "0.2.48"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "proc-macro2 0.4.30 (registry+https://github.com/rust-lang/crates.io-index)",
+ "quote 0.6.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "syn 0.15.42 (registry+https://github.com/rust-lang/crates.io-index)",
+ "wasm-bindgen-backend 0.2.48 (registry+https://github.com/rust-lang/crates.io-index)",
+ "wasm-bindgen-shared 0.2.48 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "wasm-bindgen-shared"
+version = "0.2.48"
+source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "which"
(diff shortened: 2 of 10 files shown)


singpolyma-shopify:
Interesting. I cannot reproduce the linker errors you get. I just make a fresh clone to be sure no weird local state and then checked out this branch.  Running `cargo build --target wasm32-unknown-unknown -p wasm` seems to produce the same problem I described before, where I get a `wasm.wasm` file created, but it does not seem to be linked to mruby or oniguruma (and is instead trying to import all their symbols from the environment).

singpolyma-shopify:
My other pure-C wasm32-unknown-unknown project uses `-flto` in the build step to get LLVM IR output and then links with `wasm-ld` from LLVM.  I have attached the Makefile from that project in case it might contain useful clues as compared to what `rust-lld` is trying to do here.
[Makefile.txt](https://github.com/artichoke/artichoke/files/3497296/Makefile.txt)

singpolyma-shopify:
So, though I cannot reproduce the link so far, I had a look at the bindgen output for `mrb_gc_protect` and it could be that different passes of clang vs rustc vs rust-lld are seeing different types for `mrb_value`?  That could result in a single value in some cases or an unpacked struct argument in other cases.  It looks like `MRB_NAN_BOXING` vs `MRB_WORD_BOXING` vs `boxing_no.h` (the default) would produce different layouts so maybe different passes are picking up a different one?

lopopolo:
I’m willing to bet this has something to do with rust trying to link an rlib instead of a cdylib.

I’ll continue to iterate in this tonight.

singpolyma-shopify:
```rust
pub union mrb_value__bindgen_ty_1 { pub f : mrb_float , pub p : * mut :: std :: os :: raw :: c_void , pub i : mrb_int , pub sym : mrb_sym , _bindgen_union_align : u64 , }

pub enum mrb_vtype { MRB_TT_FALSE = 0 , MRB_TT_FREE = 1 , MRB_TT_TRUE = 2 , MRB_TT_FIXNUM = 3 , MRB_TT_SYMBOL = 4 , MRB_TT_UNDEF = 5 , MRB_TT_FLOAT = 6 , MRB_TT_CPTR = 7 , MRB_TT_OBJECT = 8 , MRB_TT_CLASS = 9 , MRB_TT_MODULE = 10 , MRB_TT_ICLASS = 11 , MRB_TT_SCLASS = 12 , MRB_TT_PROC = 13 , MRB_TT_ARRAY = 14 , MRB_TT_HASH = 15 , MRB_TT_STRING = 16 , MRB_TT_RANGE = 17 , MRB_TT_EXCEPTION = 18 , MRB_TT_FILE = 19 , MRB_TT_ENV = 20 , MRB_TT_DATA = 21 , MRB_TT_FIBER = 22 , MRB_TT_ISTRUCT = 23 , MRB_TT_BREAK = 24 , MRB_TT_MAXDEFINE = 25 , }

pub struct mrb_value { pub value : mrb_value__bindgen_ty_1 , pub tt : mrb_vtype , pub __bindgen_padding_0 : u32 , }
```

This is from the `bindgen` output -- This looks very much like `mrb_value` is expected to be `i64, i32, i32` as in the error message (64 bits for at least the union padding, then 32 bits for the enum, then 32 bits for the struct padding).

However the definition from boxing_no.h (which I know must be being used because I deleted the other two):

```C
typedef struct mrb_value {
  union {
#ifndef MRB_WITHOUT_FLOAT
    mrb_float f;
#endif
    void *p;
    mrb_int i;
    mrb_sym sym;
  } value;
  enum mrb_vtype tt;
} mrb_value;
```

Is definitely at least two members, so I'm not sure why any part of this would expect just a single `i32` in the `mrb_value` position

singpolyma-shopify:
I've managed to reproduce the link errors.

Ok, so, on llvm-8 I get a lot of these link failures, but on llvm-9 I get only the two listed in the OP here.  So it cannot be `mrb_value` definition related alone or there would be many more.

Commenting out every place in `artichoke-backend` that calls `sys::mrb_protect` causes the errors to go away... even though I'm still calling `mrb_gc_protect`!  Obviously this is not a code change we could keep, but I just wanted to confirm what the linker behaviour would be.

With the commented out calls, I get a `wasm.wasm` file again, and it *still* shows importing all the `mrb_*` and `onig_*` functions from the environment, which says to me that even though we're seeing linking errors also something is not linking.  Though maybe that's why I only get the two errors on llvm-9 because the other calls are not being linked?

lopopolo:
@singpolyma-shopify I was able to verify that when commenting out the calls to `mrb_protect` and `mrb_gc_protect` that cargo builds successfully, but the resulting wasm bundles do not link the foreign C FFI functions.

Using the example of the [fork of rust-onig](https://github.com/artichoke/rust-onig/tree/wasm3) that I'm working on since it is easier to build on my Mac with:

```sh
AR=/usr/local/Cellar/llvm/8.0.0_1/bin/llvm-ar CC=/usr/local/Cellar/llvm/8.0.0_1/bin/clang cargo build --target wasm32-unknown-unknown
```

This is the resulting wasm bundle:

```console
$ wasm2wat target/wasm32-unknown-unknown/debug/onig_sys.wasm
(module
  (type (;0;) (func))
  (func $__wasm_call_ctors (type 0))
  (table (;0;) 1 1 funcref)
  (memory (;0;) 17)
  (global (;0;) (mut i32) (i32.const 1048576))
  (global (;1;) i32 (i32.const 1048610))
  (global (;2;) i32 (i32.const 1048610))
  (global (;3;) i32 (i32.const 1048576))
  (export "memory" (memory 0))
  (export "__heap_base" (global 1))
  (export "__data_end" (global 2))
  (export "__rustc_debug_gdb_scripts_section__" (global 3))
  (data (;0;) (i32.const 1048576) "\01gdb_load_rust_pretty_printers.py\00"))
```

I'm wondering if this is a bug in `rust-lld` or `llvm`.

lopopolo:
Rust tests are known to fail because I've stubbed out eval and funcall.

lopopolo:
Rather than continuing to tear my hair out over foreign C functions and the wasm linker, I think it might be more fruitful to invest in turning these C dependencies into native Rust.

I last tried running c2rust on mruby a while ago, see the [diamond branch](https://github.com/artichoke/artichoke/tree/diamond/diamond). The last time I tried to run c2rust, I got hung up on computed gotos in the VM, but this was before I knew about the macro to disable direct threading.

The other thing that was a blocker was c2rust doesn't handle source files with duplicate filenames. The patching story in mruby-sys crate is much better now which should make giving every mrbgem source file a unique filename not so bad.

It seems feasible to clone mruby-sys crate, set the appropriate macros to make the C transpilable by c2rust. I'm not sure if this will help with the wasm linking story, but building diamond with a wasm target and running wasm2wat and inspecting the output would be a quick way to test that out.

Oniguruma is much simpler to compile and I think all we'd have to do is set `ONIG_DISABLE_DIRECT_THREADING` and instrument the build with `intercept-build`.

singpolyma-shopify:
Just an update for you, today I got a build of mruby (without
artichoke) working under wasm32-unknown-unknown, so I thought I'd
share that recipe with you in case it is helpful.

I've attached a tarball of the code.  To use it go into the vendored
mruby-* folder and run:

make -f wasm.make
make -f wasm.make

(it will error the first time and you have to run it again, I know why
I just haven't fixed that part yet).

[mruby.tar.gz](https://github.com/artichoke/artichoke/files/3507061/mruby.tar.gz)

lopopolo:
With the following diff, I am able to build an mruby wasm bundle on my mac with llvm-8 that exports all of its functions.

```diff
diff --git a/vendor/mruby-bc7c5d3/wasm.make b/vendor/mruby-bc7c5d3/wasm.make
index d09083c..0a31116 100644
--- a/vendor/mruby-bc7c5d3/wasm.make
+++ b/vendor/mruby-bc7c5d3/wasm.make
@@ -10,14 +10,14 @@ COMPILE_FLAGS = -DMRB_DISABLE_STDIO -DMRB_UTF8_STRING -DMRB_DISABLE_DIRECT_THREA
                --target=wasm32-unknown-unknown \
                -flto \
                -fno-builtin-malloc -fno-builtin-realloc -nostdlib \
-               -fvisibility=hidden \
+               -fvisibility=default \
                -std=c99 \
                -ffunction-sections \
                -fdata-sections \
                -I./libc/ -I./include/ -I./mrbgems/mruby-compiler/core/

 $(OUTPUT): build/sys/mrblib/mrblib.c $(OBJ) $(MUSL) Makefile
-       wasm-ld-8 \
+       wasm-ld \
                -o $(OUTPUT) \
                --strip-all \
                --export-dynamic \
```

I'd like to upstream a [patch to mruby so we can target the clang visibility attribute to only `MRB_API` functions](https://github.com/artichoke/artichoke/commit/071de9643d790b1e22d83f322beabc65be3c51d2.patch).

This results that a `library.wasm` that can successfully call the following API functions in node (although I am having trouble marshalling the pointer to the buffer:

```js
const fs = require("fs");

// Stubs for functions pulled from musl libc.
const env = {
  __stdio_exit_needed: function() {},
  scalbnl: function() {},
  __lockfile: function() {},
  __unlockfile: function() {},
  mbrtowc: function() {},
  mbsinit: function() {},
  vsniprintf: function() {},
  __small_vsnprintf: function() {},
  fmodl: function() {}
};

// Load Wasm bundle for libmruby.a
const bundle = fs.readFileSync("./library.wasm");
const mruby = WebAssembly.instantiate(new Uint8Array(bundle), { env });

const code = "2 + 3";

mruby
  .then(mruby => {
    const mrb = mruby.instance.exports.mrb_open();
    mruby.instance.exports.mrb_full_gc(mrb);
    const buffer = mruby.instance.exports.mrb_malloc(mrb, code.length);
    console.log(mrb);
    console.log(typeof mrb);
    console.log(buffer);
    console.log(typeof buffer);
    // const value = mruby.instance.exports.mrb_load_nstring(mrb, buffer, code.length, mrb);
    const value = mruby.instance.exports.mrb_load_nstring_cxt(mrb, 272928, code.length, 0);
    console.log(value);
    console.log(`2 + 3 = ${value.i}`);
  })
  .catch(e => console.log(e));
```

This code prints `100` to the console:

```js
const fs = require("fs");

// Stubs for functions pulled from musl libc.
const env = {
  __stdio_exit_needed: function() {},
  scalbnl: function() {},
  __lockfile: function() {},
  __unlockfile: function() {},
  mbrtowc: function() {},
  mbsinit: function() {},
  vsniprintf: function() {},
  __small_vsnprintf: function() {},
  fmodl: function() {}
};

// Load Wasm bundle for libmruby.a
const bundle = fs.readFileSync("./library.wasm");
const mruby = WebAssembly.instantiate(new Uint8Array(bundle), { env });

const code = "2 + 3";

mruby
  .then(mruby => {
    const value = mruby.instance.exports.run();
    console.log(value);
  })
  .catch(e => console.log(e));
```

lopopolo:
Nightly rustc uses llvm-9: https://github.com/rust-lang/rust/pull/63302. I tried to compile artichoke with this rust:

```console
artichoke@81112097025f:/app$ rustc --version --verbose
rustc 1.39.0-nightly (f7af19c27 2019-08-15)
binary: rustc
commit-hash: f7af19c279b8b7ea3d2c21fcbd67164af8d5d968
commit-date: 2019-08-15
host: x86_64-unknown-linux-gnu
release: 1.39.0-nightly
LLVM version: 9.0
```

And we get ... more linker errors! This matches up more closely what I've been seeing in the generated wasm. For example, `mrb_sys_class_of_value` has the following signature when compiling the C lib standalone:

```wasm
(export "mrb_sys_class_of_value" (func 758))
(func (;758;) (type 16) (param i32 i32) (result i32)
```

the Rust compiled version has this signature:

```wasm
(import "env" "mrb_sys_class_of_value" (func $mrb_sys_class_of_value (type 10)))
(type (;10;) (func (param i32 i64 i32 i32) (result i32)))
```

I tried munging the type signatures manually in the wat, but the generated code pushes variables onto the stack according to the compiled in signature.

Next, I tried to compile _only_ `mruby-sys` and include a run function so the functions get linked in:

I added these flags from @singpolyma-shopify's makefile to the `mruby-sys/build.rs`:

```
-function-sections
-fdata-sections
-flto
```

And added a dependency on `wasm-bindgen` and this code to `mruby-sys/src/lib.rs`:

```rust
use wasm_bindgen::prelude::*;
/// docs
#[wasm_bindgen]
pub fn run() {
    unsafe {
        let mrb = mrb_open();
        let _ = mrb_sys_class_of_value(mrb, mrb_sys_nil_value());
    }
}
```

Which results in the following new and exciting linker error!

```
  = note: rust-lld: error: function signature mismatch: mrb_sys_class_of_value
          >>> defined as (i32, i64, i32, i32) -> i32 in /app/target/wasm32-unknown-unknown/debug/build/mruby-sys-7c7d5344d44d1927/out/libmrubysys.a(ext.o)
          >>> defined as (i32, i32) -> i32 in lto.tmp
```

The second parameter is an `mrb_value`. bindgen outputs the following code for `mrb_value`:

```rust
pub struct mrb_value {
    pub value: mrb_value__bindgen_ty_1,
    pub tt: mrb_vtype,
    pub __bindgen_padding_0: u32,
}
#[repr(C)]
#[derive(Copy, Clone)]
pub union mrb_value__bindgen_ty_1 {
    pub f: mrb_float,
    pub p: *mut ::std::os::raw::c_void,
    pub i: mrb_int,
    pub sym: mrb_sym,
    _bindgen_union_align: u64,
}
#[test]
fn bindgen_test_layout_mrb_value__bindgen_ty_1() {
    assert_eq!(
        ::std::mem::size_of::<mrb_value__bindgen_ty_1>(),
        8usize,
        concat!("Size of: ", stringify!(mrb_value__bindgen_ty_1))
    );
    assert_eq!(
        ::std::mem::align_of::<mrb_value__bindgen_ty_1>(),
        8usize,
        concat!("Alignment of ", stringify!(mrb_value__bindgen_ty_1))
    );
    assert_eq!(
        unsafe { &(*(::std::ptr::null::<mrb_value__bindgen_ty_1>())).f as *const _ as usize },
        0usize,
        concat!(
            "Offset of field: ",
            stringify!(mrb_value__bindgen_ty_1),
            "::",
            stringify!(f)
        )
    );
    assert_eq!(
        unsafe { &(*(::std::ptr::null::<mrb_value__bindgen_ty_1>())).p as *const _ as usize },
        0usize,
        concat!(
            "Offset of field: ",
            stringify!(mrb_value__bindgen_ty_1),
            "::",
            stringify!(p)
        )
    );
    assert_eq!(
        unsafe { &(*(::std::ptr::null::<mrb_value__bindgen_ty_1>())).i as *const _ as usize },
        0usize,
        concat!(
            "Offset of field: ",
            stringify!(mrb_value__bindgen_ty_1),
            "::",
            stringify!(i)
        )
    );
    assert_eq!(
        unsafe { &(*(::std::ptr::null::<mrb_value__bindgen_ty_1>())).sym as *const _ as usize },
        0usize,
        concat!(
            "Offset of field: ",
            stringify!(mrb_value__bindgen_ty_1),
            "::",
            stringify!(sym)
        )
    );
}
#[test]
fn bindgen_test_layout_mrb_value() {
    assert_eq!(
        ::std::mem::size_of::<mrb_value>(),
        16usize,
        concat!("Size of: ", stringify!(mrb_value))
    );
    assert_eq!(
        ::std::mem::align_of::<mrb_value>(),
        8usize,
        concat!("Alignment of ", stringify!(mrb_value))
    );
    assert_eq!(
        unsafe { &(*(::std::ptr::null::<mrb_value>())).value as *const _ as usize },
        0usize,
        concat!(
            "Offset of field: ",
            stringify!(mrb_value),
            "::",
            stringify!(value)
        )
    );
    assert_eq!(
        unsafe { &(*(::std::ptr::null::<mrb_value>())).tt as *const _ as usize },
        8usize,
        concat!(
            "Offset of field: ",
            stringify!(mrb_value),
            "::",
            stringify!(tt)
        )
    );
}
```

I'm not sure what's going on here, but I suspect this is a codegen error.

lopopolo:
Filed https://github.com/rust-lang/rust/issues/63626

lopopolo:
Filed https://github.com/rust-lang/rust/issues/63629 for the empty wasm bundle when building `mruby-sys`

lopopolo:
While we wait to hear back on the Rust bugs I've files, I'm working on merging parts of this branch into master.

Starting with GH-173 and will work on the onig 5.0 upgrade next.

lopopolo:
Upstream PR to rust-onig https://github.com/rust-onig/rust-onig/pull/115 to pull in onigurma patch https://github.com/kkos/oniguruma/pull/150

lopopolo:
Upstreaming of `MRB_API` config patch: https://github.com/mruby/mruby/pull/4643

lopopolo:
Updated this branch to pull in the PRs that have been extracted from it.

lopopolo:
Pulling in the `MRB_API` patch is in progress GH-178, but blocked on an upstream bug https://github.com/mruby/mruby/issues/4644

singpolyma-shopify:
Thanks for all your work on this!  I'll be following those rust bugs with interest

lopopolo:
This is now blocked on https://github.com/WebAssembly/tool-conventions/issues/88 which is potentially a year long effort.

The true path forward here is to remove the mruby C dependency which is aligned with current plans.

- implement Ruby core in Rust.
- migrate the parser to rust or use the whitequark parser.
- implement the VM in rust.
- remove explicit unwinds from the implementation of the VM and core.

lopopolo:
There’s a lot of work to do to achieve the goals in the last comment and keeping this PR open is unlikely to help.

Closing because this PR is stale.

PR #174, opened 2019-08-17
Update to onig 5.0

lopopolo:
onig 5.0 uses bindgen to generate bindings for oniguruma and is easier
to hack on. To upstream the changes in GH-172 that add
wasm32-unknown-unknown support to Artichoke, we need to build with onig
5.0.

In addition to updating the dependency, this commit updates the `Regexp`
and `MatchData` implementations to work with the new APIs exposed by
onig.

The only change is the removal of `Regex::capture_names` and the
addition of the `Regex::foreach_name` function that replaces it. This
changes the API from an iterator to a callback with a break parameter.
This change appears to be a result of upstream changes in oniguruma.

This commit does not include any of the wasm support present in our
fork: https://github.com/artichoke/rust-onig/tree/wasm3. This means that
once this commit is merged, artichoke will no longer build under
wasm32-unknown-emscripten. This is safe because the playground is pinned
to a specific version of artichoke-backend.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "winapi 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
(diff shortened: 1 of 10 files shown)


lopopolo:
Upstream PR to add back wasm32-unknown-unknown support https://github.com/rust-onig/rust-onig/pull/115

PR #177, opened 2019-08-19
Build with stable Rust 1.37.0

lopopolo:
Switch the Rust toolchain from nightly to stable 1.37.0.
Artichoke does not currently depend on any nightly features and I'd
like to keep it that way.


lopopolo:
This change sped up the build significantly. Build times on branches dropped from 10 minutes down to 5 minutes. 🏎

Issue #24, opened 2019-07-07
Revendor mruby and re-enable thread test when upstream bug is fixed


lopopolo:
The vendored version of mruby has a bug where in some situations comments do not increment the line number in the parser. This breaks tests which test stack traces. thread.rs has one of these tests that has been disabled. Pull in the upstream bugfix and re-enable the test.

Upstream bug: https://github.com/mruby/mruby/issues/4513

Disabled test:

https://github.com/lopopolo/ferrocarril/blob/329c544006b619b7b4fac9c96b6b5ec51a311a84/mruby/src/extn/core/thread.rs#L155-L194

Issue #146, opened 2019-08-07
Remove dependency on patch(1) binary in build.rs


lopopolo:
Some crates in Artichoke shell out to `patch(1)` via `bash(1)` in `build.rs` to patch sources. Replace this setup with the [patch crate](https://docs.rs/patch/) for a more self-contained build.

- [ ] `artichoke-backend`
  https://github.com/artichoke/artichoke/blob/6ad780b1a9899155bc7fe9a71a5b8eb3d16acc2f/artichoke-backend/build.rs#L103-L114
- [ ] `spec-runner`
  https://github.com/artichoke/artichoke/blob/6ad780b1a9899155bc7fe9a71a5b8eb3d16acc2f/spec-runner/build.rs#L39-L57
- [ ] `mruby-sys`
  https://github.com/artichoke/artichoke/blob/6ad780b1a9899155bc7fe9a71a5b8eb3d16acc2f/mruby-sys/build.rs#L95-L107

See GH-137, GH-55.

lopopolo:
The approach I ended up going with does not involve the patch crate.

Instead, fork each dependency we vendor and add patches to an `artichoke-vendor` branch. Vendor this branch directly and add docs explaining where the source of the vendored files lives.

PR #178, opened 2019-08-19
Update mruby

lopopolo:
Pull in the latest mruby master with https://github.com/mruby/mruby/pull/4643

This PR changes the vendoring approach. Until now, artichoke vendored upstream and applied patches at build time. Now, we fork upstream, apply patches on a branch in the fork, and vendor the fork. This approach is simpler, and will let us remove the dependency on `patch`. See GH-146.

This also fixes a bug in the parser where line number would not get incremented for some comments. Fixes GH-24.

This PR adds docs for where vendored mruby comes from.


lopopolo:
Spec failures are caused by upstream bug: https://github.com/mruby/mruby/issues/4644

lopopolo:
Blocked on GH-186.

Issue #186, opened 2019-08-20
Implement Integer#chr in Rust


lopopolo:
mruby https://github.com/artichoke/mruby/commit/c980fe279266448cf5ac8faa646d57d0f8561b10 moves `Interger#chr` into mruby-string-ext, which Artichoke no longer builds.

Implement `Integer#chr` in Rust.

Blocks GH-178.

Rust easily allows converting between Integers and characters with `i as char` syntax. Use a converter to extract a `u32` from the `Fixnum`.

PR #192, opened 2019-08-23
Implement Integer#chr in Rust

lopopolo:
Fixes GH-186.

This PR adds support for `Integer#chr` when called with no argument. `Integer#chr` with no argument requires that `self` convert to a `u8` and creates a `String` with a single byte `Vec<u8>`.


lopopolo:
Ran into a segfault when implementing this which I worked around with a4ffbe3.

Upstream report: https://github.com/mruby/mruby/issues/4660

PR #217, opened 2019-08-30
Remove dependency on mruby-proc-ext

tkbky:

### DIFF
--- a/artichoke-backend/src/extn/core/mod.rs
+++ b/artichoke-backend/src/extn/core/mod.rs
@@ -13,6 +13,7 @@ pub mod kernel;
 pub mod matchdata;
 pub mod module;
 pub mod numeric;
+pub mod proc;
 pub mod range;
 pub mod regexp;
 pub mod string;
@@ -31,6 +32,7 @@ pub fn init(interp: &Artichoke) -> Result<(), ArtichokeError> {
     matchdata::init(interp)?;
     module::init(interp)?;
     numeric::init(interp)?;
+    proc::init(interp)?;
     range::init(interp)?;
     regexp::init(interp)?;
     string::init(interp)?;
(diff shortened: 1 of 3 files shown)


tkbky:
@lopopolo 
I haven't remove `mruby-proc-ext` from `mruby-sys/build.rs` yet because `mruby-method` depends on it.

Same thing, methods in C are not implemented yet.
```
mrb_define_method(mrb, p, "lambda?",         mrb_proc_lambda,          MRB_ARGS_NONE());
mrb_define_method(mrb, p, "source_location", mrb_proc_source_location, MRB_ARGS_NONE());
mrb_define_method(mrb, p, "to_s",            mrb_proc_inspect,         MRB_ARGS_NONE());
mrb_define_method(mrb, p, "inspect",         mrb_proc_inspect,         MRB_ARGS_NONE());
mrb_define_method(mrb, p, "parameters",      mrb_proc_parameters,      MRB_ARGS_NONE());

mrb_define_class_method(mrb, mrb->kernel_module, "proc", mrb_kernel_proc, MRB_ARGS_NONE());
mrb_define_method(mrb, mrb->kernel_module,       "proc", mrb_kernel_proc, MRB_ARGS_NONE());
```

lopopolo:
`Proc` looks like it will be one of the trickier things to untangle from mruby, so I'm not too concerned that we can't get rid of this dependency yet. This is a wonderful start 💎 

I don't think `mruby-method` is required for anything in the passing spec suite right now, so maybe the right thing to do is remove it for now and add it back once we have moved the `Proc` implementation to Rust.

Merging this PR now 😄

lopopolo:
> I don't think `mruby-method` is required for anything in the passing spec suite right now, so maybe the right thing to do is remove it for now and add it back once we have moved the `Proc` implementation to Rust.

The URI package calls `Kernel.instance_method(:to_s)` in a few places, which means we can't remove `mruby-method` yet.

PR #218, opened 2019-08-30
Remove dependency on mruby-symbol-ext

tkbky:

### DIFF
--- a/artichoke-backend/src/extn/core/symbol/symbol.rb
+++ b/artichoke-backend/src/extn/core/symbol/symbol.rb
@@ -1,34 +1,39 @@
+# frozen_string_literal: true
+
 class Symbol
   include Comparable
 
   alias intern to_sym
 
   def capitalize
-    (self.to_s.capitalize! || self).to_sym
+    (to_s.capitalize! || self).to_sym
   end
 
   def downcase
-    (self.to_s.downcase! || self).to_sym
+    (to_s.downcase! || self).to_sym
   end
 
   def upcase
-    (self.to_s.upcase! || self).to_sym
+    (to_s.upcase! || self).to_sym
   end
 
   def casecmp(other)
-    return nil unless other.kind_of?(Symbol)
-    lhs =  self.to_s; lhs.upcase!
+    return nil unless other.is_a?(Symbol)
+
+    lhs = to_s
+    lhs.upcase!
     rhs = other.to_s.upcase
     lhs <=> rhs
   end
 
   def casecmp?(sym)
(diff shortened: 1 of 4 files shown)


tkbky:
The following C methods are not implemented.

```c
#ifdef MRB_ENABLE_ALL_SYMBOLS
  mrb_define_class_method(mrb, s, "all_symbols", mrb_sym_all_symbols, MRB_ARGS_NONE());
#endif
  mrb_define_method(mrb, s, "length", mrb_sym_length, MRB_ARGS_NONE());
  mrb_define_method(mrb, s, "size", mrb_sym_length, MRB_ARGS_NONE());
```

lopopolo:
> The following C methods are not implemented.
> 
> ```c
> #ifdef MRB_ENABLE_ALL_SYMBOLS
>   mrb_define_class_method(mrb, s, "all_symbols", mrb_sym_all_symbols, MRB_ARGS_NONE());
> #endif
> ```

mruby-sys does not define this macro so this method is already not implemented.

artichoke-backend/src/extn/core/symbol/symbol.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class Symbol
+  include Comparable
+
+  def capitalize
+    (to_s.capitalize! || self).to_sym
```
lopopolo: some of these implementations from mruby are a bit funky. For example, the docs for [`Symbol#capitalize`](https://ruby-doc.org/core-2.6.3/Symbol.html#method-i-capitalize) say the method is equivalent to:

```ruby
sym.to_s.capitalize.intern
```

I'd like to match the docs (even if the semantics are the same) if possible.

artichoke-backend/src/extn/core/symbol/symbol.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def downcase
+    (to_s.downcase! || self).to_sym
```
lopopolo: docs say this is equivalent to

```ruby
sym.to_s.downcase.intern
```

artichoke-backend/src/extn/core/symbol/symbol.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def casecmp?(sym)
+    c = casecmp(sym)
+    return nil if c.nil?
+
+    c.zero?
+  end
+
+  def empty?
+    to_s.empty?
```
lopopolo: docs say to check for equality with `:''`

Issue #223, opened 2019-08-31
Implement Struct class in Rust


lopopolo:
Implement `Struct` in Rust in terms of a `Vec<(name, value)>` or a `HashMap<name, value>`.

lopopolo:
This is required to be able to run `optcarrot`.

Issue #222, opened 2019-08-31
Implement Math module in Rust


lopopolo:
This module is a collection of pure functions. It should be straightforward to extract numbers with converters, call math functions from `std`, and wrap the result in a `Value`.

cdeler:
@lopopolo I would be happy, if I can take it

lopopolo:
Thanks @cdeler it's all yours 💎

lopopolo:
Fixed in GH-559.

Issue #34, opened 2019-07-08
Implement Kernel#rand and SecureRandom in Rust


lopopolo:
Remove dependency on `mruby-random` mrbgem.

lopopolo:
We should be able to use the [rand crate](https://docs.rs/rand) to implement this in pure Rust.

Elang89:
If this hasn't been resolved yet, I'd like to try my luck with it.

lopopolo:
It's all yours @Elang89 

One gotcha is you'll need to remove the `SecureRandom` stub:

https://github.com/artichoke/artichoke/blob/3d715c9a22f22b78460e6681bf634e85dfb7e62b/artichoke-backend/src/extn/stdlib/stubs.rs#L12-L15

lopopolo:
We'll also probably need the `uuid` crate with the `v4` feature.

lopopolo:
It looks like this API is implemented in Ruby and selects a backend such as openssl. I think we can do better by implementing in pure Rust and avoiding an openssl dependency.

https://github.com/artichoke/artichoke/blob/3d715c9a22f22b78460e6681bf634e85dfb7e62b/artichoke-backend/vendor/ruby/lib/securerandom.rb

lopopolo:
Kernel#rand was implemented in GH-384.

Issue #33, opened 2019-07-08
Implement Time class in Ruby Core


lopopolo:
In MRI, `Time` is both a [Core class](https://ruby-doc.org/core-2.6.3/Time.html) and a [Standard Library package](https://ruby-doc.org/stdlib-2.6.3/libdoc/time/rdoc/Time.html).

The Core class is implemented in C, see [`time.c` in CRuby](https://github.com/ruby/ruby/blob/v2_6_3/time.c). Reimplement this class in Rust with [httpdate](https://docs.rs/httpdate/) and [chrono](https://docs.rs/chrono/).

The Standard Library package is pure Ruby and can likely be imported as is once the Core class is in place. See GH-158 for an example of how to do this.

Remove dependency on `mruby-time` mrbgem.

vaibhav-y:
I'm familiar with C and the ruby-c-api from some time back, would you say this is a good issue to start with, given that libraries to handle time exist / the code to port is reasonable?

Also, how does one get started contributing to this repo? I'm picking up rust as a new language, I come from a strong C++ background.

lopopolo:
Hi @vaibhav-y 👋 I would love to leverage your experience to help implement Artichoke.

I've added a [`CONTRIBUTING.md`](https://github.com/artichoke/artichoke/blob/master/CONTRIBUTING.md) which outlines how to get set up so you can build the code and run the tests.

The time package in MRI has both C and Ruby pieces to it. Artichoke has implemented `Regexp` already which [follows a similar pattern](https://github.com/artichoke/artichoke/blob/master/artichoke-backend/src/extn/core/regexp/mod.rs).

For handling the time bits, we can probably use some combination of [httpdate](https://docs.rs/httpdate/) and [chrono](https://docs.rs/chrono/).

One thing Artichoke does that is maybe _non-obvious Rust_ is make heavy use of `Rc` and interior mutability with `RefCell`. You can check out [`manual.rs`](https://github.com/artichoke/artichoke/blob/189961ad97b9fbda1e3341fe0ede5495b2d26c2d/artichoke-backend/tests/manual.rs) for a self-contained example of how this works.

I hope this is enough to get started and I'm happy to answer any questions you have along the way.  I'm excited to have you contribute 💎

vaibhav-y:
[...]

PR #224, opened 2019-08-31
Purge mrbgems

lopopolo:
@tkbky and @lopopolo had a discussion in GH-217 about whether some mrbgems could be removed from the build.

The set of gems in the build was chosen for [artichoke/ferrocarril](https://github.com/artichoke/ferrocarril) which targets running Sinatra.

Our goals in Artichoke are to pass the specs we have implemented. Incompleteness in the Ruby core implementation is acceptable.

This PR removes the following gems for the build and gives their tracking tickets:

- mruby-math => GH-222
- mruby-time => GH-33
- mruby-struct => GH-223
- mruby-random => GH-34


lopopolo:
This PR also adds docs for what each gem's reason is for inclusion in the build.

Issue #221, opened 2019-08-31
Parallelize Ruby stdlib Rust glue generation


lopopolo:
Processing each package in the Ruby standard library and generating the Rust glue makes the artichoke-backend build a bit slow.

This problem is embarrassingly parallel and is free from filesystem race since the outputs are to content addressable locations.

We might be able to use [rayon](https://github.com/rayon-rs/rayon#parallel-iterators-and-more) to parallelize this part of the build with very little effort.

The `cc` crate uses rayon in a similar way to parallelize invoking a C compiler: https://github.com/alexcrichton/cc-rs/blob/e2f4692b1657d2bb1b8eee725a1677971b88dc4d/src/lib.rs#L946-L971.

lopopolo:
@johnwahba do you want to take a look at this ticket?

PR #237, opened 2019-08-31
parallel ruby build for artichoke-backend

johnwahba:
Build packages in parallel rather than iteratively. This takes the build time on my laptop from 30 sec => 16 sec

### DIFF
--- a/artichoke-backend/build.rs
+++ b/artichoke-backend/build.rs
@@ -3,6 +3,7 @@
 #![doc(deny(warnings))]
 
 use fs_extra::dir;
+use rayon::prelude::*;
 use std::env;
 use std::fs;
 use std::path::PathBuf;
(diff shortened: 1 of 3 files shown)


johnwahba:
fixes https://github.com/artichoke/artichoke/issues/221

artichoke-backend/build.rs
```diff
@@ -195,14 +196,16 @@ fn main() {
...
-    ] {
+    ];
+
+    packages.par_iter().for_each(|package| {
         let sources = Build::get_package_files(package)
             .trim()
             .split("\n")
             .filter(|s| !s.is_empty())
             .map(String::from)
             .collect::<Vec<_>>();
```
lopopolo: I don't think we need to collect into a `Vec<_>` anymore since we're consuming the iterator directly with rayon.

lopopolo: This might be possible with `ParallelBridge`, but the docs say this is less efficient than using a `Vec`. Our list is small so it makes sense to pay the cost to collect eagerly.

Issue #156, opened 2019-08-11
Implement the rest of the ENV API in Ruby


lopopolo:
Follow up to GH-148.

GH-148 implements `ENV::[]`, `ENV::[]=` and `ENV::to_h` in Rust. The [rest of the API](https://ruby-doc.org/core-2.6.3/ENV.html) can be implemented in Ruby in terms of these primitives.

cdeler:
@lopopolo May I take this one?

lopopolo:
@cdeler yes it's all yours.

cdeler:
@lopopolo If it's possible, could you assign the issue to me? It could help me to track it.

lopopolo:
Done.

cdeler:
@lopopolo Please could you pay an attention to my local branch? ([link](https://github.com/artichoke/artichoke/compare/master...cdeler:%23156-implement-rest-ENV-API-in-ruby?expand=1))

I'm not familiar with ruby, am I doing well?

lopopolo:
Hi @cdeler this is looking really good!

The goal is to implemented the [`ENV` API for Ruby 2.6.3](https://ruby-doc.org/core-2.6.3/ENV.html).

As with the last PR, I'd like to keep this incremental. Maybe address the feedback below and submit a PR with your progress so far? Thanks again for your help 💎 

## API Coverage

- [x] ::[]
- [x] ::[]=
- [x] ::assoc
- [x] ::clear
- [x] ::delete
- [ ] ::delete_if
- [ ] ::each
- [ ] ::each_key
- [ ] ::each_pair
- [ ] ::each_value
- [x] ::empty?
- [ ] ::fetch
- [ ] ::filter
- [ ] ::filter!
- [x] ::has_key?
- [x] ::has_value?
- [x] ::include?
- [ ] ::index
- [ ] ::inspect
- [ ] ::invert
- [ ] ::keep_if
- [x] ::key
- [ ] ::key?
- [ ] ::keys
- [x] ::length
- [ ] ::member?
- [ ] ::rassoc
- [x] ::rehash
- [ ] ::reject
- [ ] ::reject!
- [ ] ::replace
- [ ] ::select
- [ ] ::select!
- [x] ::shift
- [x] ::size
- [x] ::slice
- [ ] ::store
- [x] ::to_a
- [x] ::to_h
- [x] ::to_hash
- [x] ::to_s
- [x] ::update
- [x] ::value?
- [x] ::values
- [x] ::values_at

[...]

PR #239, opened 2019-09-01
#156 implement rest env api in ruby

cdeler:
@lopopolo I had come back from vacation and fixed your remarks (from [this](https://github.com/artichoke/artichoke/issues/156#issuecomment-521088939) comment).

### DIFF
--- a/artichoke-backend/src/extn/core/env/env.rb
+++ b/artichoke-backend/src/extn/core/env/env.rb
@@ -19,17 +19,25 @@ class EnvClass
 
     def delete(name)
         value = self[name]
-        self[name] = nil
+        return nil if value.nil?
 
-        if block_given? and value.nil?
-            yield name
-        end
+        self[name] = nil
+        yield name if block_given?
 
         value
     end
 
+    def delete_if
+        return to_enum(:delete_if) unless block_given?
+      
+        each do |key, value|
+          delete(key) if yield key, value
+        end
+        to_h
+    end
+
     def empty?
-        return self.to_h.empty?
+        self.to_h.empty?
     end
 
     def has_key?(name)
@@ -94,15 +102,13 @@ class EnvClass
 
     def shift 
         envs = self.to_h
-        
-        a_pair = envs.shift
 
-        if a_pair.nil?
-            return nil
-        else
-            self[a_pair[0]] = nil
-            return a_pair
-        end
+        return nil if envs.nil?
+        
+        name, value = envs.shift
+
+        self[name] = nil
+        return [name, value]
     end
 
     def update(hash)
--- a/artichoke-backend/src/extn/core/env/env.rb
+++ b/artichoke-backend/src/extn/core/env/env.rb
@@ -1,9 +1,35 @@
 # frozen_string_literal: true
 
 class EnvClass
+    def assoc(name)
+        value = self[name]
+
+        return value.nil? ? nil :  [name, self[name]]
+    end
+
+    def clear
+        self.to_h.each do |var_name, var_value|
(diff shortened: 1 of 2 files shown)


artichoke-backend/src/extn/core/env/env.rb
```diff
@@ -1,3 +1,119 @@
...
+  end
+
+  def to_hash
+    to_h
+  end
+
+  def shift
+    envs = to_h
+
+    return nil if envs.nil?
```
lopopolo: don't we also want to return nil if `envs.empty?`

artichoke-backend/src/extn/core/env/env.rb
```diff
@@ -1,3 +1,119 @@
 # frozen_string_literal: true
 
-ENV = EnvClass.new
+class EnvClass
+  def assoc(name)
+    value = self[name]
+
+    value.nil? ? nil : [name, self[name]]
```
lopopolo: can we structure this as:

```ruby
return nil if value.nil?
[name, value]
```

artichoke-backend/src/extn/core/env/env.rb
```diff
@@ -1,3 +1,119 @@
...
+    self[name] = nil
+    yield name if block_given?
+
+    value
+  end
+
+  def delete_if
+    return to_enum(:delete_if) unless block_given?
+
+    each do |key, value|
```
lopopolo: Are we missing a definition of `each`?

`clear` calls `to_h.each`. Should we be doing that here?

artichoke-backend/src/extn/core/env/env.rb
cdeler: > Removes an environment variable name-value pair from ENV and returns it as an Array. Returns nil if when the environment is empty.
(c) https://ruby-doc.org/core-2.6/ENV.html#method-c-shift

lopopolo: Does `to_h` on an empty `ENV` return nil or an empty hash? An empty hash is not `nil`.

artichoke-backend/src/extn/core/env/env.rb
cdeler: I'm sorry, I'm not familiar with ruby (and didn't check it), it's my fault.

artichoke-backend/src/extn/core/env/env.rb
lopopolo: oops didn't see the thumbs up 😅

lopopolo:
I've updated the checklist with the current state https://github.com/artichoke/artichoke/issues/156#issuecomment-521088939

artichoke-backend/src/extn/core/env/env.rb
cdeler: I'm sorry, I did agree with you, I've rewritten this line

artichoke-backend/src/extn/core/env/env.rb
```diff
@@ -1,3 +1,121 @@
 # frozen_string_literal: true
 
-ENV = EnvClass.new
+class EnvClass
+  def assoc(name)
+    value = self[name]
+
+    return nil if value.nil?
+
+    [name, self[name]]
```
lopopolo: @cdeler this comment might have gotten lost in the last review. Can we replace the call to `self[name]` here with the value we already pulled out of the env?

I'm worried about a race as we call back into Rust as Artichoke gains support for multi-threading.

cdeler: Done

PR #240, opened 2019-09-01
Remove Kernel::NOT_SET and Kernel#not_set?

lopopolo:
This API and constant were added in artichoke/ferrocarril#85 to load Sinatra and
work around mruby not supporting expressions as default parameters in method argument
lists.

This patch belongs in nemesis instead of artichoke core.


lopopolo:
It also appears this behavior changed since the last time I looked. With mruby@artichoke-vendor:

```
> def foo(bar = (not_set = true)); puts not_set.inspect; end
 => :foo
> foo
true
 => nil
> foo('a')
nil
 => nil
```

PR #245, opened 2019-09-02
#156 implement rest env api in ruby v2

cdeler:
@lopopolo another small piece of changes

### DIFF
--- a/artichoke-backend/src/extn/core/env/env.rb
+++ b/artichoke-backend/src/extn/core/env/env.rb
@@ -72,6 +72,40 @@ class EnvClass
     nil
   end
 
+  def replace(hash)
+    hash.each do |k, v|
+      self[k] = v
+    end
+
+    select! { |k, _| hash.key?(k) }
+  end
+
+  def select
+    return to_enum(:select) unless block_given?
+
+    to_h.select { |key, value| yield key, value }
+  end
+
+  def select!
+    return to_enum(:select!) unless block_given?
+
+    env = to_h
+
+    # collect all the keys where the block evaluates to false
+    to_remove = env.reject do |key, value|
+      yield key, value
+    end
+
+    # returns nil if no changes were made
+    return nil if to_remove.empty?
+
+    to_remove.each do |key, _|
+      delete(key)
+    end
+
+    self
+  end
+
   def shift
     envs = to_h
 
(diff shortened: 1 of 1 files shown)


artichoke-backend/src/extn/core/env/env.rb
```diff
@@ -52,63 +52,94 @@ def include?(name)
...
 
-  def to_a
-    to_h.to_a
+  def size
+    to_h.size
   end
 
-  def to_s
-    'ENV'
+  def replace(hash)
```
lopopolo: looks like this method ended up unalphabetized

artichoke-backend/src/extn/core/env/env.rb
```diff
@@ -52,63 +52,94 @@ def include?(name)
...
+    return to_enum(:select!) unless block_given?
+
+    old_env = to_h
+
+    new_envs = old_env.select { |key, value| yield key, value }
+
+    # returns nil if no changes were made
+    return nil if new_envs == old_env
+
+    clear
```
lopopolo: there is a race here where the env is empty for a bit. The docs say that this method should delete the keys that are not present in new env.

> Deletes every environment variable where the block evaluates to false.

Does this work?

```ruby
env = to_h
# collect all the keys where the block evaluates to false
to_remove = env.reject do |key, value|
  yield key, value
end

return nil if to_remove.empty?

to_remove.each do |key, _|
  delete(key)
end

self
```

This method can be aliased to `filter!`.
This method body should be used to implement `keep_if` and `select!` should be implemented with `keep_if`.

artichoke-backend/src/extn/core/env/env.rb
```diff
@@ -52,63 +52,94 @@ def include?(name)
...
-  def to_a
-    to_h.to_a
+  def size
+    to_h.size
   end
 
-  def to_s
-    'ENV'
+  def replace(hash)
+    clear
```
lopopolo: I'd rather implement this as removing the keys that are not in `hash` and then overwriting the the env with the keys in `hash` to avoid temporarily having an empty `ENV`.

cdeler: Technically all ENV methods should be protected by mutex, as ENV variables are shared state. From my point of view, they should be changed synchronously.

artichoke-backend/src/extn/core/env/env.rb
cdeler: I reordered ENV's methods

artichoke-backend/src/extn/core/env/env.rb
lopopolo: See GH-253.

PR #254, opened 2019-09-03
Remove dependency on mruby-kernel-ext

tkbky:

### DIFF
--- a/artichoke-backend/src/extn/core/kernel/kernel.rb
+++ b/artichoke-backend/src/extn/core/kernel/kernel.rb
@@ -68,13 +68,14 @@ module Kernel
       return ret if ret.is_a?(Hash)
 
       raise TypeError, "can't convert #{classname} into Hash" if ret.nil?
+
       raise TypeError, "can't convert #{classname} to Hash (#{classname}#to_hash gives #{ret.class})"
     end
 
     raise TypeError, "can't convert #{classname} into Hash"
   end
 
-  def Integer(arg, base = 0, exception: true) # rubocop:disable Naming/MethodName
+  def Integer(arg, base = 0) # rubocop:disable Naming/MethodName
     raise ArgumentError, 'base specified for non string value' if base.positive? && arg.is_a?(Numeric)
 
     classname = arg.class
--- a/artichoke-backend/src/extn/core/kernel/kernel.rb
+++ b/artichoke-backend/src/extn/core/kernel/kernel.rb
@@ -40,6 +40,26 @@ module Kernel
     e.value
   end
 
+  # Placeholder for Array spec to pass
+  def Integer(arg, base=0, exception: true)
+    return arg.to_int if arg.respond_to?(:toint)
+    return arg.to_i if arg.respond_to?(:to_i)
+    0
+  end
+
+  def Array(arg) # rubocop:disable Naming/MethodName
+    return arg if arg.is_a?(Array)
+
+    ret = nil
+    ret = arg.to_ary if arg.respond_to?(:to_ary)
+    ret = arg.to_a if ret.nil? && arg.respond_to?(:to_a)
+
+    return [arg] if ret.nil?
+    return ret if ret.is_a?(Array)
+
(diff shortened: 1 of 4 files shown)


artichoke-backend/src/extn/core/kernel/kernel.rb
```diff
@@ -40,6 +40,221 @@ def catch(tag = nil)
...
+      # no trailing -, +, i.e. "1-", "1++"
+      # no trailing _
+      # TODO: no trailing null byte \0
+      k = arg.length - 1
+      # squeeze trailing spaces
+      k += 1 while arg[k] == ' '
+      raise ArgumentError, "invalid value for Integer(): \"#{arg}\"" if arg[k] == '-' || arg[k] == '+' || arg[k] == '_'
+
+      # when base is not specified or zero, check the base letter
+      if base <= 0
```
lopopolo: Oh no 😱 This looks very complicated and hard to get right.

Now I understand why you were asking about the split between Rust and Ruby. In this case, Rust has a library function that will do all of this hard work for us in a much better tested way: [`i64::from_str_radix`](https://doc.rust-lang.org/std/primitive.i64.html#method.from_str_radix).

I've mocked up an example parsing function here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=65200add357dd7d7d87b15e44db8dd62

artichoke-backend/src/extn/core/kernel/mod.rs
```diff
@@ -175,6 +176,24 @@ impl Kernel {
         sys::mrb_sys_nil_value()
     }
 
+    // unsafe extern "C" fn raise(mrb: *mut sys::mrb_state, _slf: sys::mrb_value) -> sys::mrb_value {
```
lopopolo: I'm guessing this is part of the WIP of this PR?

scripts/spec-compliance.sh
```diff
@@ -20,10 +20,20 @@ run_library_spec() {
   popd >/dev/null
 }
 
-run_core_spec "comparable"
-run_core_spec "matchdata"
-run_core_spec "regexp"
+run_kernel_spec() {
```
lopopolo: 😍 love this

tkbky: whoops.. 😅

tkbky:
@lopopolo 

Just back from vacation. I've made some changes to the error handling as suggested.

Few questions:
1. Do you think I should remove `Kernel#Integer` for now?
2. Should I put `mruby-kernel-ext` back in `mruby-sys/build.rs` until all methods in it are implemented?

artichoke-backend/src/extn/core/kernel/kernel.rb
```diff
@@ -68,13 +68,14 @@ def Hash(arg) # rubocop:disable Naming/MethodName
...
 
       raise TypeError, "can't convert #{classname} into Hash" if ret.nil?
+
       raise TypeError, "can't convert #{classname} to Hash (#{classname}#to_hash gives #{ret.class})"
     end
 
     raise TypeError, "can't convert #{classname} into Hash"
   end
 
-  def Integer(arg, base = 0, exception: true) # rubocop:disable Naming/MethodName
```
lopopolo: this is ok since we're planning on ripping this impl out and pulling it into Rust.

mruby-sys/build.rs
```diff
@@ -37,7 +37,6 @@ impl Build {
             "mruby-sprintf",      // Kernel#sprintf, Kernel#format, String#%
             "mruby-class-ext",    // Pending removal, see GH-32
             "mruby-proc-ext",     // required by mruby-method, see GH-32
-            "mruby-struct",
```
lopopolo: 💪

tkbky:
@lopopolo 

> If the specs pass, I think we're fine to remove the gem. Would you mind adding one ticket for each method we need to implement and whether you think they can be done in Ruby or need to be implemented in Rust?

Sure, will do. 😃

Issue #200, opened 2019-08-27
Add macro to improve ArgSpec ergonomics


kenaniah:
When defining methods from within Rust to be attached to Ruby classes / modules, parsing arguments can be a bit of a pain. For example:

```rust
...
impl Hash {
    unsafe extern "C" fn assoc(mrb: *mut sys::mrb_state, slf: sys::mrb_value) -> sys::mrb_value {

        // Read 1 object argument
        let interp = unwrap_interpreter!(mrb);
        let argspec = b"o\0";
        let mut obj = <mem::MaybeUninit<sys::mrb_value>>::uninit();
        sys::mrb_get_args(
            mrb,
            argspec.as_ptr() as *const i8,
            obj.as_mut_ptr()
        );
        let obj = Value::new(&interp, obj.assume_init());
        // ... now we can do something with the argument

```

It would be nice if a macro could be used for argument parsing instead:
```rust
impl Hash {
    unsafe extern "C" fn assoc(mrb: *mut sys::mrb_state, slf: sys::mrb_value) -> sys::mrb_value {

        // Read 1 object argument
        let args: Vec<Value> = get_args!(mrb, "o");
```

Such a macro could take a string representing the [mruby-compatible argspec](https://github.com/mruby/mruby/blob/6375639/src/class.c#L548) and return a vector of `Value`s matching the arguments that were passed in. A call to a `Value`'s `.inner` allows one to work with the underlying `mrb_value` if need be.

Issue #243, opened 2019-09-02
Panic when calling Rust-implemented methods in core with args that do not match argspec


lopopolo:
```
>>> match_data = /(?<a>.)(.)(?<b>\d+)(\d)/.match("TñX1138.")
=> #<MatchData "ñX1138" a:"ñ" b:"113">
>>> match_data.end
thread 'main' panicked at 'already borrowed: BorrowMutError', src/libcore/result.rs:999:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /Users/vsts/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /Users/vsts/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:200
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:214
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:477
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:384
   8: rust_begin_unwind
             at src/libstd/panicking.rs:311
   9: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  10: core::result::unwrap_failed
             at /rustc/eae3437dfe991621e8afdc82734f4a172d7ddf9b/src/libcore/macros.rs:18
  11: core::result::Result<T,E>::expect
             at /rustc/eae3437dfe991621e8afdc82734f4a172d7ddf9b/src/libcore/result.rs:827
  12: core::cell::RefCell<T>::borrow_mut
             at /rustc/eae3437dfe991621e8afdc82734f4a172d7ddf9b/src/libcore/cell.rs:869
  13: artichoke_backend::value::ValueLike::funcall
             at artichoke-backend/src/value.rs:116
  14: <artichoke_backend::Artichoke as artichoke_backend::exception::ExceptionHandler>::last_error
             at artichoke-backend/src/exception.rs:97
[...]

PR #258, opened 2019-09-05
Add arg extraction macro to facilitate safe unwinds from sys::mrb_get_args

lopopolo:
Add a macro for extracting arguments from an mruby interpreter …
mruby calls `mrb_raise` which unwinds via longjmp if the args passed to
a function do not match the spec passed to mrb_get_args instead of
validating the argspecs declared on function register before delegating
to our Rust implementation. See mruby/mruby#4688.

This is not ideal because unwinds via longjmp do not let Rust clean up
resources (like `RefCell` borrows). This causes bugs like GH-243.

This commit introduces a macro that extracts arguments from the
mrb_state for a predefined set of argspecs. This code uses the argc
returned from `mrb_get_args` to determine which `MaybeUninit`s are
initialized.

This macro is intended to be called before unwrapping the interpreter
from the mrb_state. At this point all stack variables are `Copy` so
unwinding is safe.

Fixes GH-200.

Convert Kernel to use arg extraction macro

Partial fix for GH-243.


lopopolo:
cc @kenaniah

PR #265, opened 2019-09-05
#156 implement rest env api in ruby v3

cdeler:
@lopopolo Yet another piece of ENV object has been done

### DIFF
--- a/artichoke-backend/src/extn/core/env/env.rb
+++ b/artichoke-backend/src/extn/core/env/env.rb
@@ -126,6 +126,10 @@ class EnvClass
     to_h.key(value)
   end
 
+  def key?(name)
+    ! self[name].nil?
+  end
+
   def keys
     to_h.keys
   end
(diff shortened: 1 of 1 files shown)


artichoke-backend/src/extn/core/env/env.rb
```diff
@@ -36,10 +36,58 @@ def delete_if
...
+  def fetch(name, default = nil)
+    value = self[name]
+
+    return value unless value.nil?
+
+    return yield name if block_given?
+
+    return default unless default.nil?
+
+    raise KeyError, 'Variable ' + name + " doesn't exist"
```
lopopolo: can we convert this to use `String` interpolation?

It also appears MRI uses a different error message: `"key not found: #{name}"`

```
[2.6.3] > ENV.fetch('aaaaaaaa')
        3: from (irb):11
        2: from (irb):11:in `rescue in irb_binding'
        1: from (irb):11:in `fetch'
KeyError (key not found: "aaaaaaaa")
```

cdeler: Sure, I've fixed a remark

artichoke-backend/src/extn/core/env/env.rb
```diff
@@ -36,10 +36,58 @@ def delete_if
...
   end
 
+  def fetch(name, default = nil)
+    value = self[name]
+
+    return value unless value.nil?
+
+    return yield name if block_given?
+
+    return default unless default.nil?
```
lopopolo: ENV.fetch with an explicit nil default returns nil for a not found key.

One way we can implement this is with an expression as the default value:

```ruby
def fetch(name, default = (not_set = true))
  # ...
end
```

If no default is supplied, the expression is evaluated. Then we can `return default if not_set.nil?`

cdeler: Marvelous. Ruby is so tricky :-)

Fixed and compared with the default `irb` implementation.

PR #266, opened 2019-09-05
Implement Range#last in Ruby

### LINKED ISSUE #208: Implement Range#last
https://ruby-doc.org/core-2.6.3/Range.html#method-i-last

Not sure if we can do this in Ruby.

The mruby implementation might be a good starting point.

https://github.com/artichoke/artichoke/blob/51f393863b93405a2ae2d4fdb34b80a2eb95055f/mruby-sys/vendor/mruby/mrbgems/mruby-range-ext/src/range.c#L84-L96

tkbky:
Fixes #208

### DIFF
--- a/artichoke-backend/src/extn/core/range/range.rb
+++ b/artichoke-backend/src/extn/core/range/range.rb
@@ -22,10 +22,12 @@ class Range
   def last(*args)
     return self.end if args.empty?
 
-    raise ArgumentError, "wrong number of arguments (given #{args.length}, expected 1)" unless args.length == 1
+    raise ArgumentError, "wrong number of arguments (given #{args.length}, expected #{inspect})" unless args.length == 1
 
     arg = args[0]
-    raise TypeError, "no implicit conversion of #{arg.class} into Integer" unless arg.respond_to?(:to_int)
+    classname = arg.class
+    classname = arg.inspect if arg.nil? || arg.equal?(false) || arg.equal?(true)
+    raise TypeError, "no implicit conversion of #{classname} into Integer" unless arg.respond_to?(:to_int)
 
     n = arg.to_int
     raise TypeError, "can't convert #{arg.class} to Integer (#{arg.class}#to_int gives #{n.class})" unless n.is_a?(Integer)
--- a/artichoke-backend/src/extn/core/range/range.rb
+++ b/artichoke-backend/src/extn/core/range/range.rb
@@ -31,7 +31,7 @@ class Range
     raise TypeError, "can't convert #{arg.class} to Integer (#{arg.class}#to_int gives #{n.class})" unless n.is_a?(Integer)
 
     array = to_a
-    array.last(n.to_int)
+    array.last(n)
   end
 
   def max(&block)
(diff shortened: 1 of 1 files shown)


artichoke-backend/src/extn/core/range/range.rb
```diff
@@ -19,6 +19,21 @@ def first(*args)
     ary
   end
 
+  def last(*args)
+    return self.end if args.empty?
+
+    raise ArgumentError, "wrong number of arguments (given #{args.length}, expected 1)" unless args.length == 1
+
+    arg = args[0]
+    raise TypeError, "no implicit conversion of #{arg.class} into Integer" unless arg.respond_to?(:to_int)
```
lopopolo: this raise does the same thing with immediate objects like `nil` where it prints the value of `#inspect1 directly.

```
[2.6.3] > (0...1).last(nil)
        5: from /usr/local/var/rbenv/versions/2.6.3/bin/irb:23:in `<main>'
        4: from /usr/local/var/rbenv/versions/2.6.3/bin/irb:23:in `load'
        3: from /usr/local/var/rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        2: from (irb):4
        1: from (irb):4:in `last'
TypeError (no implicit conversion from nil to integer)
```

artichoke-backend/src/extn/core/range/range.rb
```diff
@@ -19,6 +19,21 @@ def first(*args)
     ary
   end
 
+  def last(*args)
+    return self.end if args.empty?
+
+    raise ArgumentError, "wrong number of arguments (given #{args.length}, expected 1)" unless args.length == 1
```
lopopolo: this should be `expected (0..1)`

```
[2.6.3] > (0...1).last(nil, nil)
        3: from (irb):5
        2: from (irb):5:in `rescue in irb_binding'
        1: from (irb):5:in `last'
ArgumentError (wrong number of arguments (given 2, expected 0..1))
```

artichoke-backend/src/extn/core/range/range.rb
lopopolo: Also for booleans

tkbky: whoops, totally miss that. good call out. thanks.

artichoke-backend/src/extn/core/range/range.rb
```diff
@@ -22,10 +22,12 @@ def first(*args)
   def last(*args)
     return self.end if args.empty?
 
-    raise ArgumentError, "wrong number of arguments (given #{args.length}, expected 1)" unless args.length == 1
+    raise ArgumentError, "wrong number of arguments (given #{args.length}, expected #{inspect})" unless args.length == 1
```
lopopolo: `inspect` prints debug information for the range object. I think we want the literal string `(0..1)`.

tkbky:
It just my silly mistake. 😄

PR #267, opened 2019-09-05
Implement Range#cover? in Ruby

### LINKED ISSUE #207: Implement Range#cover?
https://ruby-doc.org/core-2.6.3/Range.html#method-i-cover-3F

Not sure if we can do this in Ruby.

The mruby implementation might be a good starting point.

https://github.com/artichoke/artichoke/blob/51f393863b93405a2ae2d4fdb34b80a2eb95055f/mruby-sys/vendor/mruby/mrbgems/mruby-range-ext/src/range.c#L42-L66

tkbky:
Fixes #207 

Still fail the following 1 spec.

https://github.com/artichoke/artichoke/blob/44f1e83473decc16332eb6632e87743032dee6ac/spec-runner/vendor/spec/core/range/shared/cover_and_include.rb#L22-L29

### DIFF
--- a/artichoke-backend/src/extn/core/range/range.rb
+++ b/artichoke-backend/src/extn/core/range/range.rb
@@ -5,27 +5,22 @@ class Range
     # Failed tests:
     # Range#cover? compares values using <=>
     raise ArgumentError, "wrong number of arguments (given #{args.length}, expected 1)" unless args.length == 1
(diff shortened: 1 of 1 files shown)


lopopolo:
@tkbky this failing spec is a limitation of mruby. This will pass once we implement `Range` in Rust.

PR #277, opened 2019-09-08
Unbreak format-text.sh

lopopolo:
The linter was silently failing because there are no CSS sources in
artichoke. Explicitly set the parser so yarn doesn't choke on empty
lists of file inputs.


lopopolo:
cc @Elang89

PR #274, opened 2019-09-07
Add load function

Elang89:
Adds in support for Ruby load function.

PR #286, opened 2019-09-17
Remove Vec from load_impl

lopopolo:
load_impl had a Vec of file Paths that it looped over, despite only ever
trying one Path. This is a leftover from using require_impl as a base.

This commit removes the Vec and removes the for loop over it.

cc @Elang89


lopopolo:
this is a followup to https://github.com/artichoke/artichoke/pull/274

PR #288, opened 2019-09-21
Implement Array in Rust

lopopolo:
Replace the mruby-defined `Array` with an `Array` backed by a `VecDeque`.

This patch pulls in https://github.com/artichoke/mruby/commit/e32f4a060bde9e923d241eae8635dec2db8cf2e4 which makes modifications to mruby core (including the VM) to support swapping the `Array` impl.

PR #293, opened 2019-10-06
[WIP] Implement Kernel#Integer in Rust

### LINKED ISSUE #282: Implement Kernel#Integer in Rust
`Kernel#Integer` is a complex converter from arbitrary Ruby objects to `Integer`s, and includes String radix parsing.

@tkbky worked on a pure Ruby implementation https://github.com/artichoke/artichoke/pull/254/commits/3b218c645333cea68adb37099d2db844b2cf82e8 that is pretty complicated and seems hard to exhaustively test.

Rust std has a standard library function on numeric primitives called [`from_str_radix`](https://doc.rust-lang.org/std/primitive.i64.html#method.from_str_radix) which handles all of the parsing, overflow detection, and error handling for us.

I've mocked up a parser in Rust [[playground]](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=65200add357dd7d7d87b15e44db8dd62):

```rust
fn parse(s: &str, radix: Option<u32>) -> Result<i64, &'static str> {
    let mut buf = String::new();
    let mut parsed_radix = None;
    let s = if s.starts_with('-') {
        buf.push('-');
        &s[1..]
    } else {
        s
    };
    match &s[0..2] {
        "0b" | "0B" => {
            buf.push_str(&s[2..]);
            parsed_radix = Some(2);
        }
        "0o" | "0O" => {
            buf.push_str(&s[2..]);
            parsed_radix = Some(8);
        }
        "0x" | "0X" => {
            buf.push_str(&s[2..]);
            parsed_radix = Some(16);
        }
        prefix if &prefix[0..1] == "0" => {
            buf.push_str(&s[1..]);
            parsed_radix = Some(8);
        }
        _ => {}
    };
    match (radix, parsed_radix) {
        (Some(_), Some(_)) => Err("invalid value for Integer(): \"xxx\""),
        (Some(radix), None) => i64::from_str_radix(buf.as_str(), radix).map_err(|_| "invalid"),
        (None, Some(radix)) => i64::from_str_radix(buf.as_str(), radix).map_err(|_| "invalid"),
        (None, None) => i64::from_str_radix(buf.as_str(), 10).map_err(|_| "invalid"),
    }
}

fn main() {
    println!("{:?}", parse("0b111", None));
    println!("{:?}", parse("0b111", Some(5)));
}
```

tkbky:
Follow up of [this PR](https://github.com/artichoke/artichoke/pull/254)

Fixes #282 

It passed some tests, failed quite a number too.

Yet to work with the following cases:
1. When passed a `Fixnum`, i.e. `Integer(10)`.
1. Deal with `Float`, `Bignum`, `Rational`.
1. Deal with cases where `Object` responds to `to_int`, `to_i`.
1. Extract `exception` / keyword argument correctly.


lopopolo:
This ticket comes up in the intermediate commits: https://github.com/mruby/mruby/issues/4596

> `mrb_get_args()` does not support keyword arguments yet. So my recommendation is to implement the front-end method (that takes keyword arguments) in Ruby and forward to the C implemented method.

_Originally posted by @matz in https://github.com/mruby/mruby/issues/4596#issuecomment-515648694_

artichoke-backend/src/extn/core/kernel/integer.rs
```diff
@@ -0,0 +1,177 @@
...
+    let mut string = String::new();
+
+    // If have mutliple consecutive embedded underscores, argument error!
+    let multi_underscore_re = Regex::new(r"__+").unwrap();
+    if arg.starts_with('_') || arg.ends_with('_') || multi_underscore_re.is_match(arg) {
+        return Err(Error::InvalidValue(arg.into(), raise_exception));
+    }
+
+    // If have multiple consecutive leading/trailing signs, argument error!
+    let multi_sign_re = Regex::new(r"\+\++|\-\-+").unwrap();
```
lopopolo: i think this bit also does not require a `Regex`

artichoke-backend/src/extn/core/kernel/integer.rs
```diff
@@ -0,0 +1,177 @@
...
+    if multi_sign_re.is_match(arg) {
+        return Err(Error::InvalidValue(arg.into(), raise_exception));
+    }
+
+    // Remove embedded underscore `_`
+    // because they represent error in `from_str_radix` & `from_str`
+    let arg = arg.replace('_', "");
```
lopopolo: Passing an `&str` in the `Args` struct would make it clearer that we can replace this replacement and the above `Regex` passes with a single pass over the `&str` for validation and pushing the valid chars into a new mutable `String` buffer.

artichoke-backend/src/extn/core/kernel/integer.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    // Remove leading and trailing white space,
+    // because they represent error in `from_str_radix` & `from_str`.
+    let arg = arg.trim();
```
lopopolo: this one too :)

artichoke-backend/src/extn/core/kernel/integer.rs
```diff
@@ -0,0 +1,177 @@
...
+    } else {
+        string.push_str(arg);
+    }
+
+    match (radix, parsed_radix) {
+        (Some(radix), Some(parsed_radix)) => {
+            if radix != parsed_radix {
+                return Err(Error::InvalidValue(string, raise_exception));
+            }
+            if let Ok(v) = i64::from_str_radix(string.as_str(), radix as u32) {
```
lopopolo: `radix as u32` is an unsafe cast. Not all values that can be represented as an `i64` (including negative ones!) can be represented in a u32.

Instead we should use [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) to return an error instead of doing something unexpected.

Also, it looks like Ruby takes the absolute value of the passed radix:

```
[2.6.3] > Integer('23', -10, exception: nil)
=> 23
[2.6.3] > Integer('23', -100, exception: nil)
        3: from (irb):10
        2: from (irb):10:in `rescue in irb_binding'
        1: from (irb):10:in `Integer'
ArgumentError (invalid radix 100)
```

`-1` radix is treated as `10`?

```
[2.6.3] > Integer('23', -1, exception: nil)
=> 23
```

artichoke-backend/src/extn/core/kernel/integer.rs
```diff
@@ -0,0 +1,177 @@
...
+                Ok(interp.convert(v))
+            } else {
+                Err(Error::InvalidValue(string, raise_exception))
+            }
+        }
+        (Some(radix), None) | (None, Some(radix)) => {
+            if radix < 2 || radix > 36 {
+                return Err(Error::InvalidRadix(radix.to_string(), raise_exception));
+            }
+            if let Ok(v) = i64::from_str_radix(string.as_str(), radix as u32) {
```
lopopolo: not all platforms uses `i64` as the `Integer` type.

There is a type alias defined at `crate::types::Int`.

artichoke-backend/src/extn/core/kernel/mod.rs
```diff
@@ -108,6 +120,57 @@ impl Kernel {
...
+                }
+            }
+            Err(integer::Error::BaseSpecifiedForNonString(raise_exception)) => {
+                if raise_exception {
+                    ArgumentError::raise(interp, "base specified for non string value")
+                } else {
+                    interp.convert(None::<Value>).inner()
+                }
+            }
+            _ => RuntimeError::raise(interp, "fatal Kernel#Integer error"),
```
lopopolo: is this wildcard match arm necessary? What error enum variants are we not handling?

artichoke-backend/src/extn/core/kernel/mod.rs
```diff
@@ -108,6 +120,57 @@ impl Kernel {
         }
     }
 
+    unsafe extern "C" fn integer(mrb: *mut sys::mrb_state, _slf: sys::mrb_value) -> sys::mrb_value {
+        let interp = unwrap_interpreter!(mrb);
+
+        match integer::Args::extract(&interp) {
```
lopopolo: the pattern I've been using in GH-288 that I like is:

```rust
let (arg, base) = mrb_get_args!(mrb, required = 1, optional = 1);
let interp = unwrap_interpreter!(mrb);
let result = integer::Args::extract(arg, base).and_then(|args| integer::method(&interp, args));
match result {
    Ok(value) => value.inner(),
    Err(variant1) => FooError::raisef(interp, . ..),
    Err(...) => ...,
}
```

This gets rid of some of the nesting and makes the error handling cleaner since both `Args::extract` and `integer::method` share an error enum.

artichoke-backend/src/extn/core/kernel/mod.rs
```diff
@@ -108,6 +120,57 @@ impl Kernel {
...
+            },
+            Err(integer::Error::NoImplicitConversionToString(class_name, raise_exception)) => {
+                if raise_exception {
+                    TypeError::raisef(interp, "can't convert %S into String", vec![class_name])
+                } else {
+                    interp.convert(None::<Value>).inner()
+                }
+            }
+            Err(integer::Error::BaseSpecifiedForNonString(raise_exception)) => {
+                if raise_exception {
```
lopopolo: if we were to do the error swallowing in Rust, we could do it more concisely with

```rust
match result {
    Ok(value) => value.inner(),
    Err(integer::Error::NoImplicitConversionToString(.., raise_exception)) | Err(integer::Error::BaseSpecifiedForNonString(raise_exception)) | ... if !raise_exception => interp.convert(None::<Value>).inner(),
    ...
}
```

lopopolo:
@tkbky I think questions 1-4 become a lot easier if we can do some of the error handling and type conversion in Ruby. What do you think?

artichoke-backend/src/extn/core/kernel/integer.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        if let Ok(arg) = interp.try_convert(Value::new(interp, arg)) {
+            Ok(Self {
+                arg,
+                radix,
+                raise_exception,
+            })
+        } else {
+            Err(Error::NoImplicitConversionToString(
+                value.ruby_type().class_name().into(),
```
lopopolo: this is not the API we want.

We actually want something like:

https://github.com/artichoke/artichoke/blob/549b8b1df775427d8f8924c9d0ae072a4ca34378/artichoke-backend/src/value.rs#L105-L123

artichoke-backend/src/extn/core/kernel/integer.rs
```diff
@@ -0,0 +1,177 @@
...
+use std::collections::HashMap;
+use std::str::FromStr;
+
+use crate::convert::{Convert, TryConvert};
+use crate::types::Ruby;
+use crate::value::Value;
+use crate::Artichoke;
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub enum Error {
```
lopopolo: we should be able to make this struct `Copy` by having the `String` fields be `&'a str` instead of an owned `String`. The `&'a str` references have the same lifetime as the `Artichoke` interpreter.

artichoke-backend/src/extn/core/kernel/integer.rs
```diff
@@ -0,0 +1,177 @@
...
+        // we might want to get the argument manually.
+        let (arg, base, exception) = mrb_get_args!(mrb, required = 1, optional = 2);
+
+        let raise_exception: Option<HashMap<String, bool>> = match exception {
+            Some(exception) => match interp.try_convert(Value::new(interp, exception)) {
+                Ok(exception) => exception,
+                Err(_) => None,
+            },
+            _ => None,
+        };
```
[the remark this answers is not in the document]
tkbky: 😍

tkbky:
> @tkbky I think questions 1-4 become a lot easier if we can do some of the error handling and type conversion in Ruby. What do you think?

Now that I think about it, yes, that'll be way lot easier. Will give that a try.

artichoke-backend/src/extn/core/kernel/integer.rs
tkbky: Tried to use `&'a str` but having problem passing the `digits` owned by the function as ref to `Error`, ended up with `String`. I believe I haven't fully grasp the use of lifetimes yet. 😅

tkbky:
Current state: `Passed 471, skipped 1, not implemented 0, failed 4 specs.`. The 4 failed specs are `Bignum` & `Rational` related.

artichoke-backend/src/extn/core/kernel/integer.rs
```diff
@@ -0,0 +1,170 @@
...
+    match (radix, parsed_radix) {
+        (Some(radix), Some(parsed_radix)) => {
+            if radix != parsed_radix {
+                return Err(Error::InvalidValue(digits));
+            }
+            if let Ok(v) = Int::from_str_radix(&digits, radix.try_into().unwrap()) {
+                let result = interp.convert(v);
+                Ok(result)
+            } else {
+                Err(Error::InvalidValue(digits))
```
lopopolo: Ahh I see. Is it necessary to pass digits here? Why can’t we pass the same arg str we use elsewhere?

lopopolo:
Thanks so much @tkbky! 💎 🚀 

I’ve left some minor Rust feedback, but will merge this PR now.

Thanks for this substantial contribution. How was getting your hands dirty with Rust implementing more than just glue code? 🦀

artichoke-backend/src/extn/core/kernel/integer.rs
tkbky: Ah, yes, I should have pass `arg` to it.

tkbky:
@lopopolo 

> Re: How was getting your hands dirty with Rust implementing more than just glue code?

It wasn't easy but fun! Thanks for the opportunity to learn while contributing to this project.

Issue #249, opened 2019-09-03
Add Value::unchecked_funcall that lets Exceptions pass through


lopopolo:
Wanted in `Warning#warn` when it calls `$stderr::print`.

Add this to the `Value` trait in `artichoke-core` and implement it on the `Value` in mruby backend.

lopopolo:
See GH-281

PR #306, opened 2019-10-14
adds unchecked_funcall

stuarth:
Fixes https://github.com/artichoke/artichoke/issues/249 . Is this the intended behavior for `unchecked_funcall`?

### DIFF
--- a/artichoke-backend/src/value.rs
+++ b/artichoke-backend/src/value.rs
@@ -254,7 +254,7 @@ impl ValueLike for Value {
         func: &str,
         args: &[Self::Arg],
         block: Option<Self::Block>,
-    ) -> Result<Value, ArtichokeError> {
+    ) -> Result<Self, ArtichokeError> {
         self.interp.0.borrow_mut();
         // Ensure the borrow is out of scope by the time we eval code since
         // Rust-backed files and types may need to mutably borrow the `Artichoke` to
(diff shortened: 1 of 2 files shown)


stuarth:
@lopopolo gotcha, thanks for the clarification! This more in line with what you're hoping for?

artichoke-backend/src/value.rs
```diff
@@ -249,6 +249,68 @@ impl ValueLike for Value {
...
+        self.interp.0.borrow_mut();
+        // Ensure the borrow is out of scope by the time we eval code since
+        // Rust-backed files and types may need to mutably borrow the `Artichoke` to
+        // get access to the underlying `ArtichokeState`.
+        let (mrb, _ctx) = {
+            let borrow = self.interp.0.borrow();
+            (borrow.mrb, borrow.ctx)
+        };
+
+        let _arena = self.interp.create_arena_savepoint();
```
lopopolo: Since we’re explicitly dropping the arena index, let’s remove the underscore from the name

artichoke-backend/src/value.rs
```diff
@@ -249,6 +249,68 @@ impl ValueLike for Value {
...
+                drop(_arena);
+                drop(args);
+                drop(func);
+                (*mrb).exc = sys::mrb_sys_obj_ptr(value);
+                sys::mrb_sys_raise_current_exception(mrb);
+                unreachable!("mrb_raise will unwind the stack with longjmp");
+            }
+            value
+        };
+        Ok(Value::new(&self.interp, value))
```
lopopolo: Looks like clippy wants you to refer to `Value` with `Self`

PR #316, opened 2019-10-22
Replace bespoke byte substring searcher with bstr

lopopolo:
> >  I should be able to implement substring matching

> Out of curiosity, why not just use bstr's substring search, since it sounds like you are already depending on bstr?

_Originally posted by @BurntSushi in https://github.com/BurntSushi/rust-memchr/issues/61#issuecomment-544349809_

`bstr`'s twoway implementation has more optimizations than I could ever dream of implementing. I was turned of from trying it initially because of a slower experience using the `twoway` crate. bstr's twoway uses `memchr` and is faster than my handrolled memchr only implementation.

This PR also adds an option `--with-fixture` to the `artichoke` binary which loads a file into a global variable `$fixture`.

### DIFF
--- a/artichoke-backend/src/extn/core/string/scan.rs
+++ b/artichoke-backend/src/extn/core/string/scan.rs
@@ -1,3 +1,4 @@
+use bstr::ByteSlice;
 use std::str;
 
 use crate::convert::{Convert, RustBackedValue};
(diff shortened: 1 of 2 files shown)


artichoke-backend/src/extn/core/string/scan.rs
```diff
@@ -11,94 +12,6 @@ use crate::types::Ruby;
...
-                    start = idx + 1;
-                    if idx + patlen <= strlen {
-                        if &string[idx + 1..idx + patlen] == rest {
-                            matches += 1;
-                            start = idx + patlen;
-                        }
-                    } else {
-                        break;
-                    }
-                }
```
BurntSushi: Just as a note for future reference: the other issue with this implementation of substring search is that it will degrade pretty rapidly in certain pathological cases. e.g., with a needle of `aaaaaaaab` on a haystack of `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa...`. You'll basically wind up comparing the needle at every position of the input. This sort of behavior is endemic to approaches which use memchr acceleration, and serves as a cautionary tale for benchmarking. It's very easy to do well on certain cases while performing _much worse_ in other cases. This is why benchmarking substring search is so difficult. Your benchmarks need a good sampling of both needles and haystacks.

bstr's implementation has protections against the sub-optimal behavior, mostly by keeping track of how effective the memchr acceleration is. If it turns out that memchr is skipping only a small number of bytes on average, then the search will dynamically disable the use of memchr and fallback to the standard Two-Way algorithm.

Issue #256, opened 2019-09-04
Range-backed Arrays


lopopolo:
In thinking how to optimize these benchmarks from [RubyBench](https://rubybench.org/ruby/ruby/releases?result_type=(1..1_000_000).last(10000)):

```yaml
benchmark:
  - (1..1_000_000).last(100)
  - (1..1_000_000).last(1000)
  - (1..1_000_000).last(10000)
```

We can implement a backend for `Array` that is a `Range`. Continuous sequences of `Integer`s or `String`s (even if they are only created from `Range` APIs like `Range#to_a` or `Range#last`) can store the `Range` representation and derive `Array` methods either directly like `Array#[]` or via conversion through a Rust `Iterator`.

This creates no additional garbage (it is equivalent to a local binding to a `Range`, or an `Rc::clone`).

This turns `Range#last` into a constant time operation.

lopopolo:
cc @johnwahba

lopopolo:
Blocked on GH-280.

lopopolo:
GH-320 is merged so this is unblocked.

We would add a Range-backed `Array` as an `ArrayType` in the `integer_range` module.

Expose a constructor for it in `Artichoke::Array` and wire it up in `Range#to_a`.

lopopolo:
GH-458 removed the polymorphic `Array` backends. This ticket is obsolete.

PR #320, opened 2019-10-29
Replace Array implementation with a class cluster

lopopolo:
This commit extends and replaces the existing `VecDeque` backed `Array`
implementation with a cluster of 7 concrete `ArrayType`s.

The basic type is `Buffer`, which uses a `Vec` for storage.

There are several specialized types for `Array`s with fixed contents:

- Empty
- One element
- Two element
- Hole, a run of `nil`s
- Repeated runs of a single element

The `Array` struct that gets serialized into an `mrb_value` is a new
type wrapper struct around a `Box<dyn ArrayType>`. `Array` implements a
class cluster similar to `NSArray` in Foundation SDK via dynamic
dispatch on trait objects.

Not all `ArrayType`s are growable, so mutating methods accept a
reallocation out parameter. The public facade handles reallocations by
delegating to an `Aggregate` struct that is a `Vec` of `ArrayType`s.

Some properties that the Artichoke `Array` has that MRI does not:

- Constant time concat
- Partial reallocation of aggregates
- Constant time reverse, and insertion in some circumstances
- Allocation-free mutations in some circumstances
- Sparse arrays
- Arrays larger than `usize::MAX` elements

```
artichoke 0.1.0 (2019-10-29 revision 1895) [x86_64-darwin]
[Rust 1.38.0 (rev 625451e) on x86_64-apple-darwin]
>>> a = []
=> []
>>> a[100_000_000_000] = 'end'
=> "end"
>>> a[0, 0] = Array.new(200_000_000_000, 'repeated'); nil
=> nil
>>> a.length
=> 300000000001
>>> b = a.reverse; nil
=> nil
>>> b.first(10)
=> ["end", nil, nil, nil, nil, nil, nil, nil, nil, nil]
```


lopopolo:
This PR enables implementing GH-256.

lopopolo:
Some discussion from Twitter: https://twitter.com/eregontp/status/1189115812864544769

dsisnero:
Are you planning on adding single type strategies like the paper - ie
IntStorageStrategy - StringStorageStrategy - ObjectStorageStrategy

lopopolo:
I would like to. It should be easy to specialize the buffer type to store unboxed ints and strings.

Non “immediate” values require also storing the original `RBasic` pointer to enable GC marking and maintain persistent object Id across element accesses.

@dsisnero is this something you’d like to take on?

lopopolo:
I filed GH-326 which I think captures these ideas as a followup.

PR #322, opened 2019-10-30
Fix segfault and borrow errors when running the entire Array spec suite

lopopolo:
The entire suite of `Array` specs is runnable with:

```shell
./scripts/run-spec.sh core array
```

The previous `VecDeque` implementation of `Array` was able to run the specs to completion and had the following pass rate:

```
Passed 654, skipped 116, not implemented 99, failed 345 specs.
```

The class cluster implementation was not able to run the specs to completion:

- Improperly calling `block.funcall("call")` instead of using `mrb_yield` caused segfaults.
- Some methods on `Array` required a mutable borrow _and_ call back into the VM to parse arguments. These funcalls can trigger GC runs. Marking `Array`s requires a shared borrow, which panics.

The fixes are:

- Introduce a `Block` type that properly (and only) implements `yield`.
- Modify `mrb_get_args!` macro to extract block arguments into a `Block` type.
- Disable GC before acquiring mutable borrows and reenable it upon exit. This is a hack.

Real, non-hack solutions to disabling GC include hoisting arg extraction to before taking the mutable borrow or wrapping the boxed trait object in `Array` with a `RefCell`.

With this PR, the array spec suite is runnable to completion again, and the new implementation passes more specs than the old one!

```
Passed 686, skipped 116, not implemented 98, failed 314 specs.
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,7 @@ jobs:
   rust:
     docker:
       - image: circleci/rust:1.38.0-buster
+    resource_class: large
     steps:
       - checkout
       - restore_cache:
--- a/artichoke-backend/src/extn/core/array/mod.rs
+++ b/artichoke-backend/src/extn/core/array/mod.rs
@@ -7,7 +7,7 @@ use crate::extn::core::exception::{
     ArgumentError, Fatal, RangeError, RubyException, RuntimeError, TypeError,
 };
 use crate::types::Int;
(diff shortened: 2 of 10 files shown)


lopopolo:
With this patch, builds on CircleCI kept getting SIGKILLed for exceeding resource limits.

The latest commit bumps the container resources for the main Rust build to 4 CPUs and 8GB of RAM.

PR #325, opened 2019-11-03
Regex crate regexp backend

lopopolo:
This is a WIP PR that adds the Rust Regex crate as an alternate backend to Artichoke's `Regexp`.

This PR is a code dump of a proof of concept I wrote earlier in the summer. It is disabled and hidden behind an undeclared crate feature. The patterns in Artichoke have changed significantly since this code was written, notably argument extraction, error propagation and exception handling, and the introduction of boxed trait objects.

The following work needs to be done to get this merged:

- Rip out enum based dispatch between onig and regex types.
- Introduce a `RegexpType` trait that provides the core functionality to support the Ruby API. This API should prefer to return Rust native types instead of boxed `Value`s.
- Alter the `Regexp` struct to be a newtype wrapper around `Box<dyn RegexpType>`. A newtype wrapper is a tuple-like struct with a single field.
- Move the mruby glue and module init function to an `mruby` module in the `regexp` module. The `mruby` module should contain the VM initializer (safe) and unsafe entrypoints that trampoline to safe rust functions.
- The safe rust entrypoints live in a `trampoline` module and return `Result<Value, Box<dyn RubyException>>`. These trampolines extract the `Regexp` from the receiver `Value` and delegate to the `Regexp` implementation methods.

The `Array` and `ENV` implementations are the templates to refer to when implementing the design.

### DIFF
--- a/artichoke-backend/src/extn/core/regexp/backend/mod.rs
+++ b/artichoke-backend/src/extn/core/regexp/backend/mod.rs
@@ -1,0 +1,1 @@
+pub mod regex;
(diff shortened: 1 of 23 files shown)


lopopolo:
Closed by GH-342.

PR #327, opened 2019-11-03
Enable compile-time selection of ENV backend

lopopolo:
This commit introduces the `artichoke-system-environ` feature. This feature
is enabled by default. When `artichoke-system-environ` is enabled, `ENV` will
delegate to the system environment via `std::env`.

When this feature is disabled, `ENV `will fallback to a `HashMap<Vec<u8>, Vec<u8>>`.
Using an in-memory backend is desirable in Wasm environments where (modulo an
emscripten shim) there is no environ. This also helps Artichoke tell a better
sandboxing story.

This commit also makes use of a technique introduced in the
`Kernel#Integer` patch which exposes functionality through an `Artichoke`
module that the Core delegates to. `Artichoke::Environ#[]`,
`Artichoke::Environ#[]=`, and `Artichoke::Environ#to_h` provide the same
glue that originally came from `EnvClass`.

`Artichoke::Environ` provides the same interface but as self methods
that memoize an `Environ` instance. This allows the Artichoke runtime to
inject a different concrete `ENV` backend per interpreter.

Shout out to @cdeler and @tkbky for laying the foundational work to make
this PR possible.

PR #328, opened 2019-11-03
Avoid double allocation of MRB_TT_DATA objects

lopopolo:
There is a better API than mrb_obj_new for allocating MRB_TT_DATA objects
called mrb_data_object_alloc which allows to directly inject a data pointer
and avoid going through the object's initialize method.

Prior to this change, every Array serialization to Ruby from Rust resulted
in _two_ arrays being allocated _and_ the first being leaked.

This commit removes the gross workaround for passing arguments to the
Regexp constructor in the Regexp impl of RustBackedObject.


lopopolo:
Found this when working on fixing segfaults in GH-327.

PR #332, opened 2019-11-04
Add missing indirection via Array access macros in mruby impl

lopopolo:
Just a little bit of UB and type confusion.


lopopolo:
Extracted from GH-327.

PR #348, opened 2019-11-12
Disable unneeded converters for collection types

lopopolo:
This improves clean build times on my 8 core system from 1m30s to 58s.


lopopolo:
This patch improves build times in CI by ~1 minute

PR #357, opened 2019-11-13
Implement Array#shuffle and Array#shuffle! in Rust

nerdinand:
I'm a Rubyist and new to Rust, so please be nice! ;-)

This is work in progress, missing:

- [x] hole.rs implementation of `shuffle_bang`? (I'm not exactly sure what a "hole" array is)
- [ ] options hash (passing a `Random` instance): `a.shuffle!(random: Random.new(1))  #=> [1, 3, 2]`

### DIFF
--- a/artichoke-backend/Cargo.toml
+++ b/artichoke-backend/Cargo.toml
@@ -12,8 +12,8 @@ log = "0.4"
 memchr = "2"
 once_cell = "1"
 onig = "5"
-regex = "1"
 rand = "0.7"
+regex = "1"
 
 [dependencies.artichoke-core]
 path = "../artichoke-core"
--- a/artichoke-backend/Cargo.toml
+++ b/artichoke-backend/Cargo.toml
@@ -13,6 +13,7 @@ memchr = "2"
 once_cell = "1"
 onig = "5"
 regex = "1"
+rand = "0.7"
 
 [dependencies.artichoke-core]
 path = "../artichoke-core"
@@ -44,6 +45,3 @@ default-features = false
 default = ["artichoke-array", "artichoke-system-environ"]
 artichoke-array = []
 artichoke-system-environ = []
-
-[dependencies.rand]
-version = "0.7.2"
--- a/artichoke-backend/Cargo.toml
+++ b/artichoke-backend/Cargo.toml
@@ -44,3 +44,6 @@ default-features = false
 default = ["artichoke-array", "artichoke-system-environ"]
 artichoke-array = []
 artichoke-system-environ = []
+
+[dependencies.rand]
+version = "0.7.2"
--- a/artichoke-backend/src/extn/core/array/backend/aggregate.rs
+++ b/artichoke-backend/src/extn/core/array/backend/aggregate.rs
@@ -342,8 +342,15 @@ impl ArrayType for Aggregate {
 
     fn shuffle(&mut self, interp: &Artichoke) -> Result<(), Box<dyn RubyException>> {
         let _ = interp;
+
+        let mut buffer = Vec::with_capacity(self.0.len());
+        for idx in 0..self.0.len() {
+            buffer.push(self.0.get(idx).unwrap());
+        }
+    
         let mut rng = rand::thread_rng();
-        self.0.shuffle(&mut rng);
+        buffer.shuffle(&mut rng);
+        
         Ok(())
     }
 }
--- a/artichoke-backend/src/extn/core/array/backend/aggregate.rs
+++ b/artichoke-backend/src/extn/core/array/backend/aggregate.rs
@@ -340,10 +340,13 @@ impl ArrayType for Aggregate {
         Ok(Box::new(Self::with_parts(parts)))
     }
 
-    fn shuffle_bang(&mut self, interp: &Artichoke) -> Result<Box<dyn ArrayType>, Box<dyn RubyException>> {
+    fn shuffle_bang(
+        &mut self,
+        interp: &Artichoke,
+    ) -> Result<Box<dyn ArrayType>, Box<dyn RubyException>> {
         let _ = interp;
         let mut rng = rand::thread_rng();
-        self.0.shuffle(&mut rng); 
+        self.0.shuffle(&mut rng);
         Ok(self.box_clone())
     }
 }
--- a/artichoke-backend/src/extn/core/array/backend/aggregate.rs
+++ b/artichoke-backend/src/extn/core/array/backend/aggregate.rs
@@ -340,13 +340,13 @@ impl ArrayType for Aggregate {
         Ok(Box::new(Self::with_parts(parts)))
     }
 
-    fn shuffle_bang(
+    fn shuffle(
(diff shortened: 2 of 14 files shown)


nerdinand:
Looks like the `Random` class is implemented in C in MRI: https://github.com/ruby/ruby/blob/master/random.c
So I guess we'd have to implement it in Rust? This will probably need some proper planning...

lopopolo:
The build failure was my bad (merged a bunch of PRs in parallel last night). Should be fixed if you rebase to pick up GH-358.

artichoke-backend/src/extn/core/array/backend/fixed/two.rs
```diff
@@ -217,4 +217,15 @@ impl ArrayType for Two {
         let _ = interp;
         Ok(backend::fixed::two(self.1.clone(), self.0.clone()))
     }
+
+    fn shuffle_bang(
+        &mut self,
+        interp: &Artichoke,
+    ) -> Result<Box<dyn ArrayType>, Box<dyn RubyException>> {
+        let _ = interp;
+        if rand::random() {
```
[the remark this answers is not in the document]
nerdinand: Thanks! ^^

artichoke-backend/src/extn/core/array/backend/aggregate.rs
```diff
@@ -337,4 +339,14 @@ impl ArrayType for Aggregate {
...
         Ok(Box::new(Self::with_parts(parts)))
     }
+
+    fn shuffle_bang(
+        &mut self,
+        interp: &Artichoke,
+    ) -> Result<Box<dyn ArrayType>, Box<dyn RubyException>> {
+        let _ = interp;
+        let mut rng = rand::thread_rng();
+        self.0.shuffle(&mut rng);
```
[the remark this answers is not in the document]
nerdinand: Hmm ok, so how would you convert the aggregate to a buffer? Looping over the parts and collecting them somehow in a Vec? And then constructing a buffer from that?

nerdinand:
@lopopolo here's another question: how do I skip the specs that can't pass yet (e.g. because `Random` isn't implemented)?

lopopolo:
@nerdinand, normally I’d say stub out `Random` to get the tests running without throwing and let the spec runner ignore the `NotInplementedError`s raised by the `Array#shuffle!` method.

However it looks like your change is tickling a segfault in the VM that I’ve been unable to figure out.

I’d suggest disabling the spec suite in `scripts/lint.sh` and `.circleci/config.yml` To unblock you.

artichoke-backend/Cargo.toml
```diff
@@ -13,6 +13,7 @@ memchr = "2"
 once_cell = "1"
 onig = "5"
 regex = "1"
+rand = "0.7"
```
lopopolo: Can you alphabetize `rand` above `regex`?

artichoke-backend/src/extn/core/array/backend/aggregate.rs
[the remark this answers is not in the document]
lopopolo: That is the approach taken here: https://github.com/artichoke/artichoke/blob/d8a2b74076ce81690934531aa53de1ba3894f7a0/artichoke-backend/src/extn/core/array/mod.rs#L34-L40

Another way we might go about this is to lazily shuffle with reservoir sampling. For example, shuffling an `Array` and then calling `ary[4]` doesn't require the whole `Array` to be shuffled. It just requires us to randomly assign one index.

The `as_vec` approach is simpler though and I'd advocate for starting with that before getting fancy.

artichoke-backend/src/extn/core/array/backend/fixed/two.rs
```diff
@@ -217,4 +219,12 @@ impl ArrayType for Two {
         let _ = interp;
         Ok(backend::fixed::two(self.1.clone(), self.0.clone()))
     }
+
+    fn shuffle(&mut self, interp: &Artichoke) -> Result<(), Box<dyn RubyException>> {
+        let _ = interp;
+        if rand::random() {
+            mem::swap(&mut self.0, &mut self.1);
+        }
```
lopopolo: this right here is a nice bit of Rust.

artichoke-backend/src/extn/core/array/backend/aggregate.rs
```diff
@@ -337,4 +339,18 @@ impl ArrayType for Aggregate {
...
+
+        let mut buffer = Vec::with_capacity(self.0.len());
+        for idx in 0..self.0.len() {
+            buffer.push(self.0.get(idx).unwrap());
+        }
+    
+        let mut rng = rand::thread_rng();
+        buffer.shuffle(&mut rng);
+        
+        Ok(())
```
nerdinand: @lopopolo This doesn't work yet. The problem is my limited understanding of Rust... 
I don't understand how to convert the aggregate to a buffer and assign it to `self` (is that even what I should be doing?)...

artichoke-backend/src/extn/core/array/backend/aggregate.rs
```diff
@@ -337,4 +339,18 @@ impl ArrayType for Aggregate {
...
+    fn shuffle(&mut self, interp: &Artichoke) -> Result<(), Box<dyn RubyException>> {
+        let _ = interp;
+
+        let mut buffer = Vec::with_capacity(self.0.len());
+        for idx in 0..self.0.len() {
+            buffer.push(self.0.get(idx).unwrap());
+        }
+
+        let mut rng = rand::thread_rng();
+        buffer.shuffle(&mut rng);
```
lopopolo: There two ways to make the result of the shuffle take effect.

The first is to call backend::buffer::Buffer::from(buffer) and replace the parts of the aggregate with this single part.

The alternative approach is to introduce a `realloc` out parameter to the shuffle signature similar to pop. The realloc parameter allows the aggregate to signal to the `Array` container that it needs to change representation. You would pass the buffer back to the container this way.

lopopolo:
Hey @nerdinand sorry for all the changes here, but in addition to the `ArrayType`s there is also a new `InlineBuffer` type introduced in GH-382 that a rebase will pick up.

lopopolo:
@nerdinand GH-384 is merged so we should be able to implement the `Random` parameter.

Keyword arguments are tricky/impossible to work with in mruby, so we might take a similar approach to `Kernel#Integer` which is to extract the kwargs and forward to a method in the `Artichoke` namespace.

This lets us do most of the error handling in Ruby where it is convenient.

lopopolo:
Thanks for the PR. Since it has been a while since the author responded, I’m closing this PR for being stale. Please reopen if you’d like to pick this back up.

nerdinand:
Sorry @lopopolo i got overwhelmed with this. And I don't have the time to learn me some proper Rust at the moment, unfortunately... Maybe I'll come back some day with more time...

lopopolo:
No worries at all. We both learned a lot about how to approach the issue from this PR 🚀 💎

PR #338, opened 2019-11-05
Add a build on Windows with the Rust msvc toolchain

lopopolo:
The build installs Rust, MRI, and executes `cargo build`.

Tests are run via `cargo test` but they currently segfault. Test failures are suppressed but tests are still built to ensure they compile.

PR #360, opened 2019-11-15
Modify build.rs to build artichoke-backend under Windows

lopopolo:
- Thread target triple through to all build parts.
- Invoke minirake via ruby binary instead of via shebang.
- Disable auto importer on expect, pty, and syslog packages
  which are not available on windows.


lopopolo:
This PR was extracted from GH-338.

Issue #303, opened 2019-10-14
Null out mruby “host” build


lopopolo:
https://github.com/artichoke/artichoke/issues/32#issuecomment-520263996 Mentioned that once we removed the dependency on the mruby “ext” family of gems we would be able to make the mruby “host” build a noop.

GH-288 sucked the mrblib Ruby sources into Artichoke `extn` which means we no longer require `mrbc` on the host build.

Duplicate the `true` configuration for all commands from the `sys` build to the host build.

Turn on the parallel feature for `cc` crate.

PR #361, opened 2019-11-15
Null out the mruby host build

lopopolo:
mruby requires a host build that is not configurable enough to build
Artichoke. This has meant that Artichoke builds the host build to
bootstrap a 'sys' crossbuild that is used to produce the static library.

In the beginning, mruby's build process was used directly to produce
the static library that build.rs would link into Artichoke. More
recently, the sys crossbuild would only do C source generation and
build.rs would discover and compile these sources with the cc
crate.

Now that Artichoke no longer depends on mrblib or Ruby sources in
mrbgems, mrbc is no longer required and the host build is no longer
needed.

This PR replaces the host build with a noop. This changes the code
emitted by the mruby build process to eliminate gem irep embedding.
The bootstrap gembox no longer includes or builds any mrbgems.

Fixes GH-303.

This PR was extracted from GH-338.

### DIFF
--- a/artichoke-backend/bootstrap.gembox
+++ b/artichoke-backend/bootstrap.gembox
@@ -9,6 +9,4 @@
 # This gembox can become empty and the default build nulled out once the
 # Artichoke runtime is complete.
 MRuby::GemBox.new do |conf|
-  conf.gem core: 'mruby-compiler'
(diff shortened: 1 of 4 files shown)


lopopolo:
GH-32

Issue #317, opened 2019-10-23
State::str_intern should take T: Into<Cow<'static, [u8]>>


lopopolo:
Most `Symbol`s are `&'static str` in the source.

Convert the API to take byte slices instead.

e.g. turn

https://github.com/artichoke/artichoke/blob/b9ae8501d9f53c5f583b590d11f2804679875667/artichoke-backend/src/state.rs#L248

into

```rust
    pub fn sym_intern<T>(&mut self, sym: T) -> sys::mrb_sym
    where
        T: Into<Cow<'static, [u8]>>
```

and turn

https://github.com/artichoke/artichoke/blob/b9ae8501d9f53c5f583b590d11f2804679875667/artichoke-backend/src/state.rs#L25

into

```rust
    symbol_cache: HashMap<Cow<'static, [u8]>, sys::mrb_sym>,
```

This will require updating callsites of `sym_intern` to send byte literals or call `as_bytes` on a `String`.

lopopolo:
This change avoids allocations when Symbols are known at compile time as `&'static` literals.

PR #364, opened 2019-11-15
Reduce allocations of Symbols backed by &'static byte strings

lopopolo:
The `Symbol` cache in `State` held owned `String`s as the keys.
This was suboptimal because:

1. Symbols can contain arbitrary byte content.
2. Most symbols are known at compile time and have &'static lifetime.

This PR changes the type of the keys in the `Symbol` cache to
`Cow<'static, [u8]>`. This allows the state to avoid allocating a byte
vector unless it needs to to satisy ownership constraints.

Fixes GH-317.


lopopolo:
The lint error here is easy to fix by assigning the symbols to constants.

Additionally, we can avoid allocations of common capture globals (`$0` through `$9`) by delegating to a factory function for global names that returns a `Cow` with a match on `n`.

PR #365, opened 2019-11-17
Improve CircleCI build

lopopolo:
- Add sccache on Linux and Windows.
- Unify all linting tasks into a single job.
- Rename `rust` build and test on Linux to
  match Windows job.
- Use `debian:buster-slim` as the base image
  on Linux.
- Bump docs deploy base image to LTS node.
- Split ruby/spec suite testing into its own job.
- Suppress errors in ruby/spec job.
- Use `shfmt` to discover sources for `shellcheck`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,11 @@ jobs:
       - image: debian:buster-slim
     resource_class: large
     steps:
+      - run:
+          name: Setup Imagr
+          command: |
+            apt-get update
+            apt-get install -y curl git ssh
       - checkout
       - restore_cache:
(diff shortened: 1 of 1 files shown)


lopopolo:
Putting this branch on hold until I can improve build times by building a base image. This exercise has made it more clear what needs to be in that base image.

lopopolo:
This PR enables marking the ruby/spec suite as not required for merge.

Follow up: create a base image for the Linux builder that pulls the setup out of the circleci definition.
Follow up: ditto for the Linux linter.

Most of the slowness on Linux comes from building Ruby from scratch.

PR #366, opened 2019-11-19
Fix String#delete_suffix / prefix methods

ChrisBr:
There were two issues with the delete methods

* In case the prefix / suffix does not match, we need to return
a copy of the string
* The methods were destructive

Disclaimer: Patch is untested!

### DIFF
--- a/artichoke-backend/src/extn/core/string.rb
+++ b/artichoke-backend/src/extn/core/string.rb
@@ -272,9 +272,11 @@ class String
   end
 
   def delete_prefix(prefix)
-    self[0...prefix.length] = '' if start_with?(prefix)
+    raise TypeError, "no implicit conversion of #{prefix.class} into String" unless prefix.is_a?(String)
 
-    self
+    return self[prefix.length..-1] if start_with?(prefix)
+
+    dup
   end
 
   def delete_prefix!(prefix)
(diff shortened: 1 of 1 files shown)


artichoke-backend/src/extn/core/string.rb
```diff
@@ -261,9 +261,9 @@ def delete!(*args)
   end
 
   def delete_prefix(prefix)
-    self[0...prefix.length] = '' if start_with?(prefix)
+    return self[0...prefix.length] if start_with?(prefix)
```
lopopolo: Should this be

```ruby
return self[prefix.length..-1] if start_with?(prefix)
```

artichoke-backend/src/extn/core/string.rb
```diff
@@ -272,9 +272,9 @@ def delete_prefix!(prefix)
   end
 
   def delete_suffix(suffix)
-    self[-suffix.length..-1] = '' if end_with?(suffix)
+    return self[-suffix.length..-1] if end_with?(suffix)
```
lopopolo: Ditto above. Should this be

```ruby
return self[0..-suffix.length] if end_with?(suffix)
```

ChrisBr:
>  It looks like the slices need to be updated since we’re no longer being destructive

Absolutely right 👍 Updated! Also added a basic type check.

PR #369, opened 2019-11-22
Bump windows orb dependency to 2.2.0

lopopolo:


lopopolo:
Build script panics at bindgen:

```
C:\Program Files\LLVM\lib\clang\9.0.0\include\inttypes.h:21:15: fatal error: 'inttypes.h' file not found
C:\Program Files\LLVM\lib\clang\9.0.0\include\inttypes.h:21:15: fatal error: 'inttypes.h' file not found, err: true
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', src\libcore\result.rs:1165:5
```

lopopolo:
This PR is stale and fixing the build is probably not a productive use of time.

Issue #335, opened 2019-11-05
Change ArrayType::reverse to do an in-place reversal


lopopolo:
Currently, `ArrayType::reverse` returns a reversed copy of the array. `Array::reverse_bang` is implemented in terms of this API.

Many `ArrayType` backends do not require allocations to do an in-place reversals, so this is wasteful.

Instead, make `ArrayType::reverse` do an in-place reversal and implement the Ruby method `Array#reverse` in Ruby like:

```ruby
def reverse
  dup.reverse!
end
```

This ticket fell out of GH-334.

PR #370, opened 2019-11-22
Change ArrayType::reverse to do an in-place reversal

aivensong:
This PR addresses #335. 

- Implement `Array#reverse` in ruby
- Update `ArrayType::reverse` to do an in-place reversal
- Cleanup the now unused `reverse` function in trampoline and update `reverse_bang` to use the in-place reversal
- Cleanup `ary_reverse` method in mruby

### DIFF
--- a/artichoke-backend/src/extn/core/array/backend/buffer.rs
+++ b/artichoke-backend/src/extn/core/array/backend/buffer.rs
@@ -227,11 +227,12 @@ impl ArrayType for Buffer {
 
     fn reverse(&mut self, interp: &Artichoke) -> Result<(), Box<dyn RubyException>> {
         let _ = interp;
-        let ln = self.0.len();
         let mut i = 0;
-        while i < ln / 2 {
-            self.0.swap(i, ln - i - 1);
+        let mut j = self.0.len() - 1;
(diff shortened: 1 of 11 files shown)


artichoke-backend/src/extn/core/array/backend/buffer.rs
```diff
@@ -225,12 +225,15 @@ impl ArrayType for Buffer {
         Ok(interp.convert(self.0.pop()))
     }
 
-    fn reverse(&self, interp: &Artichoke) -> Result<Box<dyn ArrayType>, Box<dyn RubyException>> {
+    fn reverse(&mut self, interp: &Artichoke) -> Result<(), Box<dyn RubyException>> {
         let _ = interp;
-        let mut buffer = Vec::with_capacity(self.0.len());
-        for idx in (0..self.0.len()).rev() {
-            buffer.push(self.0[idx].clone());
+        let mut i = 0;
```
lopopolo: I think this code can be replaced with

```rust
self.0.reverse();
```

aivensong:
Makes sense! Much cleaner that way, pushed an update.

Issue #356, opened 2019-11-13
Remove bash dependency from development toolchain


lopopolo:
Bash is not present in powershell-esque environments.

The scripts in  the `scripts` directory should be ported to either JS/node or Ruby to rely on dependencies already in the project.

For example, format-text.sh can be implemented as a node script with [`node-walk`](https://www.npmjs.com/package/walk) and calling prettier in javascript.

PR #376, opened 2019-11-23
Reimplement scripts/lint.sh as a Node script

lopopolo:
scripts/lint.sh is slow. A clean run takes over 30 seconds.

scripts/lint.sh and all of the bash scripts it depends on also do not
run on Windows.

This commit introduces a Node script that will run anywhere Node does.
All lint steps are run in parallel. In practice, the script executes as
quickly as `cargo clippy`. On a clean build this takes 2 seconds.

The script accepts an optional argument `--check` which puts the script
into enforcement mode where it will error on failed lint and formatting.

The script executes the following lints:

- Format text sources with prettier
- Format and wrap markdown sources with prettier
- ESLint
- rustfmt
- clippy
- rust doc
- shfmt
- shellcheck
- clang-format
- rubocop

Like the scripts/lint.sh script that it replaces, scripts/lint.js is
robust to an uninitialzed developer environment. Unlike the old script,
it does not attempt to auto install Ruby or JavaScript dependencies.

This PR is to support GH-356.


lopopolo:
This PR speeds up the linter job from 2:45 to 2:08.

Issue #363, opened 2019-11-15
Add constants for Regexp globals


lopopolo:
Global variables are addressed by a `Symbol` with the name of the global, e.g. `$&`.

The `Regexp` implementation creates lots of globals all over the place.

Define constants for the global names.

lopopolo:
Closed by GH-381.

PR #381, opened 2019-11-23
Use mrb_intern_static for interned Symbols from Rust

lopopolo:
mruby has a basic notion of lifetimes with two APIs for creating
Symbols:

- `mrb_intern`, which takes ownership of a borrowed Symbol by doing an
  allocation and a memcpy.
- `mrb_intern_static`, which takes a Symbol whose lifetime outlives the
  `mrb_state` and does no copy or free.

`mrb_intern_static` is only ergonomic for static String literals in C,
but when adding a `Cow<'static, [u8]>` to a `HashMap` in Rust, the bytes
will be cleaned up automatically when the interpreter closes.

This avoids an allocation for every `Symbol` created in Rust.

### DIFF
--- a/artichoke-backend/src/state.rs
+++ b/artichoke-backend/src/state.rs
@@ -257,7 +257,7 @@ impl State {
         let interned = self
             .symbol_cache
             .entry(sym)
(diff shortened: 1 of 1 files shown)


lopopolo:
Fixes GH-363.

PR #382, opened 2019-11-23
Reimplement Array with a buffer type that optionally inlines

lopopolo:
`InlineBuffer` is a new Array implementation based on `Vec` and
`ArrayVec`. `Array`s with 8 elements or fewer are inlined into the
struct body. Larger `Arrays` are stored elsewhere on the heap in a
`Vec`.

`InlineBuffer` has a similar but not identical API to `ArrayType`. It partially implements `ArrayType` with this API. The other `ArrayType` backends are disabled with this
commit.

`InlineBuffer` supports moving `Array` contents from inline to dynamic
storage and back according to resizes and mutations.

`InlineBuffer` comes with a massive simplification of the modifications
to the mruby VM required to inject an Artichoke `Array`. This is
possible because:

- `InlineBuffer` stores `sys::mrb_value` rather than `Value`.
- `InlineBuffer` is coercable to `*const sys::mrb_value` and
  `*mut sys::mrb_value`.
- `InlineBuffer` supports a `set_len` API to enable mutating via raw
  pointers in the mruby VM.

This patch speeds up execution of the passing ruby/specs slightly.

I hope this patch will eliminate the segfaults and allow deploying the
playground with all features and maybe upgrading to the LLVM emcc
backend.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,6 +30,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "artichoke-backend"
 version = "0.1.0"
 dependencies = [
+ "arrayvec 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "artichoke-core 0.1.0",
  "artichoke-vfs 0.5.0-alpha",
  "backtrace 0.3.40 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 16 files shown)


lopopolo:
Thanks @DarkDimius for the idea to do this.

PR #397, opened 2019-11-29
Enable hash converters behind feature

### LINKED ISSUE #349: Re-enable all converters behind a feature flag
GH-348 disabled lots of converters for collection types to improve compilation speed.

These converters are part of the public API and should be enable-able behind a crate feature.

vinistock:
Closes #349.

This PR puts all the hash converters that were commented out in #348 behind a feature.

Please do provide feedback about the module and the feature name. Also, let me know if I missed something.


artichoke-backend/Cargo.toml
```diff
@@ -52,3 +52,4 @@ artichoke-array = []
 artichoke-debug = ["backtrace"]
 artichoke-random = ["rand"]
 artichoke-system-environ = []
+artichoke-all-converter = []
```
lopopolo: can you alphabetize this list?

artichoke-backend/src/convert/array.rs
```diff
@@ -184,153 +185,304 @@ array_to_ruby!(&'a [Option<Float>]);
...
-// array_to_ruby!(HashMap<String, &'a str>);
-// array_to_ruby!(HashMap<&'a str, Value>);
-// array_to_ruby!(HashMap<&'a str, bool>);
-// array_to_ruby!(HashMap<&'a str, Vec<u8>>);
-// array_to_ruby!(HashMap<&'a str, Int>);
-// array_to_ruby!(HashMap<&'a str, Float>);
-// array_to_ruby!(HashMap<&'a str, String>);
-// array_to_ruby!(HashMap<&'a str, &'a str>);
+#[cfg(feature = "artichoke-all-converters")]
+mod hash_convertions {
```
lopopolo: can this module be called `optional`?

artichoke-backend/src/convert/array.rs
```diff
@@ -1,4 +1,5 @@
-// use std::collections::HashMap;
+#[cfg(feature = "artichoke-all-converters")]
+use std::collections::HashMap;
```
lopopolo: if this `use` statement is only required by the optional converters which are nested in a module, can you add that use directive into that module?

As it stands, I don't know if this compiles.

artichoke-backend/Cargo.toml
lopopolo: based on the flag used in the code, it looks like this should be

```toml
artichoke-all-converters = []
```

converters with the trailing `s`.

vinistock:
I've addressed your comments. Completely missed the hash and nilable files, apologies!

I used `use super::*;` to get the parent module's imports and definitions. I'm unsure if that's the appropriate way of doing it. Please let me know if you have other concerns!

Issue #31, opened 2019-07-07
Don't mutably borrow in Define::define


lopopolo:
This code looks like it should work but it panics:

```rust
let metrics = interp
            .borrow()
            .module_spec::<Metrics>();
            .and_then(|spec| spec.borrow().value(&interp));
```

See if there is a way to fix this. Maybe by interning the symbol when we have the mutable borrow in `def_module` and passing it to `module::Spec`?

see also artichoke/artichoke#21.

Issue #27, opened 2019-07-07
Add APIs on class::Spec and module::Spec for defining constants


lopopolo:
Currently, regexp::init defines constants with eval and `format!`.

Add an API that allows defining a constant that takes anything that implements `FromMrb<Value>` in the same way that you add a method to these specs.

lopopolo:
Closed by GH-526.

Issue #19, opened 2019-07-28
Implement extend for class and module specs


lopopolo:
Support the following from Rust:

```ruby
module Inc
  def inc; end
end

module Mod
  extend Inc
end

class Cls
  extend Inc
end
```

lopopolo:
Add these APIs to core.

Issue #18, opened 2019-07-28
Implement include module for class and module specs


lopopolo:
Support the following from Rust:

```ruby
module Inc
  def inc; end
end

module Mod
  include Inc
end

class Cls
  include Inc
end
```

PR #398, opened 2019-12-01
Remove Rc and RefCell from class and module spec API

lopopolo:
This PR changes the API for creating classes and modules and defining methods on them.

The API now looks like this:

```rust
    let spec = class::Spec::new("Integer", None, None);
    class::Builder::for_spec(interp, &spec)
        .add_method("chr", Integer::chr, sys::mrb_args_opt(1))
        .add_method("size", Integer::size, sys::mrb_args_none())
        .define()?;
    interp.0.borrow_mut().def_class::<Integer>(spec);
```

- `Rc` and `RefCell` are no longer part of the API.
- `Spec` types are read only and immutable.
- Defining items on specs is split into a separate builder type.
- Minimized interior mutability on `module::Spec` to a single symbol.
- Path forward to defining constants on classes and modules, GH-27.
- Path forward for module APIs, GH-18, GH-19.

Fixes GH-21.
Fixes GH-31.


lopopolo:
This PR fixes the definition of `module::Builder::add_self_function` to add a self function instead of the module function it was defining before. There is a new API to define module functions.

PR #406, opened 2019-12-04
Changes .inspect/.to_s to return Vec<u8>

### LINKED ISSUE #390: Captured IO should support arbitrary byte content
The state in artichoke backend supports capturing IO into a String buffer.

You can `puts` binary data which lossily converts data to `<unknown>`.

Not sure if this happens because of `Value::to_s` or `State::print`.

Found this when trying to print the results of `Random::urandom`.

tanaythan:
Fixes #390. The root of the bug stems from `Value::to_s` cannot convert some byte sequences to UTF-8, so the fix suggested is to use the lossy UTF-8 conversion. I tried doing this in the implementation of `TryConvert<Value, &'a str> for Artichoke`, but ran into temporary value issues.

### DIFF
--- a/artichoke-backend/src/extn/core/integer/mod.rs
+++ b/artichoke-backend/src/extn/core/integer/mod.rs
@@ -50,7 +50,7 @@ impl Integer {
                 &interp,
                 format!(
                     "encoding parameter of Integer#chr (given {}) not supported",
-                    encoding.inspect()
+                    String::from_utf8_lossy(&encoding.inspect()).to_string()
                 ),
             )))
         } else {
--- a/artichoke-backend/src/extn/core/integer/mod.rs
+++ b/artichoke-backend/src/extn/core/integer/mod.rs
@@ -49,10 +49,7 @@ impl Integer {
             let mut message = b"encoding parameter of Integer#chr (given".to_vec();
             message.extend(encoding.inspect());
             message.extend(b") not supported");
-            Err(Box::new(NotImplementedError::new_raw(
-                &interp,
-                message,
-            )))
+            Err(Box::new(NotImplementedError::new_raw(&interp, message)))
         } else {
             // When no encoding is supplied, MRI assumes the encoding is
             // either ASCII or ASCII-8BIT.
--- a/artichoke-backend/src/extn/core/integer/mod.rs
+++ b/artichoke-backend/src/extn/core/integer/mod.rs
@@ -46,7 +46,7 @@ impl Integer {
         let interp = unwrap_interpreter!(mrb);
         let encoding = encoding.map(|encoding| Value::new(&interp, encoding));
         let result: Result<Value, Box<dyn RubyException>> = if let Some(encoding) = encoding {
-            let mut message = b"encoding parameter of Integer#chr (given".to_vec();
+            let mut message = b"encoding parameter of Integer#chr (given ".to_vec();
             message.extend(encoding.inspect());
             message.extend(b") not supported");
             Err(Box::new(NotImplementedError::new_raw(&interp, message)))
--- a/artichoke-backend/src/extn/core/integer/mod.rs
+++ b/artichoke-backend/src/extn/core/integer/mod.rs
@@ -46,12 +46,12 @@ impl Integer {
         let interp = unwrap_interpreter!(mrb);
         let encoding = encoding.map(|encoding| Value::new(&interp, encoding));
         let result: Result<Value, Box<dyn RubyException>> = if let Some(encoding) = encoding {
-            Err(Box::new(NotImplementedError::new(
+            let mut message = b"encoding parameter of Integer#chr (given".to_vec();
+            message.extend(encoding.inspect());
+            message.extend(b") not supported");
+            Err(Box::new(NotImplementedError::new_raw(
                 &interp,
-                format!(
-                    "encoding parameter of Integer#chr (given {}) not supported",
-                    String::from_utf8_lossy(&encoding.inspect()).to_string()
-                ),
+                message,
             )))
         } else {
             // When no encoding is supplied, MRI assumes the encoding is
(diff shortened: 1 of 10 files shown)


lopopolo:
A good example of this idea of only converting at the edges is `RubyException::message` and its `fmt::Debug` impl

artichoke-backend/src/extn/core/integer/mod.rs
```diff
@@ -50,7 +50,7 @@ impl Integer {
                 &interp,
                 format!(
                     "encoding parameter of Integer#chr (given {}) not supported",
-                    encoding.inspect()
+                    String::from_utf8_lossy(&encoding.inspect()).to_string()
```
lopopolo: There is an API to construct an exception from a byte vector called `NotImplementedError::new_raw`. You'd use it like:

https://github.com/artichoke/artichoke/blob/2421bfcc24a9035575319a5205f10252437cb0a1/artichoke-backend/src/extn/core/env/backend/memory.rs#L74-L78

artichoke-backend/src/extn/core/kernel/mod.rs
```diff
@@ -90,7 +90,7 @@ impl Kernel {
         let interp = unwrap_interpreter!(mrb);
 
         for value in args.iter() {
-            let s = Value::new(&interp, *value).to_s();
+            let s = String::from_utf8_lossy(&Value::new(&interp, *value).to_s()).to_string();
```
lopopolo: let's add a followup to also have `State::print` and `State::puts` take `Vec<u8>` as well. Thanks for putting in this work to make that next step more obvious :)

artichoke-backend/src/extn/core/kernel/mod.rs
```diff
@@ -103,7 +103,7 @@ impl Kernel {
                     do_puts(interp, &value);
                 }
             } else {
-                let s = value.to_s();
+                let s = String::from_utf8_lossy(&value.to_s()).to_string();
```
lopopolo: ditto let's make `State::puts` take a byte vector in a followup PR.

artichoke-backend/src/extn/core/regexp/mod.rs
```diff
@@ -134,7 +134,9 @@ impl Regexp {
             let encoding = match enc::parse(&encoding) {
                 Ok(encoding) => Some(encoding),
                 Err(enc::Error::InvalidEncoding) => {
-                    let warning = format!("encoding option is ignored -- {}", encoding.to_s());
+                    let encoding_bytes = encoding.to_s();
+                    let encoding_string = String::from_utf8_lossy(&encoding_bytes);
```
lopopolo: TODO: modify `Warn::warn` to take a byte slice.

artichoke-backend/src/value.rs
```diff
@@ -158,7 +158,8 @@ impl Value {
     ///
     /// This function can never fail.
     pub fn to_s_debug(&self) -> String {
-        format!("{}<{}>", self.ruby_type().class_name(), self.inspect())
+        let inspected_str = String::from_utf8_lossy(&self.inspect()).to_string();
+        format!("{}<{}>", self.ruby_type().class_name(), inspected_str)
```
lopopolo: This code can be modified to avoid an allocation by assigning `self.inspect` to a temporary:

```rust
let inspect = self.inspect();
format!("{}<{}>", self.ruby_type().class_name(), String::from_utf8_lossy(inspect))
```

artichoke-backend/src/value.rs
```diff
@@ -367,9 +368,9 @@ impl ValueLike for Value {
         unsafe { sys::mrb_sys_obj_frozen(mrb, inner) }
     }
 
-    fn inspect(&self) -> String {
-        self.funcall::<String>("inspect", &[], None)
-            .unwrap_or_else(|_| "<unknown>".to_owned())
+    fn inspect(&self) -> Vec<u8> {
+        self.funcall::<Vec<u8>>("inspect", &[], None)
+            .unwrap_or_else(|_| Vec::new())
```
lopopolo: this can line can be simplified to `unwrap_or_default()`

artichoke-backend/src/value.rs
```diff
@@ -381,9 +382,9 @@ impl ValueLike for Value {
         self.funcall::<bool>("respond_to?", &[method], None)
     }
 
-    fn to_s(&self) -> String {
-        self.funcall::<String>("to_s", &[], None)
-            .unwrap_or_else(|_| "<unknown>".to_owned())
+    fn to_s(&self) -> Vec<u8> {
+        self.funcall::<Vec<u8>>("to_s", &[], None)
+            .unwrap_or_else(|_| Vec::new())
```
lopopolo: this line can be simplified to `unwrap_or_default()`

artichoke-backend/src/value.rs
```diff
@@ -395,7 +396,8 @@ impl Convert<Value, Value> for Artichoke {
 
 impl fmt::Display for Value {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.to_s())
+        let string_repr = String::from_utf8_lossy(&self.to_s()).to_string();
```
lopopolo: ditto above it is not necessary to turn the `Cow` into an owned `String`

`String::from_utf8_lossy` only allocates a `String` if the byte slice was not UTF-8. By keeping the byte vector alive, we keep the `Cow::Borrowed` alive.

Assigning self.to_s() to a let binding allows us to avoid the `to_string`.

I also prefer calling `.as_slice()` instead of relying on deref coercion to turn `Vec`s and `String`s into slices because I find it to be more explicit and more clearly communicate intent.

artichoke-backend/src/extn/core/regexp/mod.rs
lopopolo: See GH-415.

artichoke-backend/src/extn/core/kernel/mod.rs
lopopolo: See GH-415.

PR #407, opened 2019-12-04
[WIP] Implement String#index with bstr

### LINKED ISSUE #323: Implement String#index with bstr for String patterns
String#index is a specialized version of String#scan.

It might be possible to implement String#index in terms of String#scan, but it will likely be more efficient to create a specialized impl.

aivensong:
Closes #323. 

This PR implements String#index in Rust. Functionally seems to be working, opening a WIP PR to get some initial feedback. 

TODO:
- [ ] Tests

### DIFF
--- a/artichoke-backend/src/extn/core/string/index.rs
+++ b/artichoke-backend/src/extn/core/string/index.rs
@@ -14,25 +14,28 @@ fn start_index(
     interp: &Artichoke,
     string: &[u8],
     offset: Option<Value>,
-) -> Result<i64, Box<dyn RubyException>> {
+) -> Result<Option<usize>, Box<dyn RubyException>> {
     if let Some(offset) = offset {
-        let mut start = if let Ok(start) = offset.clone().try_into::<Int>() {
-            start
-        } else if let Ok(start) = offset.funcall::<Int>("to_int", &[], None) {
-            start
+        let index = offset.implicitly_convert_to_int()?;
+        let index = if index < 0 {
+            let idx = usize::try_from(-index).map_err(|_| {
+                Fatal::new(interp, "Expected positive position to convert to usize")
+            })?; 
+            string.len().checked_sub(idx)
         } else {
-            let offset_type_name = offset.pretty_name();
-            return Err(Box::new(TypeError::new(
-                interp,
-                format!("no implicit conversion {} into Integer", offset_type_name),
-            )));
+            let idx = usize::try_from(index).map_err(|_| {
+                Fatal::new(interp, "Expected positive position to convert to usize")
+            })?;
+            Some(idx)
         };
-        if start < 0 {
-            start += string.len() as i64;
+        if let Some(idx) = index  {
+            if idx > string.len() {
+                return Ok(None);
+            }
         }
-        Ok(start)
+        Ok(index)
     } else {
-        Ok(0)
+        Ok(Some(0_usize))
     }
 }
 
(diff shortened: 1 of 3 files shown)


artichoke-backend/src/extn/core/string/index.rs
```diff
@@ -0,0 +1,100 @@
...
+            interp,
+            "Unable to convert Ruby String Receiver to Rust byte slice",
+        )
+    })?;
+    if let Ruby::Symbol = pattern.ruby_type() {
+        Err(Box::new(TypeError::new(
+            interp,
+            format!("type mismatch: {} given", pattern.pretty_name()),
+        )))
+    } else if let Ok(regexp) = unsafe { Regexp::try_from_ruby(interp, &pattern) } {
```
lopopolo: extracting a `Regexp` from the pattern value is not free. Can we move the byte slice case above this one?

artichoke-backend/src/extn/core/string/index.rs
```diff
@@ -0,0 +1,100 @@
...
+        let start = start_index(interp, string, offset)?;
+        if start < 0 || start > string.len() as i64 {
+            return Ok(interp.convert(None::<Value>));
+        }
+        let start = usize::try_from(start).unwrap();
```
lopopolo: please explicity handle failed `TryFrom` conversions, even if all we do is return `Fatal`.

artichoke-backend/src/extn/core/string/index.rs
```diff
@@ -0,0 +1,100 @@
...
+        if let Some((begin, _)) = regexp.borrow().inner().pos(interp, &string[start..], 0)? {
+            let result = begin + start;
```
lopopolo: add a comment that this can never wrap because the pattern is guaranteed to be less than `usize` bytes.

artichoke-backend/src/extn/core/string/index.rs
```diff
@@ -0,0 +1,100 @@
...
+            let result = Int::try_from(result)
```
lopopolo: Result is measured in bytes. Does Ruby return byte index or character index?

It looks like char index:

```
$ irb
[2.6.5] > "💎a".index("a")
=> 1
```

artichoke-backend/src/extn/core/string/index.rs
```diff
@@ -0,0 +1,100 @@
...
+                .map_err(|_| Fatal::new(interp, "Index pos does not fit in Integer"))?;
+            Ok(interp.convert(result))
+        } else {
+            Ok(interp.convert(None::<Value>))
+        }
+    } else {
+        let pattern_type_name = pattern.pretty_name();
+        let pattern_bytes = match (
+            pattern.clone().try_into::<&[u8]>(),
+            pattern.funcall::<&[u8]>("to_str", &[], None),
```
lopopolo: `Value::funcall` calls into the mruby VM. This operation is not free! The fast and most common path is when pattern _is_ a `String`, so let's attempt to process that case first. The only call into the VM if we need to.

artichoke-backend/src/extn/core/string/index.rs
```diff
@@ -0,0 +1,100 @@
...
+        ) {
+            (Ok(a), Ok(_)) | (Ok(a), Err(_)) => a,
+            (Err(_), Ok(b)) => b,
+            (Err(_), Err(_)) => {
+                return Err(Box::new(TypeError::new(
+                    interp,
+                    format!("type mismatch: {} given", pattern_type_name,),
+                )))
+            }
```
lopopolo: processing things sequentially also significantly simplifies this match condition.

artichoke-backend/src/extn/core/string/index.rs
```diff
@@ -0,0 +1,100 @@
...
+        let pattern_bytes = match (
+            pattern.clone().try_into::<&[u8]>(),
+            pattern.funcall::<&[u8]>("to_str", &[], None),
+        ) {
+            (Ok(a), Ok(_)) | (Ok(a), Err(_)) => a,
+            (Err(_), Ok(b)) => b,
+            (Err(_), Err(_)) => {
+                return Err(Box::new(TypeError::new(
+                    interp,
+                    format!("type mismatch: {} given", pattern_type_name,),
```
lopopolo: nit: there is a trailing comma in the `format!` args.

artichoke-backend/src/extn/core/string/index.rs
```diff
@@ -0,0 +1,100 @@
...
+                )))
+            }
+        };
+        let start = start_index(interp, string, offset)?;
+        if start < 0 || start > string.len() as i64 {
+            return Ok(interp.convert(None::<Value>));
+        }
+        let start = usize::try_from(start).unwrap();
```
lopopolo: please explicity handle failed TryFrom conversions, even if all we do is return Fatal.

artichoke-backend/src/extn/core/string/index.rs
```diff
@@ -0,0 +1,100 @@
...
+        let string = &string[start..];
+        if let Some(result) = string.find(pattern_bytes) {
+            let result = result + start;
+            let result = Int::try_from(result)
+                .map_err(|_| Fatal::new(interp, "Index pos does not fit in Integer"))?;
```
lopopolo: Ditto above, this result is in bytes but Ruby returns it in characters.

I'm not sure if this is easy to do given we have bytes not a str. Maybe bstr can help? I'm happy if we leave this as a TODO.

artichoke-backend/src/extn/core/string/index.rs
```diff
@@ -0,0 +1,100 @@
...
+    pattern: Value,
+    offset: Option<Value>,
+) -> Result<Value, Box<dyn RubyException>> {
+    let string = value.clone().try_into::<&[u8]>().map_err(|_| {
+        Fatal::new(
+            interp,
+            "Unable to convert Ruby String Receiver to Rust byte slice",
+        )
+    })?;
+    if let Ruby::Symbol = pattern.ruby_type() {
```
lopopolo: thank you for catching this edge case. I haven't found a good way to encapsulate this distinction in the converters yet. Any ideas?

lopopolo:
To run the ruby specs, you can add a line to `scripts/spec.rb` like this one:

https://github.com/artichoke/artichoke/blob/830bb3026f21a9246a04f92e80d95b6d1dc94357/scripts/spec.rb#L216

```ruby
  runner.register(Spec.new('core', 'string', 'index'))
```

aivensong:
@lopopolo thanks for the thorough feedback! Been busy lately but going to address your feedback this week if not this weekend.

artichoke-backend/src/extn/core/string/index.rs
aivensong: I'll get back to you after I feel like I understand converters / Artichoke in general more 😅

lopopolo:
This PR is stale and there has been a bit of churn since this branch last had activity.

@aivensong  if you'd like to pick this up again, please open a new PR. For now, closing.

Issue #410, opened 2019-12-08
Dividing two integers returns a float


CodeGradox:
While testing artichoke against ruby 2.6.3 I noticed I got different results when running a program. Turns out artichoke returns a float instead of an int when dividing two integers.
Is this intentional or a bug?

lopopolo:
Hi @CodeGradox thanks for reporting this. This is both a limitation of Artichoke (missing method impl in Fixnum) and a limitation of the mruby VM Artichoke build atop (which we embed and actively modify).

Both of these deficiencies are fixed in GH-411.

CodeGradox:
Thanks for the quick fix.
I was working on advent of code and was a bit disappointed with ruby's performance so I wanted to compare with artichoke. There as still a few features missing so I can't run most of my solutions.
Maybe artichoke will be good enough for the next years calendar.

Keep up the good work! :+1:

PR #411, opened 2019-12-08
Integer division

lopopolo:
@CodeGradox Thanks for reporting GH-410. 💪 💎 

This PR overrides the implementation of `/` method attached to `Integer` class.

This PR also modifies the implementation of the `OP_DIV` opcode in the murby VM.

For speed, mruby pulls `/` out of the normal method dispatch flow and attempts to process operands as `Integer`s and `Float`s before falling back to method dispatch. This approach was hardcoded to return `Float` and promoted all `Integer` operands to `Float`s. This PR adds special handling for the two `Integer` operand case and adds a `raise ZeroDivisionError` in case of a zero divisor.

```console
$ RUST_BACKTRACE=1 cargo run --bin airb
artichoke 0.1.0 (2019-12-08 revision 2171) [x86_64-darwin]
[Rust 1.39.0 (rev 4560ea7) on x86_64-apple-darwin]
>>> x = 3
=> 3
>>> y = 2
=> 2
>>> x.send('/', y)
=> 1
>>> y.send('/', x)
=> 0
>>> x / y
=> 1
>>> y / x
=> 0
>>>
```

This code doesn't pass ruby/spec because Artichoke doesn't have BIGINT support.

Closes GH-410.

### DIFF
--- a/artichoke-backend/src/extn/core/integer/div.rs
+++ b/artichoke-backend/src/extn/core/integer/div.rs
@@ -30,3 +30,38 @@ pub fn method(
         )))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use artichoke_core::eval::Eval;
+    use artichoke_core::value::Value;
+    use quickcheck_macros::quickcheck;
+
+    use crate::types::Int;
+
+    #[quickcheck]
+    fn integer_division_vm_opcode(x: Int, y: Int) -> bool {
+        let interp = crate::interpreter().expect("init");
+        let mut result = true;
+        match (x, y) {
+            (0, 0) => result &= interp.eval(b"0 / 0").is_err(),
+            (x, 0) | (0, x) => {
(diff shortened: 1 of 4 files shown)


artichoke-backend/src/extn/core/integer/div.rs
```diff
@@ -0,0 +1,109 @@
...
+    } else if let Ok(y) = other.try_into::<types::Float>() {
+        if y == 0.0 {
+            if x > 0 {
+                Ok(interp.convert(Float::INFINITY))
+            } else if x < 0 {
+                Ok(interp.convert(Float::NEG_INFINITY))
+            } else {
+                // x == 0
+                Ok(interp.convert(Float::NAN))
+            }
```
lopopolo: not sure if this is necessary or if Rust float division does the right thing here.

PR #412, opened 2019-12-09
Sync Array FFI prototypes from ffi to C header

lopopolo:
There was drift between the definitions in array/ffi.rs and
artichoke.h in mruby-sys. clang was more forgiving of this
than emscripten.

This change is necessary to deploy the playground.


lopopolo:
Filed GH-413 to finally address the build drift.

PR #435, opened 2020-01-19
Make spec constructors and builders fallible

lopopolo:
All specs own a `CString` of their name to pass into mruby.

Previously, turning the `name: &str` parameter into a `CString` called `expect` on the `Result`. This method is fallible if the `name` contains a `NUL` byte.

GH-434 made `Context` creation fallible for the same reasons. Pushing errors out to callers made `Kernel#require` behavior more predictable.

GH-430 attempted to solve this issue for class and module specs by using an unsafe API to provide a default `CString`. This masked failure with an implicit default rather than allowing callers to know they provided bad input.

This PR makes `class::Spec::new`, `module::Spec::new`, `method::Spec::new`, and the associated method builder functions on `class::Builder` and `module::Builder` fallible (return `Result<Self, ArtichokeError>`).

In practice this means inserting more `?` operators in `init` methods.

Two bonus changes:

- Remove stale clippy allow in `extn::core::artichoke`.
- Make `method::Spec.name` take a Cow to avoid an allocation when passing in a `&'static str`.

### DIFF
--- a/artichoke-backend/scripts/auto_import/rust_glue.rs.erb
+++ b/artichoke-backend/scripts/auto_import/rust_glue.rs.erb
@@ -5,10 +5,10 @@ use crate::{Artichoke, ArtichokeError};
 pub fn init(interp: &Artichoke) -> Result<(), ArtichokeError> {
     <% constants.each do |(constant, klass)| %>
     <% if klass == "Class" %>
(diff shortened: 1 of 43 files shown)


lopopolo:
Not sure what happened with the labels there. Was getting a bunch of unicorn 500 pages when trying to submit this PR.

Issue #340, opened 2019-11-06
Implement RubyException for artichoke_backend::exception::Exception


lopopolo:
Complete the migration to boxed trait objects for returning errors.

This ticket should only touch the `ArtichokeError` enum to make the types match. It is not necessary to migrate away from `ArtichokeError`.

Issue #43, opened 2019-05-23
Turn MRB_TT_EXCEPTION Values into MrbError::Exec


lopopolo:
If an exception is thrown in the context of a C API call, the returned `mrb_value` is of type `mrb_vtype::MRB_TT_EXCEPTION`. Turn this exception into an `Err(MrbError::Exec(_))`.

Maybe as part of a `funcall` interface on `Value`?

Issue #436, opened 2020-01-20
Investigate whether calls to mrb_yield need to be wrapped in mrb_protect


lopopolo:
Is the implementation of `Block::yield_arg` safe if the `Block` raises?

lopopolo:
This appears to work

```console
>>> Array.new(10) { raise '10' }
Traceback (most recent call last)
	2: from (airb):6
	1: from (airb):6:in initialize
RuntimeError (10)
```

PR #437, opened 2020-01-20
Return raisable Exceptions from Eval::eval and friends

lopopolo:
This is a large PR that touches nearly every file in `artichoke-backend`. This is another refactor the world PR similar to GH-311.

This PR changes the core error type in `extn`, `eval::Eval`, `Block::yield_arg`, and `Value::funcall` to return an opaque `Exception` type that implements `RubyException`.

All errors returned by code that executes or evaluates code on the interpreter are re-raisable via `artichoke_backend::exception::raise`. This means errors when yielding to a block can be passed back out to mruby trampolines, `Kernel::require` can catch, propagate, and re-raise errors without losing backtraces or causing UB via `longjmp` unwinding.

This PR removes `Eval::unchecked_eval` and `Value::unchecked_funcall`.

This PR introduces `artichoke_backend::extn::prelude` which is intended to be glob imported in extn code to make sweeping refactors like this easier in the future. It has enabled removing several legacy re-exports in `artichoke-backend`.

As a bonus, this PR adds ANSI coloring to `airb` traceback output to match MRI `irb`.

Followup: GH-436 to investigate whether B`lock::yield_arg` needs to be wrapped in an `mrb_protect`.

This PR undoes GH-306 to add `Value::unchecked_funcall`.

Fixes GH-340.
Fixes GH-43.


lopopolo:
Now that `RubyException`s are plumbed all the way through, testing exceptions in Rust tests is much more ergonomic.

lopopolo:
This PR adds another prelude in the `test` module, conditionally compiled only for test builds, that includes many common imports for tests. This again enables sweeping refactors.

Issue #116, opened 2019-08-04
Add global variable get and set APIs to artichoke-core


lopopolo:
These are needed by `Regexp` and `MatchData` for setting special Regexp globals.

lopopolo:
Running into this a lot when refactoring in the `RegexpType` backends. They access the raw `mrb_state` a lot for globals and they shouldn't have to.

lopopolo:
Fixed by GH-562.

PR #442, opened 2020-01-24
[WIP] Destructure State

lopopolo:
This PR was an experiment and I do not intend to merge it.

This PR was born out of an attempt to implement a `Symbol` table in Rust and attach it to the inner `State` on the `Artichoke` interpreter. This turned out to be impossible (or at least extremely difficult). Most code that interacts with the `State` does not really pass borrowck, enabled by the State being wrapped in a `RefCell`. The worst offenders are APIs like `Class::rclass` which are accessed via a borrow on the `State` and take an `Artichoke`.

Thus, the goal was to get rid of the `Rc<RefCell<_>>` and get normal borrowing semantics.

To achieve this, the `Artichoke` type was rewritten to have two fields: a `NonNull<mrb_state>` and a `Box<State>`. This was mistake number one because I now had to rewrite all of artichoke-backend to deal with the covariants of the `mrb` member.

The `Box<State>` was moved in and out of the userdata pointer on the `mrb_state` when crossing over FFI boundaries. This mostly worked, but introduced lots of panicking pathways.

Borrowing rules with one big state meant that APIs had to be moved into the state to access local borrows while also using an `&mut mrb_state`. `RustBackedValue` was rewritten to be a veneer over calls to the `State`.

`Regexp` code is awful. It heavily relies on dropping down to the C API level of abstraction which made refactoring super tedious. The layout of the `MatchData` code with one file per trampoline also made refactoring tedious.

Positives include lots of unsafe code disappearing. The `NonNull` abstraction is really great.

All APIs that mutate interpreter state now take an interpreter!. This is much more explicit at the cost of some ergonomics (e.g. `Value::funcall`).

By virtue of touching every file in `extn` I discovered some opportunities to refactor, like using `implicitly_convert_to_int` or redesigning backend APIs to avoid dependencies on the interpreter. I'll try to extract some of these changes.

Ultimately, this PR failed because it tried to do too much. Similarly large PRs in lines of code (GH-311, GH-437) had singular purpose in migrating the error abstraction. This PR redesigned the state, rewrote `extn`, altered FFI semantics, and changed mutability requirements of nearly every API.

### DIFF
--- a/artichoke-backend/scripts/auto_import/rust_glue.rs.erb
+++ b/artichoke-backend/scripts/auto_import/rust_glue.rs.erb
@@ -1,13 +1,13 @@
 use crate::extn::prelude::*;
 
-pub fn init(interp: &Artichoke) -> InitializeResult<()> {
+pub fn init(interp: &mut Artichoke) -> InitializeResult<()> {
     <% constants.each do |(constant, klass)| %>
     <% if klass == "Class" %>
     let spec = crate::class::Spec::new("<%= constant %>", None, None)?;
-    interp.0.borrow_mut().def_class::<<%= constant %>>(spec);
+    interp.state_mut().def_class::<<%= constant %>>(spec);
     <% elsif klass == "Module" %>
     let spec = crate::module::Spec::new("<%= constant %>", None)?;
-    interp.0.borrow_mut().def_module::<<%= constant %>>(spec);
+    interp.state_mut().def_module::<<%= constant %>>(spec);
     <% else %>
     // Skipping constant <%= constant %> with class <%= klass %>
     <% end %>
--- a/artichoke-backend/scripts/auto_import/rust_glue.rs.erb
+++ b/artichoke-backend/scripts/auto_import/rust_glue.rs.erb
@@ -6,7 +6,7 @@ pub fn init(interp: &mut Artichoke) -> InitializeResult<()> {
     let spec = crate::class::Spec::new("<%= constant %>", None, None)?;
     interp.state_mut().def_class::<<%= constant %>>(spec);
     <% elsif klass == "Module" %>
-    let spec = crate::module::Spec::new("<%= constant %>", None)?;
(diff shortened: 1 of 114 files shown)


lopopolo:
There are a few remaining pieces before this PR will be resolved:

- Add global variable get and set APIs GH-116.
- Modify `Regexp` and `MatchData` implementations to use global variable APIs.

Once that happens, the only dependencies on `sys` in `extn` will be FFI glue for C trampolines.

The strategy for removing the `Rc` wrapper from the `State` and instead using a plain `Box` is:

- The public API type `Artichoke` will have two private fields:
  - `Option<Box<State>>`
  - `Option<NonNull<sys::mrb_state>>`
- Calling `Artichoke::into_raw`:
  - Moves the `Box<State>` out of the option, calls `Box::into_raw` on it, and serializes the state into the `mrb_state` `ud` (userdata) pointer.
  - Moves the `mrb_state` out of the option.
- `Artichoke::into_raw` is called before crossing over an FFI boundary, e.g. `Eval::eval`, `Value::funcall`, or `Block::yield_arg`.
- When rematerializing an interpreter when returning to Rust code from an FFI boundary, initialize a new `Artichoke` by moving the `Box<State>` out of the `ud` pointer.

It is necessary to allow for partially initialized `Artichoke` (and returning `InterpreterExtractError`) both in the public API _and_ in `extn` code because both of these context can call back into C.

Reusing the same type in each context also allows deferring the migration to an existential type for the interpreter in `extn` function signatures.

lopopolo:
Another missing non-trivial piece is removing the `Artichoke` member from `Value` which requires threading the interpreter parameter through all `Value` functions like `Value::implicitly_convert_to_int`.

lopopolo:
Realizing the new approach is interior mutability by another name to move in and out of FFI.

lopopolo:
Fixed by GH-670.

PR #444, opened 2020-01-24
Convert some syntax in converters to preferred style

lopopolo:
`Vec::<String>::new()` instead of `<Vec<String>>::new()`.


lopopolo:
Extracted from GH-442.

PR #445, opened 2020-01-24
chore: cargo update

lopopolo:
this update removes a bunch of versions of `rand_core` from the lockfile.


lopopolo:
The only thing left bringing in `failure` is the old version of `bindgen` pulled in by `onig_sys`.

PR #449, opened 2020-01-25
Minimize unsafe in ExceptionHandler::last_error

lopopolo:
Use Rust patterns like mem::replace and NonNull to minimize unsafe in
ExceptionHandler impl to one pointer deref and one FFI funcall.

Add safety proofs.


lopopolo:

PR #450, opened 2020-01-25
Replace to_int funcalls in Rust with Value::implicitly_convert_to_int

lopopolo:


lopopolo:
Spun off GH-452 to do this same exercise for `String`.

PR #455, opened 2020-01-26
Remove dependency on artichoke_core::value::Value in artichoke-backend converters

lopopolo:
Remove instances of value.try_into and instead use the TryConverter trait, sometimes
with assisted inference.


lopopolo:

PR #470, opened 2020-01-29
Move low-level protect code into sys::protect module

lopopolo:
This commit extracts the low-level `mrb_protect` API out of
`Value::funcall`, `Eval::eval`, and `Block::yield_arg`. Module
`sys::protect` exposes functions for each use case that call the
appropriate mruby APIs and wrap each interpreter call in `mrb_protect`
to prevent `longjmp` unwinds from unwinding Rust frames.

This fixes a bug where `mrb_protect` was not present for block yields.

The `sys::protect` family of functions properly extract exceptions from
the interpreter and return `Result<sys::mrb_value, sys::mrb_value>`.
This signigifcantly simplifies the exception handler which has been
refactored into a function instead of a trait.

This change makes explicit an exception handling step that worked by
accident before: some errors like `SyntaxError` do not get returned out
of `mrb_protect` and are only present on the `exc` property of the
`mrb_state`.

This commit makes use of a "move out of `Box` and destructure" pattern I recently learned about.

This change significantly reduces the amount of unsafe.

### DIFF
--- a/artichoke-backend/src/eval.rs
+++ b/artichoke-backend/src/eval.rs
@@ -1,54 +1,11 @@
 use artichoke_core::eval::Eval;
-use std::ffi::c_void;
-use std::mem;
 
 use crate::exception::Exception;
-use crate::exception_handler::ExceptionHandler;
+use crate::exception_handler;
 use crate::extn::core::exception::Fatal;
-use crate::sys::{self, DescribeState};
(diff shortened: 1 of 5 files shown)


lopopolo:
This PR helps the refactoring work in GH-442 by minimizing the `&mut self` surface area.

Issue #486, opened 2020-02-01
Rethink converters


lopopolo:
The macro-based approach to generating converters for collections was a convenient way to get coverage, but it is showing some weaknesses.

## Implicit `Clone` trait bound

https://github.com/artichoke/artichoke/blob/d1c4ff68a5f46259b409ef93a6080bad1eed7c43/artichoke-backend/src/convert/array.rs#L60

## Lots of unnecessary allocations

Funneling all derived impls down to the base `Value` case means lots of marshalling in and out of collections.

## Blanket impls limit extensibility

It is hard to add impls for new e.g. reference types.

# Approach?

At the very least, prune the set of converters to the bare minimum required to implement Artichoke and see if that makes it feasible to manually implement the traits.

PR #487, opened 2020-02-01
Remove macro-generated converter implementations

lopopolo:
Remove the 1000+ Convert and TryConvert implementations. Remove the
`artichoke-all-converters` feature.

Manually implement all converters required by the current application.

Converters are specialized for each type and do not funnel through the
base `Value` converter for collections.

This change is in preparation for converter work falling out of GH-442
which will require converters that allocate on the interpreter heap as
`&mut self`.

This patch removes a bunch of code and speeds up compile times. (Shaves
5s or 10% off a clean build on my machine.)

Fixes GH-486.

### DIFF
--- a/artichoke-backend/Cargo.toml
+++ b/artichoke-backend/Cargo.toml
@@ -55,6 +55,5 @@ default-features = false
 
 [features]
 default = ["artichoke-random", "artichoke-system-environ"]
-artichoke-all-converters = []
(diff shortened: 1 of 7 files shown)


lopopolo:
More performance opts in the new commits, implementing `FromIterator` for `InlineBuffer` and removing intermediate allocations in the converters.

PR #488, opened 2020-02-02
Split converter traits by whether they allocate on the interpreter heap

lopopolo:
This PR is followup to GH-442.

This PR introduces `ConvertMut` and `TryConvertMut` traits that mirror `Convert` and `TryConvert` except the take `&mut self`. The `Mut` varieties should be used when the conversion requires allocating on the interpreter heap.

For example, `Copy` types like `bool` or `u8` can be stored inline in the `sys::mrb_value` without allocating. These types of conversions would impl `Convert`. Types like `&[u8]` and `HashMap<bool, i64>` allocate objects on the interpreter heap so they impl `ConvertMut`.

The `ConvertMut` split has caused more function signature changes in `extn`.

This PR simplifies the integer converters by funelling though `Int::from` whenever possible.

This PR fixes several bugs when setting `Regexp` globals after a match in the `RegexUtf8` backend where the `$1`, `$2`, etc globals would be set to capture 0.

### DIFF
--- a/artichoke-backend/src/convert.rs
+++ b/artichoke-backend/src/convert.rs
@@ -10,7 +10,11 @@ mod string;
 
 /// Re-export from [`artichoke_core`](artichoke_core::convert::Convert).
 pub use crate::core::convert::Convert;
+/// Re-export from [`artichoke_core`](artichoke_core::convert::ConvertMut).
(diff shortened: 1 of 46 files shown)


lopopolo:
Followup to GH-487.

Issue #490, opened 2020-02-02
Add regex::Binary implementation of RegexType


lopopolo:
`regex` crate has a `Binary` regex type that can match on `&[u8]` instead of `&str`. Add an implementation of `RegexpType` that uses this as a backend.

Add a unified `regex`-based backend that delegates to `regex_binary` or `regex_utf8` based on whether the haystack input is valid UTF-8.

This should follow a similar approach to `InlineBuffer` or `Rand` where a specialized API based on `&str` is added to `RegexUtf8` which the `RegexpType` implementation delegates to after input validation. The unified `Regex` backend would call these specialized APIs to avoid double UTF-8 validation.

lopopolo:
Dupe of https://github.com/artichoke/artichoke/issues/136?

Issue #489, opened 2020-02-02
Make oniguruma an optional dependency


lopopolo:
Oniguruma is a C-based dependency and is not necessary for most `Regexp` functionality.

Add `Oniguruma` dependency and `RegexpType` implementation behind an optional feature that is enabled by default.

lopopolo:
I think the way I'd like to lay out the features:

- `core-regexp-regex`: optional but enabled by default, activates the Rust `regex` crate backend.
- `core-regexp-oniguruma`: optional but enabled by default, activates the `onig` crate backend.

When either one or the other backend is active, only construct regexp backends of this type.
When no backends are active, `regexp` module is disabled and not built.
When both backends are active, the existing hybrid backend is used.

Expose two builds of the `spec-runner` to enable testing both regexp backends.

PR #491, opened 2020-02-02
Clean up extn::core::regexp

lopopolo:
Some refactorings to make the intent of the `Regexp` core implementation more clear.

- Move `RegexpType` to the backend module to be more similar to other core types like `ENV`, `Array`, and `Random`.
- Add type alias for complex types and remove clippy allows for type complexity for code hygiene.
- Remove unimplemented `RegexpType` backend modules. These were empty files.
- Filed tickets to track implementation of desired modules: GH-489, GH-490.
- Created a new `regexp::backend::regex` module to reflect the desired cluster of `regex` crate-based implementations: GH-490.
- Move `RegexUtf8` to `backend::regex::utf8`.


lopopolo:
This is prep for the massive task of migrating all 92 calls of `sys::mrb_gv_set` to a new global variable abstraction in `extn::regexp` to work on GH-442.

PR #500, opened 2020-02-10
Fix std::error deprecations

lopopolo:
- Remove `description` implementations. `description` is soft
  deprecated.
- Convert `cause` implementations to `source`. `cause` is deprecated and
  `source` supports downcasting.
- Remove empty `source` implementations.
- Inline static str when code previously called `description`.
- Rename `RubyException::backtrace` to `RubyException::vm_backtrace` to
  address upcomming `std::backtrace` name reservation.
- Add `WriteError::into_inner` for extracting inner `fmt::Error` for use
  in `Debug` and `Display` impls.
- Replace `String::from_utf8_lossy` with `string::escape_unicode` in
  several `Debug` and `Display` impls.
- Add a `std::error::Error` supertrait to `RubyException`.
- Add missing error impls.
- Update frontend backtrace callsite.

### DIFF
--- a/artichoke-backend/src/def.rs
+++ b/artichoke-backend/src/def.rs
@@ -213,15 +213,7 @@ impl fmt::Display for ConstantNameError {
     }
 }
 
-impl error::Error for ConstantNameError {
-    fn description(&self) -> &str {
-        "Invalid constant name contained a NUL byte"
-    }
-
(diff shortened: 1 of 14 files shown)


lopopolo:
The warning on `RubyException::backtrace` comes from making the `std::error::Error` super trait bound since the addition of `std::backtrace` will add a `backtrace` method to `Error`.

PR #501, opened 2020-02-10
Rename HashOfStringToArrayOfInt type alias to NameToCaptureLocations

lopopolo:
The new name actually names the type by its purpose, which makes the code easier
to read.


lopopolo:
This type alias lives in the Regexp backend trait.

PR #504, opened 2020-02-11
Update bindgen to latest (0.53.1)

lopopolo:
This change pulls in artichoke/rust-onig@11bc679 which similarly updates
bindgen in our fork of `onig_sys`.

This upgrade took a while because `artichoke-backend` depends on
`bindgen` with no default features and 0.52.0 added a new `runtime`
feature which is required to have `clang-sys` load libclang correctly.


lopopolo:
while I'm mucking around with deps, running a cargo update too.

PR #512, opened 2020-02-12
Return Exception error type from TryConvert implementations

lopopolo:
This PR continues the push toward removing the `ArtichokeError` type and moving toward a unified error abstraction where all APIs return a variant of `dyn RubyException`.

This PR introduces two new error types in `artichoke_backend::convert`:

- `UnboxRubyError`: returned when failing to unbox a Ruby `Value` into a Rust native type. Raisable as a Ruby `TypeError`.
- `BoxIntoRubyError`: returned when failing to box a Rust value into a Ruby `Value`. This usually occurs when the Ruby box does not have sufficient capacity to store the Rust value. For example, trying to box a `u64` on wasm32. Raisable as a Ruby `TypeError`.

To wire up this change, the converter traits in `artichoke-core` have been modified to encode the error type as an associated type.

`ValueLike::itself` has been removed.

This PR significantly simplified most converter implementations, eliminated many panicking branches, and allowed errors to be propagated as is instead of wrapped.

One thing to look into is whether `RubyException::message` should return a `Cow<[u8]>`.

Expect a followup PR to remove `ArtichokeError` and `BootError`.

### DIFF
--- a/artichoke-backend/src/convert.rs
+++ b/artichoke-backend/src/convert.rs
@@ -62,6 +62,7 @@ pub struct UnboxRubyError {
 }
 
 impl UnboxRubyError {
+    #[must_use]
     pub fn new(value: &Value, into: Rust) -> Self {
         Self {
             from: value.ruby_type(),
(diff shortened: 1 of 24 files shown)


lopopolo:
This PR removes now-unnecessary `map_err` transformations on `Exception`-returning APIs. The only remaining transforms change the semantics of the error, e.g. defining the `Random::DEFAULT` constant returning a `NotDefined::ClassConstant` instead of `NotDefined::Class`.

PR #516, opened 2020-02-13
Handle nested option of Vec without cloning in MatchData#[]

lopopolo:
The iterator expression returns:

    Option<&Option<Vec<u8>>>

This code is refactored to avoid a clone by using a nested pattern and
coercing the resultant `&Vec<u8>` to a slice, which can be correctly
processed by the converter.

This avoids an unnecessary allocation and makes the code easier to read.


lopopolo:
This PR was fixed up in GH-519

PR #519, opened 2020-02-13
Simplify nested Option conversion in MatchData#[]

lopopolo:
Followup to GH-516.

Use `Option::as_deref` to turn `&Option<Vec<u8>>` into `Option<&[u8]>`.
This gives us a nested owned `Option` which lets us call `flatten`.


lopopolo:
this is the net change between GH-516 and this PR:

Screen Shot 2020-02-13 at 12 59 52 AM

Issue #339, opened 2019-11-05
Handle non-UTF-8 OsStr on Windows


lopopolo:
I [asked this question on Twitter](https://twitter.com/artichokeruby/status/1191775611808731142) and got a great answer from @zkat

https://github.com/zkat/srisum-rs/blob/d375909685ae5d100dc5f832b74e09cdf16c3512/src/compute.rs#L23-L42

```rust
// NOTE: This whole dance is to allow us to print out
// the OsStr for the filenames exactly as we received
// them. It's a little bit of overkill, but it guarantees
// the user gets back what they gave and bypasses any
// weird encoding issues with these filenames.
print!("{} ", sri);

#[cfg(unix)]
let output = f.as_bytes();

#[cfg(windows)]
let output: Vec<u16> = f.encode_wide().collect();

let stdout = std::io::stdout();
let mut lock = stdout.lock();
for item in output {
    lock.write_all(&item.to_be_bytes())
        .expect("Failed to write bytes");
}
println!();
```

Our approach will be slightly different because we need a `Vec<u8>` instead of sinking to an `io::Write`.

These two functions need to be modified:

https://github.com/artichoke/artichoke/blob/17502fb91e1f6c2c0bf2462681e2d200685db746/artichoke-backend/src/fs/mod.rs#L142-L158

https://github.com/artichoke/artichoke/blob/17502fb91e1f6c2c0bf2462681e2d200685db746/artichoke-backend/src/fs/mod.rs#L171-L187

In the worst case on Windows, we'll need to allocate a new `Vec<u8>`, so change the signature of all implementations to return `Borrow<'a, [u8]>`.

Windows impl will try to convert to/from `str` first to avoid allocations and fall back to `encode_wide` + `buf.extend(wide.to_be_bytes())` in the non-UTF-8 case.

I also think as part of this change we'll want to move these functions into a platform compat module. They don't really belong in the FS.

The VM and core should only ever deal with `Vec<u8>` and convert as needed to and from `OsStr` at the edges of API boundaries.

lopopolo:
Blocked on GH-338.

lopopolo:
This is no longer blocked now that the windows build is operational

lopopolo:
Now that GH-496 has landed, the right place to add this code is the windows ffi module.

https://github.com/artichoke/artichoke/blob/c2e7710a4099421bf38096e944236e8750fd717e/artichoke-backend/src/ffi/windows.rs#L7-L20

[...]

PR #543, opened 2020-02-26
Replace path converters with bstr functions from ByteSlice

lopopolo:
The `bstr` crate that `artichoke-backend` already depends on implements
converters between `&[u8]` and `&OsStr` in each direction. This replaces
the (identical) implementations in the `ffi` module.

Docs: https://docs.rs/bstr/0.2.11/bstr/#file-paths-and-os-strings

This offloads non-core functionality to a high-quality crate, which is a
maintainability win. This also removes some conditional compilation,
which is another maintainability win.

The public functions for the conversions in FFI remain because they wrap
the `bstr` calls in `Results` with `RubyException` `Err` variants.


lopopolo:
This PR partially closed https://github.com/artichoke/artichoke/issues/339

PR #548, opened 2020-02-29
Bump rust-embed to 5.5.0

lopopolo:
Version 5.5.0 brings in a bugfix that makes paths relative to
`Cargo.toml` even in workspaces. This eliminates the need to copy
`mspec` into the `$OUT_DIR` in `build.rs`.

Because paths are relative to the crate root, the `RustEmbed` derive can
path into `vendor/mspec` directly. This makes hacking on the spec-runner
eaisier since no subsequent builds are required in debug builds to pick
up changes to `mspec`.

This also allows us to drop the `interpolate-folder-path` feature which
removes a transitive dependency on `shellexpand`.

Because there is no build script anymore, build dependencies can go
away.


lopopolo:
Rebase to pull in buildfix from GH-549.

PR #549, opened 2020-03-01
Install cargo-deny from public GitHub release

lopopolo:
Use a blessed build instead of compiling from source. This should
speed up this workflow significantly and remove a failure mode
(compilation failures from mismatched stable compiler and public
release).


lopopolo:
This fixes a build failure in GH-548.

PR #562, opened 2020-03-08
Add core APIs for getting, setting, and removing global variable bindings

lopopolo:
Add `Globals` trait to `artichoke-core` that enables interpreters to get, set, and unset global variables.

This trait wraps mruby with one of the last remaining uses of `unsafe` in `extn` (excluding `mruby` and `ffi` modules).

The core implementation and one prototype of its use in `String#scan` are merged here before tackling `onig` and `utf8` `Regexp` backends in a separate PR.

Followup to GH-557.
Relates to GH-442.


lopopolo:
Fixes GH-116.

PR #572, opened 2020-03-18
Handle non-UTF-8 OsStr on Windows

### LINKED ISSUE #339: Handle non-UTF-8 OsStr on Windows

dylni:
I noticed [this issue](https://github.com/artichoke/artichoke/issues/339), so I thought I would help move it forward and see how my crate handles a larger project.

Closes #339

### DIFF
--- a/artichoke-backend/src/extn/core/env/backend/system.rs
+++ b/artichoke-backend/src/extn/core/env/backend/system.rs
@@ -51,12 +51,9 @@ impl EnvType for System {
             Ok(None)
         } else {
             let name = ffi::bytes_to_os_str(name)?;
-            if let Some(value) = std::env::var_os(name) {
-                let value = ffi::os_str_to_bytes(value.as_os_str())?;
-                Ok(Some(value.to_vec().into()))
-            } else {
-                Ok(None)
-            }
+            Ok(std::env::var_os(name)
+                .map(ffi::os_string_to_bytes)
+                .map(Cow::Owned))
         }
     }
 
@@ -120,14 +117,14 @@ impl EnvType for System {
         }
     }
 
-    fn as_map(&self, interp: &Artichoke) -> Result<HashMap<Vec<u8>, Vec<u8>>, Exception> {
+    fn as_map(&self, interp: &Artichoke) -> HashMap<Vec<u8>, Vec<u8>> {
         let _ = interp;
         let mut map = HashMap::default();
         for (name, value) in std::env::vars_os() {
-            let name = ffi::os_str_to_bytes(name.as_os_str())?;
-            let value = ffi::os_str_to_bytes(value.as_os_str())?;
-            map.insert(name.to_vec(), value.to_vec());
+            let name = ffi::os_string_to_bytes(name);
+            let value = ffi::os_string_to_bytes(value);
+            map.insert(name, value);
         }
-        Ok(map)
+        map
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,6 +49,7 @@ dependencies = [
  "num_cpus",
  "once_cell",
  "onig",
+ "os_str_bytes",
  "quickcheck",
  "quickcheck_macros",
  "rand",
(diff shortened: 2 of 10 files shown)


lopopolo:
Checks not running for PRs from forks is fixed in GH-574.

dylni:
Thanks. I was wondering what was blocking them.

lopopolo:
The crate referred to in the PR description is https://crates.io/crates/os_str_bytes:

> Traits for converting between byte sequences and platform-native strings

lopopolo:
Thanks for the PR @dylni. I'll try to review it tonight (PDT, UTC-7).

dylni:
Great. Thanks for the interest.

artichoke-backend/src/extn/core/env/backend/system.rs
```diff
@@ -120,14 +117,14 @@ impl EnvType for System {
...
 
-    fn as_map(&self, interp: &Artichoke) -> Result<HashMap<Vec<u8>, Vec<u8>>, Exception> {
+    fn as_map(&self, interp: &Artichoke) -> HashMap<Vec<u8>, Vec<u8>> {
         let _ = interp;
         let mut map = HashMap::default();
         for (name, value) in std::env::vars_os() {
-            let name = ffi::os_str_to_bytes(name.as_os_str())?;
-            let value = ffi::os_str_to_bytes(value.as_os_str())?;
-            map.insert(name.to_vec(), value.to_vec());
+            let name = ffi::os_string_to_bytes(name);
```
lopopolo: converting directly from `OsString` _into_ `Vec<u8>` saves an allocation.

artichoke-backend/src/extn/core/env/backend/system.rs
```diff
@@ -51,12 +51,9 @@ impl EnvType for System {
...
         } else {
             let name = ffi::bytes_to_os_str(name)?;
-            if let Some(value) = std::env::var_os(name) {
-                let value = ffi::os_str_to_bytes(value.as_os_str())?;
-                Ok(Some(value.to_vec().into()))
-            } else {
-                Ok(None)
-            }
+            Ok(std::env::var_os(name)
+                .map(ffi::os_string_to_bytes)
```
lopopolo: ditto convert from owned to owned.

dylni:
Ok @lopopolo. `bstr` does still require valid UTF-8 on Windows, but I understand if you want to use a crate that has existed longer for this.

PR #575, opened 2020-03-21
Disable unicode cargo feature in bstr

lopopolo:
Disable `bstr` `unicode` cargo feature since it is unused in
`artichoke-backend`. This change removes some dependencies of `bstr`
from the `Cargo.lock` file.

Remove `bstr` dependency from `artichoke-frontend`. This dependency
likely has not been used since the introduction of the `string` module
in `artichoke-backend`.


lopopolo:
This PR removed 2 deps from `Cargo.lock`.

PR #576, opened 2020-03-21
Disable with-dirs cargo feature in rustyline

lopopolo:
Disable `rustyline` `with-dirs` feature since it is unused in
`artichoke-frontend`. This change removes some dependencies of
`rustyline` from the `Cargo.lock`.

The `with-dirs` feature of `rustyline` activates tab completion behavior
for paths that start with `~`. `airb` uses `rustyline` to input Ruby
source code, not paths, so this optional feature is not being used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,18 +19,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arrayref"
-version = "0.3.6"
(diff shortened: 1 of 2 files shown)


lopopolo:
This PR removes 9 deps from `Cargo.lock`.

Issue #478, opened 2020-01-30
Print backtraces of top-level exceptions in ruby frontend


lopopolo:
MRI `ruby` pretty-prints backtraces when running code:

```console
$ ruby -e 'Integer "\xFF"'
Traceback (most recent call last):
	1: from -e:1:in `<main>'
-e:1:in `Integer': invalid value for Integer(): "\xFF" (ArgumentError)
```

Artichoke does not:

```console
$ ./target/debug/artichoke -e 'Integer "\xFF"'
ArgumentError (invalid value for Integer(): "\xFF")
```

Add backtrace printing to these spots in `artichoke-frontend`:

https://github.com/artichoke/artichoke/blob/6dc6d2f7aad87dc999769d0b7f9b6733ce514edb/artichoke-frontend/src/ruby.rs#L111
https://github.com/artichoke/artichoke/blob/6dc6d2f7aad87dc999769d0b7f9b6733ce514edb/artichoke-frontend/src/ruby.rs#L143
https://github.com/artichoke/artichoke/blob/6dc6d2f7aad87dc999769d0b7f9b6733ce514edb/artichoke-frontend/src/ruby.rs#L194

The MRI backtrace is formatted with bold and underline like this:

<img width="779" alt="Screen Shot 2020-01-29 at 9 15 08 PM" src="https://user-images.githubusercontent.com/860434/73422024-7e742280-42dc-11ea-974f-1c0dc9c91a4d.png">

lopopolo:
This is an example of how something similar is done in the REPL frontend

https://github.com/artichoke/artichoke/blob/6dc6d2f7aad87dc999769d0b7f9b6733ce514edb/artichoke-frontend/src/repl.rs#L172-L198

lopopolo:
Fixed by GH-594, GH-606.

Thanks @afaur 💎

PR #594, opened 2020-03-24
Print backtraces of top-level exceptions in ruby frontend #478

afaur:
Submitted for review to address issue https://github.com/artichoke/artichoke/issues/478
Please let me know if there is a suggestion for a test that would cover this change that I could add.

### DIFF
--- a/src/ruby.rs
+++ b/src/ruby.rs
@@ -8,8 +8,8 @@ use std::io::{self, Read, Write};
 use std::path::{Path, PathBuf};
 use structopt::StructOpt;
 
-use crate::backend::exception::RubyException;
 use crate::backend::exception::Exception;
+use crate::backend::exception::RubyException;
 use crate::backend::extn::core::exception::{IOError, LoadError};
 use crate::backend::ffi;
 use crate::backend::state::parser::Context;
@@ -165,7 +165,10 @@ fn load_error(file: &OsStr, message: &str) -> Result<String, Exception> {
     Ok(buf)
 }
 
-fn handle_exception(interp: &mut artichoke_backend::Artichoke, exc: artichoke_backend::exception::Exception) -> Result<(), Exception> {
+fn handle_exception(
+    interp: &mut artichoke_backend::Artichoke,
+    exc: artichoke_backend::exception::Exception
+) -> Result<(), Exception> {
     if let Some(backtrace) = exc.vm_backtrace(interp) {
         writeln!(
             io::stderr(),
--- a/src/ruby.rs
+++ b/src/ruby.rs
@@ -2,11 +2,13 @@
 //!
 //! Exported as `ruby` and `artichoke` binaries.
 
+use ansi_term::Style;
 use std::ffi::{OsStr, OsString};
-use std::io::{self, Read};
+use std::io::{self, Read, Write};
 use std::path::{Path, PathBuf};
 use structopt::StructOpt;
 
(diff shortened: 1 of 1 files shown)


src/ruby.rs
```diff
@@ -69,7 +71,10 @@ pub fn entrypoint() -> Result<(), Exception> {
         io::stdin()
             .read_to_end(&mut program)
             .map_err(|_| IOError::new(&interp, "Could not read program from STDIN"))?;
-        let _ = interp.eval(program.as_slice())?;
```
lopopolo: the `let _ = ...` here is saying: I don't want to deal with this error. Now we do want to deal with this error 🎉 

Instead of `map_err`, I think the code would communicate that intent more clearly by looking like the `-e` command handling below. Maybe something like:

```rust
if let Err(exc) = interp.eval(program.as_slice()) {
    handle_exception(&mut interp, &exc)?;
}
```

This code propagates failures to print stack traces the same way we propagate errors to read code from stdin.

src/ruby.rs
```diff
@@ -141,7 +149,11 @@ fn execute_program_file(programfile: &Path, fixture: Option<&Path>) -> Result<()
             }
         }
     };
-    let _ = interp.eval(program.as_slice())?;
+
+    let _ = interp
+        .eval(program.as_slice())
+        .map_err(|exc| handle_exception(&mut interp, &exc));
```
lopopolo: I think this bit of error handling should be structured as above.

src/ruby.rs
```diff
@@ -152,3 +164,38 @@ fn load_error(file: &OsStr, message: &str) -> Result<String, Exception> {
     string::format_unicode_debug_into(&mut buf, &path)?;
     Ok(buf)
 }
+
+fn handle_exception(
+    interp: &mut artichoke_backend::Artichoke,
+    exc: &artichoke_backend::exception::Exception,
```
lopopolo:
```suggestion
    exc: &Exception,
```

src/ruby.rs
```diff
@@ -2,12 +2,14 @@
...
 //! Exported as `ruby` and `artichoke` binaries.
 
+use ansi_term::Style;
 use std::ffi::{OsStr, OsString};
-use std::io::{self, Read};
+use std::io::{self, Read, Write};
 use std::path::{Path, PathBuf};
 use structopt::StructOpt;
 
 use crate::backend::exception::Exception;
```
lopopolo:
```suggestion
use crate::backend::exception::{Exception, RubyException};
```

src/ruby.rs
```diff
@@ -2,12 +2,14 @@
...
+use crate::backend::exception::RubyException;
```
lopopolo:
```suggestion
```

lopopolo:
@afaur I'm not sure if you know this, but you can use `cargo` to auto format Rust code.

If you installed `rust` with `rustup`, you can do:

```sh
rustup component add rustfmt
```

Then you can run:

```sh
cargo fmt
```

lopopolo:
Artichoke doesn't have the concept of a UI test right now (not sure of the exact name here, stealing this terminology from rustc which uses it for testing outputted compiler errors).

I don't know what to do right now, so I filed https://github.com/artichoke/artichoke/issues/595 to investigate.

lopopolo:
Merging as is and following up in a fixup PR. Thanks for the contribution @afaur

PR #606, opened 2020-03-28
Address followups to GH-594 artichoke entrypoint backtrace formatting

lopopolo:
This commit adds some polish to backtrace printing. Followup to GH-594.

- Fix case of `KeyError` message returned from `Hash#fetch`.
- Set exit code on exception in artichoke endpoint even if backtrace is
  successfully printed.
- Clean up imports.
- Use new globals APIs instead of delegating to `sys`, GH-562.
- Rename `handle_exception` to `format_backtrace_into` and pass an
  `io::Write`.
- Make `format_backtrace_into` public.
- Reorder exception message and exception name in output.
- Print top frame with no indent like MRI.

### DIFF
--- a/artichoke-backend/src/extn/core/hash/hash.rb
+++ b/artichoke-backend/src/extn/core/hash/hash.rb
@@ -178,7 +178,7 @@ class Hash
     elsif !not_set
       default
     else
-      raise KeyError, "Key not found: #{key.inspect}"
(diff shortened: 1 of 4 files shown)


lopopolo:
Backtraces from Artichoke compared with MRI:

Screen Shot 2020-03-28 at 7 00 58 AM

PR #607, opened 2020-03-28
Trampoline as VM Compat Layer: Regexp

lopopolo:
Continuation of GH-605. Push conversions of `RegexpType` parameters and outputs out to the trampoline layer in `Regexp` and `String`.

This PR changes the signatures for the internals of `Regexp#=~`, `Regexp#named_captures`, and `String#scan`.

This approach continues to simplify the implementation of the core types and makes them easier to call as standalone APIs.

This code picks up the `pos` and `usize` handling improvements of GH-600 and applies them to `Regexp` and its backends.

This commit fixes a bug where `$&` was not unset if there was no match.

### DIFF
--- a/artichoke-backend/src/extn/core/regexp/backend/mod.rs
+++ b/artichoke-backend/src/extn/core/regexp/backend/mod.rs
@@ -10,6 +10,13 @@ pub mod regex;
 type NilableString = Option<Vec<u8>>;
 type NameToCaptureLocations = Vec<(Vec<u8>, Vec<usize>)>;
 
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub enum Scan {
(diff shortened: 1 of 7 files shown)


artichoke-backend/src/extn/core/regexp/backend/mod.rs
```diff
@@ -8,7 +8,14 @@ pub mod onig;
 pub mod regex;
 
 type NilableString = Option<Vec<u8>>;
-type NameToCaptureLocations = Vec<(Vec<u8>, Vec<Int>)>;
+type NameToCaptureLocations = Vec<(Vec<u8>, Vec<usize>)>;
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub enum Scan {
```
lopopolo: We should implement `ConvertMut` for `Scan`.

PR #610, opened 2020-03-28
Implement Numeric coercion

lopopolo:
This PR pokes around in the numeric core classes. The intent is to implement `Integer`, `Float`, and `Numeric` in Rust.

To achieve this, the `Integer` and `Float` types in `extn::core` are now tuple structs that wrap and underlying primitive type. Ruby core APIs are dangled off this struct. These structs have `Convert`, `ConvertMut`, and `TryConvert` support.

This PR implements `numeric::coerce` which drives all of the numeric conversions in `numeric.c` in MRI.

This PR implements `Float#%` using the new converter and container types. Coercions require returning an arbitrary numeric type from any arithmetic operations, which is represented by `numeric::Outcome` sum type.

1 additional Integer divide spec passes due to this PR.

### DIFF
--- a/artichoke-backend/src/extn/core/float/mod.rs
+++ b/artichoke-backend/src/extn/core/float/mod.rs
@@ -42,8 +42,39 @@ pub fn init(interp: &mut Artichoke) -> InitializeResult<()> {
     Ok(())
 }
 
-#[derive(Debug)]
-pub struct Float;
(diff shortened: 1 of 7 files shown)


lopopolo:
Tagging GH-605 since the ideas with `Outcome` are along these lines.

PR #618, opened 2020-03-30
Reimplement MatchData using modern patterns

lopopolo:
`Regexp` and `MatchData` were the first two `extn` modules. `MatchData`
had mruby C trampolines as associated functions on the main type and one
module per trampoline. While `Regexp` was rewritten and updated to
support multiple backends, `MatchData` has aged.

This PR rewrites `MatchData` from scratch, incorporating the patterns
from GH-605 and GH-610 as well as the updates to `Regexp` in GH-607 and
GH-617.

Significant changes below (unordered):

- Unsafe `is_range` helper moved to private `matchdata::args` module.
- `Region` stores `start` and `end` as `Bound<usize>`.
- `Region` derives `Debug`, `Clone`, `Copy`, `Hash`, `PartialEq`, `Eq`.
- `Region::from_range(range: R)` where `R: RangeBounds<usize>`.
- `Region::offset` to compute offset into the haystack.
- `impl RangeBounds<usize> for Region`.
- Use new `Region` impl accepting a `RangeBounds<usize>` in `onig` and
  `utf8` `Regexp` backends.
- Fix bug in `String::scan` where a `MatchData` is instantiated with a
  full range and immediately calles `MatchData::set_region`.
- Add custom type `Capture` for parsing capture group index for
  `MatchData::begin`, `MatchData::end`, and `MatchData::offset`.
- Add `TryConvert<&'a Value, Capture<'a>>` for extracting capture index
  in trampoline.
- Add custom type `CaptureAt` for parsing capture group index for
  `MatchData::capture_at`.
- Add `TryConvert<&'a Value, CaptureAt<'a>>` for extracting capture index
  in trampoline.
- Improved index handling, see GH-600.
- `MatchData::offset` returns `Option<[usize; 2]>` in the success case.
- `MatchData::begin` and `MatchData::end` implemented with
  `MatchData::offset`.
- All `MatchData` Rust implementations are methods on `MatchData`.
- Added `#[inline]` annotations to `MatchData` methods where
  appropriate.
- Fix bug in `MatchData#offset`, `MatchData#begin`, and `MatchData#end`
  where negative indexes were wrapped with captures length instead of
  raising `IndexError`.
- Panic-free access to haystack in `MatchData::pre` and
  `MatchData::post`.
- Move mruby C trampolines to `matchdata::mruby` module.
- Fix long-standing TODOs about unwinding past `Value`s in `MatchData` C
  entrypoints due to passing references.
- Introduce `matchdata::trampoline` Rust trampolines for argument
  conversion and result boxing. See GH-605.
- Pass through a `&mut Artichoke` everywhere.

### DIFF
--- a/artichoke-backend/src/extn/core/matchdata/args.rs
+++ b/artichoke-backend/src/extn/core/matchdata/args.rs
@@ -1,0 +1,33 @@
+use std::mem;
+
+use crate::extn::prelude::*;
+
+// TODO(GH-308): extract this function into `sys::protect`
+pub unsafe fn is_range<'a>(
+    interp: &'a Artichoke,
+    first: &Value,
+    length: Int,
+) -> Result<Option<(Int, Int)>, TypeError> {
+    use sys::mrb_range_beg_len::*;
+
+    let mut start = mem::MaybeUninit::<sys::mrb_int>::uninit();
+    let mut len = mem::MaybeUninit::<sys::mrb_int>::uninit();
+    let mrb = interp.0.borrow().mrb;
+    // NOTE: `mrb_range_beg_len` can raise.
+    // TODO(GH-308): wrap this in a call to `mrb_protect`.
+    let check_range = sys::mrb_range_beg_len(
+        mrb,
+        first.inner(),
+        start.as_mut_ptr(),
+        len.as_mut_ptr(),
+        length,
+        0_u8,
+    );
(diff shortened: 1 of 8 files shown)


lopopolo:
This PR is part of the yak shave required for GH-442.

This refactor reduces the toil that comes from modifying the `MatchData` implementation.

Issue #477, opened 2020-01-30
Kernel#require and friends LoadError does not escape filenames in exception messages


lopopolo:
When `Kernel#require`, `Kernel#require_relative` and `Kernel#load` cannot read a file, they throw a `LoadError`.

https://github.com/artichoke/artichoke/blob/6dc6d2f7aad87dc999769d0b7f9b6733ce514edb/artichoke-backend/src/extn/core/kernel/require.rs#L328-L332

This code escapes the filename with `bstr` so invalid UTF-8 filenames are escaped with `\xFF`-like sequences.

This is inconsistent with MRI, which does not escape and writes the raw bytes of the filename into the exception.

```console
$ ruby -e 'require "abc-\xFF"'
Traceback (most recent call last):
	2: from -e:1:in `<main>'
	1: from /usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
/usr/local/var/rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- abc-� (LoadError)
```

The unicode replacement character means the invalid bytes were written to the terminal.

This fix is straightforward:

- Construct the message as a `Vec<u8>`
- Pass the vec into `LoadError::new_raw`.

PR #632, opened 2020-04-06
Fix: load_error escaping invalid utf-8

dnamsons:
Closes https://github.com/artichoke/artichoke/issues/477

I hope that this is the proper way to go about solving the issue

### DIFF
--- a/artichoke-backend/src/extn/core/kernel/require.rs
+++ b/artichoke-backend/src/extn/core/kernel/require.rs
@@ -31,7 +31,7 @@ pub fn load(interp: &mut Artichoke, filename: Value) -> Result<Value, Exception>
         api.vfs.is_file(path.as_path())
     };
     if !is_file {
-        return Err(Exception::from(load_error(interp, filename)?));
+        return Err(Exception::from(load_error(interp, filename)));
     }
     let context = Context::new(ffi::os_str_to_bytes(path.as_os_str())?.to_vec())
         .ok_or_else(|| ArgumentError::new(interp, "path name contains null byte"))?;
@@ -127,7 +127,7 @@ pub fn require(
                 .mark_required(path.as_path())
                 .is_err()
             {
-                return Err(Exception::from(load_error(interp, b"internal error")?));
+                return Err(Exception::from(load_error(interp, b"internal error")));
             }
             let mut logged_filename = String::new();
             string::format_unicode_debug_into(&mut logged_filename, filename)?;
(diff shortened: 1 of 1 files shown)


artichoke-backend/src/extn/core/kernel/require.rs
```diff
@@ -281,8 +281,8 @@ impl RelativePath {
...
 }
 
-fn load_error(interp: &Artichoke, filename: &[u8]) -> Result<LoadError, Exception> {
-    let mut message = String::from("cannot load such file -- ");
-    string::format_unicode_debug_into(&mut message, filename)?;
-    Ok(LoadError::new(interp, message))
+fn load_error(interp: &Artichoke, filename: &[u8]) -> LoadError {
+    let mut message = String::from("cannot load such file -- ").into_bytes();
+    message.extend_from_slice(filename);
+    LoadError::new_raw(&interp, message)
```
lopopolo, line 287: The extra `&` is not needed on the `interp`. `LoadError::new_raw` takes an `&Artichoke`, which `interp` already is. This extra reference gets us a `&&Artichoke` which is immediately dereferenced by the compiler to get us back the `&Artichoke` (this is called auto-deref).

There is no runtime impact, but can make the code trickier to read.

lopopolo:
```suggestion
    LoadError::new_raw(interp, message)
```

artichoke-backend/src/extn/core/kernel/require.rs
```diff
@@ -281,8 +281,8 @@ impl RelativePath {
     }
 }
 
-fn load_error(interp: &Artichoke, filename: &[u8]) -> Result<LoadError, Exception> {
-    let mut message = String::from("cannot load such file -- ");
-    string::format_unicode_debug_into(&mut message, filename)?;
-    Ok(LoadError::new(interp, message))
+fn load_error(interp: &Artichoke, filename: &[u8]) -> LoadError {
+    let mut message = String::from("cannot load such file -- ").into_bytes();
```
lopopolo, line 285: As a matter of style, I dislike creating a `String` and "unparsing" into a `Vec` when we can create the `Vec` directly.

lopopolo:
```suggestion
    let mut message = b"cannot load such file -- ".to_vec();
```

dnamsons:
Thank you for the suggestions on how to improve my code!

Issue #308, opened 2019-10-20
Implement reusable function for extracting begin and end from an mruby Range


lopopolo:
`extn::core` defines an `is_range` function in several places to extract the start and end from a `Range` `Value`.

https://github.com/artichoke/artichoke/blob/e8a9f922e7750c30cc89e21b79144ca5f3973572/artichoke-backend/src/extn/core/matchdata/element_reference.rs#L68-L93

`sys::mrb_range_beg_len` will raise in some circumstances, so `is_range` should be implemented with `mrb_protect`.

Each of the existing implementations is also slightly different. Some do conversion via `to_str` and `to_int` and others do not. It would be nice to consolidate them.

PR #635, opened 2020-04-19
Add a safe function for testing whether a value is a Range

lopopolo:
Element access APIs often accept `Range` inputs to return a subslice of values. `is_range` validators exist in `Array` and `MatchData` implementations.

These implementations build on the mruby API `mrb_range_beg_len`, which can potentially raise. Calls to this function must be wrapped and have been known to be unsafe.

This PR moves the `is_range` implementation to the `sys::protect` module and uses the traits to wrap calls to `mrb_range_beg_len` in `mrb_protect`.

This PR adds `Value::is_range` that wraps `sys::protect::is_range` similar to `Value::funcall`.

This PR removes the duplicated implementations and updates arg parsing to delegate to the `Value::is_range` function.

Specs continue to pass after this change.

Fixes GH-308.


lopopolo:
Related to GH-442.

PR #648, opened 2020-04-27
Add cargo-bloat action to track metircs

lopopolo:
https://github.com/orf/cargo-bloat-action/

> Analyse and track your Rust project binary size over time. This action runs on every pull request and gives you a breakdown of your total binary size, how much each crate contributes to that size and a list of changes to your dependency tree.


lopopolo:
This action does not work. Closing.

PR #651, opened 2020-04-28
Implement Kernel#p as a trampoline method similar to Kernel#puts

### LINKED ISSUE #440: Need to have `p` method
We have `puts`  and `print`, but we don't have  `p` method.

jonkgrimes:
Fixes #440 

I've taken a stab at implementing the `p` method in the `trampoline` of the `Kernel` module. Looking at the implementation of MRI it looks like this method is used to debug Ruby objects from C so I felt the the trampoline was a good place for it.

I also wasn't sure on how to get this under test. For the time being I added it to the `Kernel` module integration tests. I ran it against the Rubyspec tests but there were quite a few prerequisites required to run.

Let me know what you think. I'm a Rust beginner and new to working with interpreters in general, so I'm totally willing to re-work this in pure Ruby as well if the trampoline isn't a good place for this.

### DIFF
--- a/artichoke-backend/src/extn/core/kernel/kernel_test.rb
+++ b/artichoke-backend/src/extn/core/kernel/kernel_test.rb
@@ -4,6 +4,9 @@
 # https://ruby-doc.org/core-2.6.3/Kernel.html
 def spec
   throw_catch
+  kernel_p_no_args
+  kernel_p_one_arg
+  kernel_p_array_args
 
   true
 end
@@ -38,6 +41,35 @@ def throw_catch
     456
   end
   raise unless result == 123
+end
+
+class Foo
+  attr_accessor :a, :b
+  def initialize(a, b)
+    @a = a
+    @b = b
+  end
+end
+
+# https://ruby-doc.org/core-2.6.3/Kernel.html#method-i-p
+def kernel_p_no_args
+  result = p
+  raise if result
+end
+
+def kernel_p_one_arg
+  f = Foo.new(1, 2)
+  result = p(f)
+  raise unless result
+end
+
+def kernel_p_array_args
+  f = Foo.new(1, 2)
+  g = Foo.new(3, 4)
+  result = p(f,g)
+  raise unless result
+  raise unless result.is_a?(Array)
+  raise unless result.length == 2
 end
 
 spec if $PROGRAM_NAME == __FILE__
--- a/artichoke-backend/src/extn/core/kernel/kernel_test.rb
+++ b/artichoke-backend/src/extn/core/kernel/kernel_test.rb
@@ -44,30 +44,30 @@ def throw_catch
 end
 
 class Foo
-  attr_accessor :a, :b
-  def initialize(a, b)
-    @a = a
-    @b = b
+  attr_accessor :bar, :baz
+  def initialize(bar, baz)
+    @bar = bar
+    @baz = baz
   end
 end
 
 # https://ruby-doc.org/core-2.6.3/Kernel.html#method-i-p
 def kernel_p_no_args
   result = p
-  raise if result
+  raise unless result.nil?
 end
 
 def kernel_p_one_arg
   f = Foo.new(1, 2)
   result = p(f)
-  raise unless result
+  raise unless result.equal?(f)
 end
 
 def kernel_p_array_args
   f = Foo.new(1, 2)
   g = Foo.new(3, 4)
-  result = p(f,g)
-  raise unless result
+  result = p(f, g)
+  raise unless result == [f, g]
   raise unless result.is_a?(Array)
   raise unless result.length == 2
 end
--- a/artichoke-backend/src/extn/core/kernel/mruby.rs
+++ b/artichoke-backend/src/extn/core/kernel/mruby.rs
@@ -15,9 +15,9 @@ pub fn init(interp: &mut Artichoke) -> InitializeResult<()> {
             sys::mrb_args_rest(),
         )?
         .add_method("load", artichoke_kernel_load, sys::mrb_args_rest())?
+        .add_method("p", artichoke_kernel_p, sys::mrb_args_rest())?
         .add_method("print", artichoke_kernel_print, sys::mrb_args_rest())?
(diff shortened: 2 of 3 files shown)


artichoke-backend/src/extn/core/kernel/kernel_test.rb
```diff
@@ -4,6 +4,9 @@
 # https://ruby-doc.org/core-2.6.3/Kernel.html
 def spec
   throw_catch
+  kernel_p_no_args
```
lopopolo, line 7: this is a great place to add tests. Thank you 💪

artichoke-backend/src/extn/core/kernel/trampoline.rs
```diff
@@ -54,6 +54,20 @@ pub fn puts(interp: &mut Artichoke, args: Vec<Value>) -> Result<Value, Exception
     Ok(interp.convert(None::<Value>))
 }
 
+pub fn p(interp: &mut Artichoke, args: Vec<Value>) -> Result<Value, Exception> {
```
lopopolo, line 57: I merged a change earlier today to have `print` and `puts` take `IntoIterator` arguments so they can be used on iterators in addition to `Vec`s. Could you change the signature to something like this?

https://github.com/artichoke/artichoke/blob/3ba3ece1939f2dee96c9442a91cbf7a6f79acc1d/artichoke-backend/src/extn/core/kernel/trampoline.rs#L21-L23

This shouldn't require any changes to the `mruby` glue.

artichoke-backend/src/extn/core/kernel/trampoline.rs
```diff
@@ -54,6 +54,20 @@ pub fn puts(interp: &mut Artichoke, args: Vec<Value>) -> Result<Value, Exception
...
+    for value in &args {
+        let display = value.inspect(interp);
+        let mut borrow = interp.0.borrow_mut();
+        borrow.output.puts(display);
```
lopopolo, line 61: unfortunately the current implementation of `output` doesn't allow this to be tested easily. There is an in-memory `output` backend, but it requires a compile time flag to flip it, which we can't change in tests.

If you can validate the output is written to stdout in `airb` or with `artichoke -e 'p 1'`, that's good enough for me.

artichoke-backend/src/extn/core/kernel/trampoline.rs
```diff
@@ -54,6 +54,20 @@ pub fn puts(interp: &mut Artichoke, args: Vec<Value>) -> Result<Value, Exception
...
+    }
+
+    match args.len() {
```
lopopolo, line 64: this is clever, but also very readable. Thank you.

artichoke-backend/src/extn/core/kernel/mruby.rs
```diff
@@ -17,6 +17,7 @@ pub fn init(interp: &mut Artichoke) -> InitializeResult<()> {
         .add_method("load", artichoke_kernel_load, sys::mrb_args_rest())?
         .add_method("print", artichoke_kernel_print, sys::mrb_args_rest())?
         .add_method("puts", artichoke_kernel_puts, sys::mrb_args_rest())?
+        .add_method("p", artichoke_kernel_p, sys::mrb_args_rest())?
```
lopopolo, line 20: can this be alphabetized before `print` to match the Ruby docs:

https://ruby-doc.org/core-2.6.3/Kernel.html#method-i-p

artichoke-backend/src/extn/core/kernel/kernel_test.rb
```diff
@@ -4,6 +4,9 @@
 # https://ruby-doc.org/core-2.6.3/Kernel.html
```
lopopolo, line 4: if the Ruby lint job is failing, you can run

```
rake lint:rubocop
```

to run autofixups.

artichoke-backend/src/extn/core/kernel/kernel_test.rb
```diff
@@ -40,4 +43,33 @@ def throw_catch
...
+  def initialize(a, b)
+    @a = a
+    @b = b
+  end
+end
+
+# https://ruby-doc.org/core-2.6.3/Kernel.html#method-i-p
+def kernel_p_no_args
+  result = p
+  raise if result
```
lopopolo, line 57:
```suggestion
  raise unless result.nil?
```

artichoke-backend/src/extn/core/kernel/kernel_test.rb
```diff
@@ -40,4 +43,33 @@ def throw_catch
...
+end
+
+def kernel_p_one_arg
+  f = Foo.new(1, 2)
+  result = p(f)
+  raise unless result
```
lopopolo, line 63:
```suggestion
  raise unless result.equal?(f)
```

artichoke-backend/src/extn/core/kernel/kernel_test.rb
```diff
@@ -40,4 +43,33 @@ def throw_catch
...
+end
+
+def kernel_p_array_args
+  f = Foo.new(1, 2)
+  g = Foo.new(3, 4)
+  result = p(f,g)
+  raise unless result
```
lopopolo, line 70:
```suggestion
  raise unless result == [f, g]
```

lopopolo:
> Looking at the implementation of MRI it looks like this method is used to debug Ruby objects from C so I felt the the trampoline was a good place for it.

Wow, awesome! Thanks for digging that far. I never would have thought to look for that. 💎

artichoke-backend/src/extn/core/kernel/trampoline.rs
jonkgrimes: Is there a way to execute either `airb` or `artichoke` as part of the normal `cargo` test suite? I was looking into how `ripgrep` executed their tests but I wasn't sure if there was something pre-existing that I haven't been able to find. It's also pretty heavyweight so didn't want to introduce a lot boilerplate and complexity if not needed.

lopopolo: Oh I didn’t mean in the test suite. I meant one off manually for now.

PR #657, opened 2020-05-02
Add Io trait for puts and print capabilities on the interpreter

lopopolo:
This trait is similar to the `Warn` trait, except it allows writing to
stdout.

This change got tied up with introducing `prelude` modules in the core,
backend, and artchoke crates. Imports got reorganized in
artichoke-backend (again) which is why there is a lot of churn in this
commit.


lopopolo:
This PR supports GH-442.

PR #688, opened 2020-05-26
Update rust-onig and bindgen to latest

lopopolo:
This PR pulls in the latest rust-onig fork which syncs master as of
rust-onig/rust-onig@742dc25ab7750c21493b666b90bdee7656d0b221. This patch
level removed most of my patches except the one that adds bindgen
support for wasm32 targets.

The fork also updates the bindgen dep, which we propagate to
artichoke-backend as well.


lopopolo:
obsoletes GH-687

Issue #26, opened 2019-07-07
Add tests for regexp::opts::parse_pattern


lopopolo:
There is a gnarly state machine and String processing going on. Needs tests, which should help as we modify this function to be backed by a nom parser.

https://github.com/artichoke/artichoke/blob/b87fa37dc360b8969ba48abdeb8c17cab4468ceb/artichoke-backend/src/extn/core/regexp/opts.rs#L105-L200

PR #698, opened 2020-06-04
Make onig an optional dependency

lopopolo:
This PR adds a cargo feature to `artichoke` and `artichoke-backend` called `core-regexp-oniguruma`.

`core-regexp-oniguruma` is enabled by default. When the feature is enabled, `onig` dependency is built and the core `Regexp` implementation is a hybrid of the Rust `regex` crate and oniguruma.

Disabling this feature builds the core `Regexp` backend such that only the Rust `regex` crate powers it. Disabling this feature should make building for Wasm targets easier since Rust `regex` is pure Rust and the C `onig` dependency is excluded.

This PR opportunistically gives some love to the rest of the `Regexp` Rust implementation, refactoring to use more Rust idioms, iterators, and trait implementations; increase readability and performance; and add tests.

Fixes GH-489.
Fixes GH-26.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,6 +61,29 @@ jobs:
       - name: Test
         run: cargo test --workspace
 
+  build-features:
+    name: Build (non-default features)
+    runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: -D warnings
+      RUST_BACKTRACE: 1
+    steps:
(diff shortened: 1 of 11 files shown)


lopopolo:
This PR splits the `--no-default-features` and `--all-features` builds out of the OS matrix and executes them only on Linux. These builds tend to not expose platform specific bugs. The increase in parallelism should speed up the overall workflow.

PR #710, opened 2020-06-11
Remove Artichoke parameter from $exception::new for all Exceptions in Core

lopopolo:
`extn::core::exception` implements all `Exception` descendants in Ruby Core using macros because there are a bunch of them.

These macros define the inherent methods:

```rust
impl $excepton {
    pub fn new<T: Into<Cow<'static, str>>>(interp: &Artchoke, msg: T)` {}

    pub fn new_raw<T: Into<Cow<'static, [u8]>>>(interp: &Artchoke, msg: T)` {}
}
```

The interp parameters were present since the initial implementation of these structs. When `Artichoke` was clone-able, each exception struct owned a copy of the `Rc`-wrapped interpreter to resolve its classname. 042bc4b3484c9e89890b3f3114053f56141a67bb removed this struct member in favor of a `stringified` class name, but did not refactor the constructor or any call sites.

This PR changes the inherent impl to:

```rust
impl $exception {
    pub fn new() -> Self {
        Self(stringify!($exception).into())
    }
}
```

This matches the behavior of MRI when raising an `Exception` with no message or calling `Exception.new` with no message:

```console
[2.6.3] > raise RuntimeError
        4: from /usr/local/var/rbenv/versions/2.6.3/bin/irb:23:in `<main>'
        3: from /usr/local/var/rbenv/versions/2.6.3/bin/irb:23:in `load'
        2: from /usr/local/var/rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        1: from (irb):2
RuntimeError (RuntimeError)
[2.6.3] > RuntimeError.new.message
=> "RuntimeError"
```

Constructs with an explicit message have been reworked to `From` impls for owned, borrowed, and `Cow` types. This removes some of the ugliness in the `new_raw` constructor.

Removing the unused `interp` parameter from the error constructors allows culling of significant unnecessary proliferation of the interp parameter into `extn` backend implementations. This change removes most `interp` params from `Regexp` and `MatchData` backends, limiting the parameter to the trampolines. Some APIs, like `SecureRandom` now are `interp`-free which improves the ergonomics of invoking them from Rust.

### DIFF
--- a/artichoke-backend/src/block.rs
+++ b/artichoke-backend/src/block.rs
@@ -151,10 +151,7 @@ impl Block {
                     // and may result in a segfault.
                     //
                     // See: https://github.com/mruby/mruby/issues/4460
-                    Err(Exception::from(Fatal::new(
-                        arena.interp(),
-                        "Unreachable Ruby value",
-                    )))
+                    Err(Fatal::from("Unreachable Ruby value").into())
                 } else {
                     Ok(value)
                 }
--- a/artichoke-backend/src/convert/boxing.rs
+++ b/artichoke-backend/src/convert/boxing.rs
@@ -103,7 +103,7 @@ where
         if value.ruby_type() != Ruby::Data {
             let mut message = String::from("uninitialized ");
             message.push_str(Self::RUBY_TYPE);
-            return Err(Exception::from(TypeError::new(interp, message)));
+            return Err(TypeError::from(message).into());
         }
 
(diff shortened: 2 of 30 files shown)


lopopolo:
Filed https://github.com/BurntSushi/bstr/issues/59 as a result of this work.

lopopolo:
This PR improves Artichoke's adherence to the Rust API Guidelines which suggest a zero argument `new` method.

PR #720, opened 2020-06-20
Replace TypeRegistry with specialized class and module spec registries with expanded APIs

lopopolo:
- Add type parameter for customizing BuildHasher
- Expose capacity and build hasher constructors
- Expose memory management APIs - reserve, shrink to fit
- Add capacity, len, is_empty APIs
- Return old values when calling insert
- document that TypeRegistry is append only.
- add doc comments
- derive default, debug, partialeq, eq
- Add iterator over items in the registry
- Add `contains<K>()`


lopopolo:
This PR enforces that the registries are append only. Duplicate inserts panic. There are no other mutating APIs.

Issue #669, opened 2020-05-09
Return Result from ArenaIndex::new


lopopolo:
with the addition of `Artichoke::with_ffi_boundary`, this API is now fallible.

propagate an error instead of unwrapping.

use a custom error type that implements `RubyException`.

https://github.com/artichoke/artichoke/blob/e4eb05edb4a08c4463f4a206cb8fb3895ceda6b0/artichoke-backend/src/gc/arena.rs#L35-L39

lopopolo:
introduced in GH-661

lopopolo:
For a look at how to implement a custom `RubyException`, see `InterpreterAllocError` in `artichoke-backend/src/interpreter.rs`.

druzn3k:
I would love to try with this issue

druzn3k:
So, I modified the PR per request. Now `ArenaIndex::new` returns a `Result<ArenaIndex, IndexError>` instead of the `InterpreterError`.

PR #734, opened 2020-07-03
Return `Result` from ArenaIndex::new

druzn3k:
This commit try to solve issue #669. Now, if somehow fails to retrieve
the ArenaIndex in the GC, the default behavior is to return an error.
Down the path those exception are dealt with an `.expect(msg)`, as it
isn't clear at the time of writing, what should be the approach to a
failed retrieval.

Fixes GH-669.

### DIFF
--- a/artichoke-backend/src/gc/arena.rs
+++ b/artichoke-backend/src/gc/arena.rs
@@ -37,7 +37,7 @@ impl error::Error for IndexError {}
 
 impl RubyException for IndexError {
     fn message(&self) -> Cow<'_, [u8]> {
-        Cow::Borrowed(b"Failed to extract Artichoke Ruby interpreter from mrb_state")
+        Cow::Borrowed(b"Failed to create internal garbage collection savepoint")
     }
 
     fn name(&self) -> Cow<'_, str> {
--- a/artichoke-backend/src/gc/arena.rs
+++ b/artichoke-backend/src/gc/arena.rs
@@ -26,10 +26,7 @@ impl IndexError {
 
 impl fmt::Display for IndexError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "Failed to create internal garbage collection savepoint"
-        )
+        write!(f, "Failed to create internal garbage collection savepoint")
     }
 }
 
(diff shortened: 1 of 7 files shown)


artichoke-backend/src/gc/arena.rs
```diff
@@ -1,10 +1,85 @@
...
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+impl fmt::Display for IndexError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "Failed to extract Artichoke Ruby interpreter from mrb_state userdata"
```
lopopolo, line 31:
```suggestion
            "Failed to create internal garbage collection savepoint"
```

artichoke-backend/src/gc/arena.rs
```diff
@@ -1,10 +1,85 @@
...
+        )
+    }
+}
+
+impl error::Error for IndexError {}
+
+impl RubyException for IndexError {
+    fn message(&self) -> Cow<'_, [u8]> {
+        Cow::Borrowed(b"Failed to extract Artichoke Ruby interpreter from mrb_state")
```
lopopolo, line 40:
```suggestion
        Cow::Borrowed(b"Failed to create internal garbage collection savepoint")
```

artichoke-backend/src/gc/arena.rs
```diff
@@ -31,13 +106,13 @@ pub struct ArenaIndex<'a> {
...
-    pub fn new(interp: &'a mut Artichoke) -> Self {
-        let index = unsafe {
+    pub fn new(interp: &'a mut Artichoke) -> Result<Self, IndexError> {
+        unsafe {
             interp
                 .with_ffi_boundary(|mrb| sys::mrb_sys_gc_arena_save(mrb))
-                .unwrap_or_default()
-        };
-        Self { index, interp }
+                .and_then(move |index| Ok(Self { index, interp }))
```
lopopolo, line 113:
```suggestion
                .map(move |index| Self { index, interp })
```

artichoke-backend/src/gc/arena.rs
```diff
@@ -1,10 +1,85 @@
...
+        Self::default()
+    }
+}
+
+impl fmt::Display for IndexError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "Failed to create internal garbage collection savepoint"
+        )
```
lopopolo, line 32:
```suggestion
        write!(f, "Failed to create internal garbage collection savepoint")
```

PR #759, opened 2020-07-24
Add spinoso-array crate

lopopolo:
This is an experiment to move core data structures used to implement
Ruby Core and Standard Library out of `artichoke-backend`. The goal is
to have `artichoke-backend` be the VM and Ruby <-> Rust type
orchestrator via the `BoxUnboxVmValue` trait.

This idea has already been trialed with `intaglio` extracting the
`Symbol` backend.

The `spinoso` crate prefix comes from _Carciofo spinoso di Sardegna_,
the thorny artichoke of Sardinia. The idea is that the data structures
defined in the `spinoso` family of crates will form the backbone of Ruby
Core in Artichoke.

`spinoso-array` is a generalization of the `Array` and `InlineBuffer`
found in `artichoke_backend::extn::core::array`. It includes many more
slice and mutation methods which will allow defining more of Ruby
`Array` in native code at minimal maintenance cost.

`spinoso-array` includes two `Array` implementations:

- `Array`, based on `Vec` from Rust `alloc` and `std`. This backend
  contains no unsafe code.
- `SmallArray`, based on `SmallVec` from the Servo `smallvec` crate.
  This backend contains some unsafe code to implement a performant
  `set_slice` API for the case where a slice inserted into the interior
  of the array is less than the number of elements to drain.

`SmallArray` is gated behind the `small-array` feature. When the
`small-array` feature is not activated, the `smallvec` dependency is
dropped.

This crate has two `unsafe fn set_len` for manipulating Array length in
combination with raw pointers to support FFI usecases. The `SmallArray`
backend has two `unsafe` blocks to support a performant `O(n)` `set_slice`
API for an interior inserted slice that must drain `n` additional elements.
There are extensive _Safety_ comments.

`spinoso-array` has complete API documentation. Every method has an
_Examples_ section with doctests. `insert_slice` and `set_slice`, which
both have a large number of internal branches, have exhaustive
integration tests with full coverage.

Compared to the `InlineBuffer` implementation in `artichoke-backend`
that `spinoso-array` intends to replace, the API is more complete, there
are more tests, the tests do not depend on the Ruby VM, the code is
simpler, more well commented, and easier to follow, and more performant,
and there are docs.

A separate PR will replace `InlineBuffer` with one of the backends from
`spinoso-array`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -94,6 +94,14 @@ jobs:
       - name: Compile artichoke with all features
         run: cargo build --verbose --all-features
 
+      - name: Compile spinoso with no default features
+        run: cargo build --verbose --no-default-features
+        working-directory: "spinoso-array"
+
+      - name: Compile spinoso with all features
+        run: cargo build --verbose --all-features
+        working-directory: "spinoso-array"
+
   rust:
     name: Lint and format Rust
     runs-on: ubuntu-latest
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -643,6 +643,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "spinoso-array"
+version = "0.1.0"
+dependencies = [
+ "smallvec",
+]
+
+[[package]]
 name = "strsim"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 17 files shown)


lopopolo:
Realized the `unsafe` in `SmallArray::set_slice` can be replaced with two calls to `SmallVec::drain`. Patch incoming.

Issue #765, opened 2020-07-31
String#tr is broken implementation for character ranges and other regexp patterns


lopopolo:
## ruby/spec

```console
$ bundle exec rake spec
cargo run -q -- enforced-specs.toml
artichoke 0.1.0-pre.0 (2021-01-02 revision 3929) [x86_64-unknown-linux-gnu]

String#tr_s: NNNNXNSNNNNN
String#tr_s!: NNNX
String#tr: .XXXXX.XXS...
String#tr!: XXXX
String.try_convert: ........

Passed 11, skipped 2, not implemented 14, failed 14 specs.

An exception occurred during: Mock.verify_count
String#tr_s tries to convert from_str and to_str to strings using to_str
SpecExpectationNotMetError: Mock 'ab' expected to receive to_str(:any_args) exactly 1 times
but received it 0 times

(eval):653:in each
(eval):140:in each
(eval):193:in all?
(eval):653:in each
/src/lib/core/string/tr_s_spec.rb:104
(eval):653:in each
/src/lib/test/spec_runner:186:in run_specs

String#tr_s! raises a FrozenError if self is frozen
RuntimeError: frozen string

(eval):713:in tr_s!
/src/lib/core/string/tr_s_spec.rb:130:in protect
/src/lib/core/string/tr_s_spec.rb:130:in protect
(eval):193:in all?
(eval):653:in each
/src/lib/core/string/tr_s_spec.rb:134
(eval):653:in each
/src/lib/test/spec_runner:186:in run_specs

String#tr accepts c1-c2 notation to denote ranges of characters
SpecExpectationNotMetError: Expected "b-zzz" == "ifmmp"
to be truthy but was false

/src/lib/core/string/tr_spec.rb:13:in protect
(eval):193:in all?
(eval):653:in each
/src/lib/core/string/tr_spec.rb:99
(eval):653:in each
/src/lib/test/spec_runner:186:in run_specs

String#tr pads to_str with its last char if it is shorter than from_string
SpecExpectationNotMetError: Expected "A-HH." == "HE..."
to be truthy but was false

/src/lib/core/string/tr_spec.rb:21:in protect
(eval):193:in all?
(eval):653:in each
/src/lib/core/string/tr_spec.rb:99
(eval):653:in each
/src/lib/test/spec_runner:186:in run_specs

String#tr raises an ArgumentError a descending range in the replacement as containing just the start character
[...]

PR #770, opened 2020-08-07
Fix: String#tr implementation

dnamsons:


artichoke-backend/src/extn/core/string/string.rb
```diff
@@ -687,12 +687,11 @@ def to_str
   end
 
   def tr(from_str, to_str)
-    # TODO: Support character ranges c1-c2
```
dnamsons, line 690: It seems like character ranges have been implemented already, therefore i removed this TODO comment

spec-runner/vendor/spec/core/string/tr_spec.rb
```diff
@@ -19,6 +19,7 @@
   it "pads to_str with its last char if it is shorter than from_string" do
     "this".tr("this", "x").should == "xxxx"
     "hello".tr("a-z", "A-H.").should == "HE..."
+    "abcd".tr("a-z", "xxx").should == "xxxx"
```
lopopolo, line 22: ruby-specs come from upstream. I'd consider this source file immutable. Can you back out this change?

If you think we're missing a test, add an integration test for `String` like these for `Array`:

https://github.com/artichoke/artichoke/blob/b99e3c1163381cec8ab505e33c2201dbe67c7ade/artichoke-backend/src/extn/core/array/mod.rs#L418-L440

https://github.com/artichoke/artichoke/blob/b99e3c1163381cec8ab505e33c2201dbe67c7ade/artichoke-backend/src/extn/core/array/array_functional_test.rb

artichoke-backend/src/extn/core/string/string.rb
lopopolo: The below use of `Regexp` is a hack. This does method does not properly validate that ranges are ascending.

For example, this must error:

```
[2.6.6] > "hello".tr("a-y", "z-b")
        5: from /usr/local/var/rbenv/versions/2.6.6/bin/irb:23:in `<main>'
        4: from /usr/local/var/rbenv/versions/2.6.6/bin/irb:23:in `load'
        3: from /usr/local/var/rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        2: from (irb):6
        1: from (irb):6:in `tr'
ArgumentError (invalid range "z-b" in string transliteration)
```

Can you restore the comment?

lopopolo:
I tried this commit locally, and while it does improve spec compliance, there are still many ruby-spec failures. I'd like to take this PR with the changes I've requested.

Because `String#tr` still has some work to do, can you also edit the issue description to remove the "autoclose" annotation so GitHub does not close the attached ticket? Linking to it is fine.

lopopolo:
Please squash and force push your changes so we don't add any extraneous commits to the specs directory.

dnamsons:
@lopopolo Thanks for the review!

I've applied the changes you've asked

artichoke-backend/src/extn/core/string/string_test.rb
```diff
@@ -42,4 +43,8 @@ def string_unary_minus
   raise unless s.itself == 'abababa'
 end
 
+def string_tr
```
lopopolo, line 46: oh I didn't realize we had one of these for `String` 🚀

lopopolo:
link to https://github.com/artichoke/artichoke/issues/765

Issue #725, opened 2020-06-26
Implement Symbol#inspect


lopopolo:
Requires: https://github.com/artichoke/artichoke/issues/219

dnamsons:
Is is possible to work on this?

lopopolo:
Yes. Please, go for it 👍

dnamsons:
@lopopolo Thanks! 

I'm pretty sure on how to go about implementing this up to the point where specific encoding is required, e.g. 
`:''.inspect => ":\"\""`

Do you have any tips or ideas on how one might approach this?

lopopolo:
This function will handle unicode escapes for non UTF-8 characters.

https://github.com/artichoke/artichoke/blob/039f1a01c4f2a983c2b4ef9734fe856d7ff8757b/artichoke-backend/src/string.rs#L22-L61

I wouldn't worry about encoding. Artichoke treats all strings as maybe UTF-8.

lopopolo:
Oh I don't know if this was clear from the issue description, but this function is best implemented in Rust.

PR #772, opened 2020-08-07
Implement Symbol#inspect

dnamsons:
Progress on #725

Are tests necessary? if so, where should i create them?

### DIFF
--- a/artichoke-backend/src/extn/core/symbol/mod.rs
+++ b/artichoke-backend/src/extn/core/symbol/mod.rs
@@ -54,6 +54,19 @@ impl Symbol {
             &[]
         }
     }
+
+    #[must_use]
+    pub fn inspect(self, interp: &'_ mut Artichoke) -> Vec<u8> {
+        if let Ok(Some(bytes)) = interp.lookup_symbol(self.into()) {
+            // pattern length + ':'
+            let mut inspect = Vec::with_capacity(bytes.len() + 1);
+            inspect.push(b':');
+            inspect.extend(bytes);
+            inspect
+        } else {
+            Vec::new()
+        }
+    }
 }
 
 impl From<SymbolId> for Symbol {
--- a/artichoke-backend/src/extn/core/symbol/mruby.rs
+++ b/artichoke-backend/src/extn/core/symbol/mruby.rs
@@ -16,6 +16,7 @@ pub fn init(interp: &mut Artichoke) -> InitializeResult<()> {
         .add_method("empty?", artichoke_symbol_empty, sys::mrb_args_none())?
         .add_method("length", artichoke_symbol_length, sys::mrb_args_none())?
         .add_method("to_s", artichoke_symbol_to_s, sys::mrb_args_none())?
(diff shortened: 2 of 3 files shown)


dnamsons:
I did not handle the cases i mentioned in the issue(`:''.inspect => ":\"\""` and other symbols that can only be created by wrapping the symbol name in quotes), because it seems to depend on the implementation of symbol name creation & storage.

What I mean by this - in MRI 2.6.3, the symbol `:""` has the same output, but in this implementation it outputs only `:`.

artichoke-backend/src/extn/core/symbol/mod.rs
```diff
@@ -54,6 +54,19 @@ impl Symbol {
...
         }
     }
+
+    #[must_use]
+    pub fn inspect(self, interp: &'_ mut Artichoke) -> Vec<u8> {
+        if let Ok(Some(bytes)) = interp.lookup_symbol(self.into()) {
+            // pattern length + ':'
+            let mut inspect = Vec::with_capacity(bytes.len() + 1);
+            inspect.push(b':');
+            inspect.extend(bytes);
```
lopopolo, line 64: I think using `Vec::extend_from_slice` is more efficient here.

artichoke-backend/src/extn/core/symbol/mod.rs
```diff
@@ -54,6 +54,19 @@ impl Symbol {
             &[]
         }
     }
+
+    #[must_use]
+    pub fn inspect(self, interp: &'_ mut Artichoke) -> Vec<u8> {
```
lopopolo, line 59: Ideally this function would return a custom iterator instead of a `Vec`

artichoke-backend/src/extn/core/symbol/mod.rs
```diff
@@ -54,6 +54,19 @@ impl Symbol {
...
+        if let Ok(Some(bytes)) = interp.lookup_symbol(self.into()) {
+            // pattern length + ':'
+            let mut inspect = Vec::with_capacity(bytes.len() + 1);
+            inspect.push(b':');
```
lopopolo, line 63: To be correct, this function must parse the byte contents to determine if they should be wrapped in quotes or otherwise escaped. For example:

```
:""
: "$-www"
: "\xFF"
:"0123456789"
```

lopopolo: I had trouble formatting this on mobile. Please forgive the extra spaces.

artichoke-backend/src/extn/core/symbol/mruby.rs
```diff
@@ -102,3 +103,19 @@ unsafe extern "C" fn artichoke_symbol_to_s(
         Err(exception) => exception::raise(guard, exception),
     }
 }
+
+#[no_mangle]
+unsafe extern "C" fn artichoke_symbol_inspect(
```
lopopolo, line 108: This looks good thanks.

Issue #532, opened 2020-02-21
Declaratively manage labels for this repository


lopopolo:
https://github.com/lannonbr/issue-label-manager-action

lopopolo:
See also prior art here: https://github.com/artichoke/project-infrastructure/tree/master/github/modules

lopopolo:
When rolling this out, the `B-mruby`, `B-MRI`, and `A-compiler` labels were input incorrectly, causing the labels to be lost.

@lopopolo has gone through all 800-ish issues to put the labels back in place. This has enabled adding new labels (like `A-security`) as well as adding `B-mruby` in more places.

PR #775, opened 2020-08-08
Manage repository labels with a GitHub Action

lopopolo:
Declare repository labels with a JSON file.

This PR builds on the same techniques used in many other Artichoke organization
repositories.

This PR normalizes some of the label texts for labels already present in this
repo and fleshes out some categories, for example adding O-macOS and O-linux,
or adding B-Artichoke for eventual Rust VM backend plans.

Fixes GH-532.

### DIFF
--- a/.github/labels.json
+++ b/.github/labels.json
@@ -1,0 +1,277 @@
+[
+  {
+    "name": "A-build",
+    "color": "f7e101",
+    "description": "Area: CI build infrastructure."
+  },
+  {
+    "name": "A-build-target",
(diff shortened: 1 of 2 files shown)


lopopolo:
When rolling this out, the `B-mruby`, `B-MRI`, and `A-compiler` labels were input incorrectly, causing the labels to be lost.

@lopopolo has gone through all 800-ish issues to put the labels back in place. This has enabled adding new labels (like `A-security`) as well as adding `B-mruby` in more places.

PR #777, opened 2020-08-08
Re-generate missing A-compiler label

lopopolo:
Another copy paste error.


lopopolo:
See GH-776 and GH-775.

Issue #219, opened 2019-08-31
Add a parser to determine if a char sequence is a valid symbol literal


lopopolo:
mruby has a function called `symname_p` to determine if a symbol is a valid literal which it uses when generating the result for `Symbol#inspect`.

Although a `Symbol` can contain any byte content, only some literals can be printed without surrounding quotes like `:'a symbol with spaces is not a literal'`.

Use the [BNF for Ruby 1.4.6](https://cse.buffalo.edu/~regan/cse305/RubyBNF.pdf) as a starting point (it has changed, e.g. to add emoji identifiers):

```
SYMBOL : :FNAME | :VARNAME
FNAME : IDENTIFIER | .. | "|" | ˆ | & | <=> | == | === | =˜
        | > | >= | < | <= | + | - | * | / | % | **
        | << | >> | ˜ | +@ | -@ | [] | []=
VARNAME : GLOBAL | @IDENTIFIER | IDENTIFIER
GLOBAL : $IDENTIFIER | $any_char | $-any_char
IDENTIFIER : sequence in /[a-zA-Z_]{a-zA-Z0-9_}/.
```

lopopolo:
See https://github.com/mruby/mruby/issues/4678

PR #781, opened 2020-08-11
Add spinoso-symbol crate

lopopolo:
This continues the experiement started with `spinoso-array` (#759) to extract
data structures from `artichoke-backend` and generically implement Ruby
Core.

`spinoso-symbol` goes further than `spinoso-array` by implementing
`Symbol` APIs that depend on an Artichoke interpreter. It does this by
having an `artichoke` Cargo feature that pulls in `artichoke-core`. APIs
that require resolving the underlying `Symbol` byte content take a
generic parameter `T where T: Intern`.

`spinoso-symbol` implements an `AllSymbols` iterator that is `'static`
and implements a buggy version of `Symbol#inspect` with an iterator (#725).

`spinoso-symbol` resolves `Symbol` byte contents for ASCII and Unicode
casecmp functions using `focaccia`. `spinoso-symbol` also depends on
`bstr`.

`spinoso-symbol` re-exports some types from `focaccia`.

`spinoso-symbol` is agnostic over symbol table backend.

`spinoso-symbol` has a `std` feature that implements `std::error::Error`
on some types.

This PR adds most `Symbol` specs to `enforced-specs.yaml`.

PR #787, opened 2020-08-15
Implement ruby/spec-compatible Symbol#inspect

lopopolo:
Introduce a parser for Ruby identifiers which can be used to determine
whether `Symbol` byte content must be wrapped in quotes.

This commit iterates on the `Symbol#inspect` implementation introduced
in GH-772 and GH-781. Fix the buggy implementation of `Symbol#inspect`
by only wrapping in quotes if the byte contents are not a valid Ruby
identifier.

The new ident parser is gated behind a new `ident-parser` Cargo feature
in `spinoso-symbol` that exports an `IdentifierType` enum. `IdentifierType`
is constructed by parsing a `&str` or `&[u8]`. The logic for the ident
parser is modeled on the `rb_enc_symname_type` function from MRI.

This commit modifies the `Inspect` iterator to replace all `bool` fields
with two-element enums for readability. `Inspect` now properly escapes
quote characters for non-ident `Symbol`s.

The ident parser a property testing suite using the symbols present in
MRI after interpreter boot. The ident parser and `Inspect` iterator pass
all ruby/specs associated with `Symbol` parsing.

Fixes GH-725.
Fixes GH-219.


lopopolo:
I split the fixtures out of `ident.rs` to make the source file easier to navigate and review.

This bit of organization made it easy to add a functional test for `inspect.rs` as well.

PR #788, opened 2020-08-15
Add Symbol#to_proc to enforced ruby/specs

lopopolo:
This required moving from a stabby lambda definition to a more relaxed
lambda expression, checking for no receiver, and overriding Proc#parameters.

Enabled by #787.

### DIFF
--- a/artichoke-backend/src/extn/core/symbol/symbol.rb
+++ b/artichoke-backend/src/extn/core/symbol/symbol.rb
@@ -74,7 +74,18 @@ class Symbol
   end
 
   def to_proc
-    ->(obj, *args, &block) { obj.__send__(self, *args, &block) }
+    pr = lambda do |*args, &block|
+      raise ArgumentError, 'no receiver given' if args.empty?
+
+      obj, *rest = *args
+      obj.__send__(self, *rest, &block)
+    end
+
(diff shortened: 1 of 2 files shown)


artichoke-backend/src/extn/core/symbol/symbol.rb
```diff
@@ -74,7 +74,18 @@ def swapcase
...
   def to_proc
-    ->(obj, *args, &block) { obj.__send__(self, *args, &block) }
+    pr = lambda do |*args, &block|
+      raise ArgumentError, 'no receiver given' if args.empty?
+
+      obj, *rest = *args
+      obj.__send__(self, *rest, &block)
+    end
+
+    def pr.parameters
```
lopopolo, line 84: this is gross and probably means this constructor needs to move to Rust and be represented by an enum somehow.

spec-runner/enforced-specs.yaml
```diff
@@ -89,8 +89,7 @@ core:
       - succ
       - swapcase
       - symbol
-      # Requires working `Symbol#inspect` impl
-      # - to_proc
+      - to_proc
```
lopopolo, line 92: 🎉

Issue #528, opened 2020-02-20
Fix skipped tests on Windows


lopopolo:
GH-527 got the Windows test running instead of segfaulting as they do with CircleCI.

Several tests fail and have been skipped with:

```rust
#[cfg_atr(target_os = "windows", should_panic)]
```

These tests are marked with the ticket string `GH-528` (this ticket) to make finding them easier.

## Failing Tests

Failing tests have to do with path handling and the `\` separator on Windows.

- Backtraces and some assertions hardcode path strings to use `/`.
- Sources with the `__FILE__` magic constant raise `SyntaxError` on Windows.

### artichoke-backend

```
    extn::core::thread::tests::thread_abort_on_exception
    extn::stdlib::forwardable::tests::forwardable
    extn::stdlib::forwardable::tests::forwardable_another_example
    extn::stdlib::forwardable::tests::forwardable_def_instance_delegator
    fs::tests::absolutize_relative_path_dedot_parent_dir
```

### spec-runner

```
test mspec::tests::mspec_framework_loads ... FAILED
```

## How To Work on This

If you'd like to work on this ticket, you can get the tests to start failing by removing the `should_panic` attributes that decorate the test functions.

Then run `cargo test --workspace` to see the failures.

lopopolo:
Some of these were fixed in https://github.com/artichoke/artichoke/pull/936#issuecomment-736043207.

PR #831, opened 2020-09-05
Remove trait objects from the artichoke-backend virtual filesystem

lopopolo:
This PR removes the `Filesystem` trait from `artichoke_backend::fs`. This trait is a holdover from long ago when Artichoke depended on the `filesystem` crate (https://github.com/artichoke/ferrocarril/pull/1!).

This significant refactor of the `fs` module follows the same approach I've been using in the Spinoso crates: make multiple backends that are source compatible so they can be swapped out with Cargo feature flags.

This change introduces three `fs` backends:

- `Memory`: an in-memory backend that stores file contents in a `HashMap` and uses a `HashSet` to track `$LOADED_FEATURES`. `Memory` filesystems support storing extension hooks, which are used to initialize gems in Artichoke.
- `Native`: a host filesystem adapter that stores and reads file contents from local disk and uses a `HashSet` to track `$LOADED_FEATURES`.
- `Hybrid`: a composite `fs` backend that mounts a `Memory` instance at a prefix and forwards all operations to it that match this prefix. All other filesystem operations target the host filesystem.

This change also stores paths inside the `fs` module as `BString`. This means paths on Windows must be valid UTF-8 so they can be normalized (converting `\` to `/`). I expect this to fix a good many `should_panic` tests once this branch gets pushed to CI.

Fixes https://github.com/artichoke/artichoke/issues/528.

### DIFF
--- a/artichoke-backend/src/eval.rs
+++ b/artichoke-backend/src/eval.rs
@@ -189,8 +189,6 @@ mod tests {
     }
 
     #[test]
-    // TODO(GH-528): fix failing tests on Windows.
-    #[cfg_attr(target_os = "windows", should_panic)]
     fn file_magic_constant() {
         let mut interp = interpreter().unwrap();
         interp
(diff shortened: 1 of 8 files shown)


lopopolo:
rebased to latest trunk

PR #835, opened 2020-09-11
Add spinoso-math crate

lopopolo:
This commit extracts the `Math` implementation from artichoke-backend
`extn` into its own crate. This enables many quality improvements:

- The crate is `no_std` with an optional `std` feature for a`Error`
  impls.
- The `Math` implementation does not depend on `artichoke-backend` or
  `artichoke-core`.
- All functions in the math implementation are documented with example
  doctests.
- `DomainError` is a `Copy` type that is const constructed around a
  `&'static str`.
- Several math functions with high branching (e.g. `gamma`) were
  refactored to use `match` expressions in more places.

The `extn::core::math` module in `artichoke-backend` re-exports the core
`Math` handle, constants, and error types from `spinoso-math`. The `extn`
module wraps the underlying crate in a `trampoline` module, similar to
other spinoso extractions.

This commit introduces a new `CoerceToNumeric` trait in `artichoke-core`
for implementing the interpreter capabilities of coercing boxed Ruby
values into native numeric types. This trait is used in the `Math`
trampolines for unboxing values to `f64`. This change bumps the
`artichoke-core` version to 0.4.0.

This commit moves `Value::ruby_type` definition from
`artichoke_backend::value::Value` to `artichoke_core::value::Value`.

This extracts `Value::pretty_name` from `artichoke-backend` and
implements it generically as a free function in `artichoke-core`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -96,6 +96,22 @@ jobs:
         run: cargo build --verbose --all-features
         working-directory: "spinoso-array"
 
+      - name: Compile spinoso-array with no default features
+        run: cargo build --verbose --no-default-features
+        working-directory: "spinoso-env"
+
+      - name: Compile spinoso-array with all features
+        run: cargo build --verbose --all-features
+        working-directory: "spinoso-env"
+
+      - name: Compile spinoso-array with no default features
+        run: cargo build --verbose --no-default-features
+        working-directory: "spinoso-math"
+
+      - name: Compile spinoso-array with all features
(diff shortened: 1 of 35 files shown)


lopopolo:
force-pushed fixes:

- `rake fmt`
- Add missing `cfg`s to allow building artichoke without the `Math` module.
- Add no-default-features builds to CI for `spinoso-env` and `spinoso-math`.

lopopolo:
force pushed clippy and doc fixes.

PR #846, opened 2020-09-15
Improve test coverage of spinoso-random

lopopolo:
This commit expands the reproducibility tests in `spinoso-random`.

The tests now use a set of known good vectors generated from MRI Ruby
2.6.3. These vectors are generated in  Ruby and serialized to statics in
Rust test sources. See `spinoso-random/scripts/reproducibility.rb`.

Test vectors are bigger and more exhuastive. Vectors exist for u32, f64,
and u8 generation. All test vectors have 4096 samples.

The golden master tests have greater coverage. In addition to testing
the `Mt` Mersenne Twister implementation, the `Random` wrapper struct is
tested as well, since this is the public API exposed to the Ruby VM.

This commit also adds test to `spinoso_random::rand` using
`assert!(matches!(...))`. These tests assert error conditions, that the
correct types of `ArgumentError` are returned, and that the bounds of
the returned random numbers are as expected.

The `rand` tests also encode the differences between
`spinoso_random::rand` and `spinoso_securerandom::random_number`.

A basic correctness test for `spinoso_random::urandom` is added that
ensures the current platform source of randomness behaves appropriately.
This test supplements the existing doctest.

Followup to https://github.com/artichoke/artichoke/pull/836.


lopopolo:
the big LOC in this diff is due to the test vectors.

PR #860, opened 2020-10-01
Fix Time locals in airb REPL

lopopolo:
The `airb` REPL calls `#inspect` on each value returned by
`Artichoke::eval`. `Time` raises `NotImplementedError` in its `#inspect`
implementation. There is a nasty interaction here between the
`Time#inspect` implementation and the mruby locals tables that cause a
local bound to a `Time` to get reassigned to the `NotImplementedError`.

```console
artichoke 0.1.0-pre.0 (2020-10-01 revision 3626) [x86_64-apple-darwin]
[rustc 1.46.0 (04488afe3 2020-08-24)]
>>> t = Time.now
=>
>>> t.month
	1: from (airb):2
NoMethodError (undefined method 'month')
>>> t.class
=> NilClass
>>>
```

With this patch, the REPL behaves like this:

```console
artichoke 0.1.0-pre.0 (2020-10-01 revision 3627) [x86_64-apple-darwin]
[rustc 1.46.0 (04488afe3 2020-08-24)]
>>> t = Time.now
=> Time<Time#inspect is not implemented>
>>> t.month
=> 9
```

### DIFF
--- a/artichoke-backend/src/extn/core/time/trampoline.rs
+++ b/artichoke-backend/src/extn/core/time/trampoline.rs
@@ -155,9 +155,22 @@ pub fn asctime(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
 }
 
 pub fn to_string(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
-    let _ = interp;
     let _ = time;
-    Err(NotImplementedError::new().into())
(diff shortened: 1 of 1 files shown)


lopopolo:
I don't know what that _nasty interaction_ is and have chosen to work around it rather than understand the root cause.

PR #865, opened 2020-10-03
Remove pretty name

lopopolo:
This PR removes the `pretty_name` method from `artichoke-core`, which is likely the grossest generic code in Artichoke:

```rust
pub fn pretty_name<'a, V, T>(value: V, interp: &mut T) -> &'a str
where
    V: Copy + Value<Artichoke = T>,
    T: TryConvert<V, Option<bool>, Error = V::Error>
        + TryConvertMut<V, &'a str, Error = V::Error>
        + TryConvertMut<<<V as Value>::Value as Value>::Value, &'a str, Error = V::Error>,
    V::Value: Value<Artichoke = T>,
    <V::Value as Value>::Value: Value<Artichoke = T, Error = V::Error>,
{
    // ...
}
```

This function is replaced with a new `Debug` trait and a method for retrieving the inspect name for the class of the given value: `inspect_type_name_for_value`. The name is better too.

This PR bumps `artichoke-core` to version 0.6.0.

This PR fixes a long-standing bug with the stringification of `Ruby::Fixnum` so that it returns `Integer` and adds tests for generating the pretty name.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,7 +61,7 @@ dependencies = [
 
 [[package]]
 name = "artichoke-core"
-version = "0.5.0"
+version = "0.6.0"
 
 [[package]]
 name = "atty"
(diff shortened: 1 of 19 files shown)


lopopolo:
Rebase to pull in fix for https://github.com/artichoke/artichoke/pull/866

PR #866, opened 2020-10-03
Doc alias does not allow spaces with latest nightly

lopopolo:
Breakage caused by https://github.com/rust-lang/rust/pull/76811


lopopolo:
See also https://github.com/rust-lang/rust/issues/76705#issuecomment-703123847

PR #868, opened 2020-10-04
Simplify uses of foreach_name in Onig Regexp backend

lopopolo:
Unnest, use explicit for loops for readability.


lopopolo:
I manually testes the `specs` still pass by overriding the `Regexp` constructor to always use the `onig` backend.

PR #870, opened 2020-10-06
Folds and max value

lopopolo:
Replace explicit calls to Iterator::fold with Iterator::min

use `::MAX` associated constants instead of `::max_value` associated
function.

Use `cmp` instead of `partial_cmp` since we're comparing unsigned
integers which are total.

---
Use `::MAX` and `::MIN` instead of `::max_value()` and `::min_value()` for integer primitives


lopopolo:
latest nightly broke the doc build. fixed in https://github.com/artichoke/artichoke/pull/871

PR #875, opened 2020-10-09
Fix: Do not leak Random seed in fmt::Debug implementation in spinoso-random

lopopolo:
Previously, printing a `spinoso_random::Random` with the debug format specifier would output something like:

```
---- random::tests::fmt_debug_does_not_leak_seed stdout ----
thread 'random::tests::fmt_debug_does_not_leak_seed' panicked at 'Random { mt: Mt {}, seed: [123456, 0, 0, 0] }', spinoso-random/src/random/mod.rs:325:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    random::tests::fmt_debug_does_not_leak_seed
```

which leaks the internal seed.

It is typical for PRNGs to hide internal state in `fmt::Debug` implementations.

This commit alters the debug repr of `Random` to print `Random {}` and enforces this behavior with a test.

### DIFF
--- a/spinoso-random/src/random/mod.rs
+++ b/spinoso-random/src/random/mod.rs
@@ -1,3 +1,4 @@
+use core::fmt;
 use core::mem::size_of;
 
 use crate::{InitializeError, NewSeedError};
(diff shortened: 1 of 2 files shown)


lopopolo:
This latest commit adds similar tests to the internal `Mt` mersenne twister for completeness.

PR #896, opened 2020-11-07
Rubocop 1.2.0

lopopolo:
Bump RuboCop to 1.2.0

Test annotations in new version. See #894.


lopopolo:
It looks like annotations only get displayed if the sources that trigger the warnings are included in the PR https://gist.github.com/ybiquitous/72a1d7cf0eb9bbb9255d84d1ee904881

PR #912, opened 2020-11-21
Clean up win_flex_bison installation PowerShell in ci workflow

lopopolo:
- Use Invoke-WebRequest
- Extract directly to TEMP with a pipeline
- Factor artifact URL to an env variable for readability
- Collapse 'Check bison' build steps to one step that executes a bash script


lopopolo:
I've given up trying to get this to work. Existing code functions as is, it just is not very pretty

PR #972, opened 2020-12-23
Convert spec-runner configuration file from YAML to TOML

lopopolo:
This commit makes a breaking change to the spec-runner by changing the
configuration format from YAML to TOML. This commit bumps the version
number of the `spec-runner` to 0.3.0.

The YAML structure for declaring tagged specs is not the best. This
commit converts the structure from:

`Map<Vec<NamedSuite>>`

to:

`Map<Family, Map<Suite, Specs>>`

using TOML as the configuration format.

This commit removes `serde_yaml` and its dependencies and replaces it
with the `toml` crate, which has zero dependencies.

The config format could be translated to TOML and the model reused as
is, but the resulting TOML would have lots of e.g. `[[core]]` nested
table blocks which are a bit ugly. This new configuration format ensures
that each API suite is only declared once.

The `model` module is reimplemented for the new config format and the
model and spec setup code is refactored a bit to improve readability.


lopopolo:
This change in the data model also removes the janky `skip_all` magic spec. This behavior is now represented with `include = "none"`.

lopopolo:
The motivation of this change was to build all artichoke workspaces with a `-Z minimal-versions` lockfile.

`yaml-rust` currently does not build with its minimal version of `linked-hash-map`, see https://github.com/chyh1990/yaml-rust/pull/162#issuecomment-749748022. This prevents `serde_yaml` from building with `-Z minimal-versions`.

Additionally `linked-hash-map` has had security issues in the past, so if we can avoid depending on it, let's do it: https://github.com/RustSec/advisory-db/issues/298.

PR #981, opened 2020-12-27
Amortize double pointer deref on Artichoke State

lopopolo:
The inner Rust `State` is stored as a `Option<Box<State>>` property on the
`Artichoke` interpreter. When `artichoke-backend` implements interpreter
traits from `artichoke-core`, it is common to access the inner `State`
using `Option::as_ref` and `Option::as_mut` at the beginning of a method
to unwrap the `State` or error out. These APIs return `&Box<State>` and
`&mut Box<State>` (aka pointers to a pointer), which results in a double
indirection.

Replace all calls to unwrap the `State` with `Option::as_deref` and
`Option::as_deref_mut` to retrieve a `&State` and `&mut State`.
Implementations of `artichoke-core` traits often unwrap the `State` once
before performing several operations on it. This change results in
following the double pointer once via `Deref` to amortize its cost over
the entire function call.

This PR inspired filing https://github.com/rust-lang/rust-clippy/issues/6393.

### DIFF
--- a/artichoke-backend/src/class.rs
+++ b/artichoke-backend/src/class.rs
@@ -48,7 +48,7 @@ impl<'a> Builder<'a> {
         T: Any,
         U: Into<Cow<'static, str>>,
     {
-        let state = self.interp.state.as_ref().ok_or_else(InterpreterExtractError::new)?;
+        let state = self.interp.state.as_deref().ok_or_else(InterpreterExtractError::new)?;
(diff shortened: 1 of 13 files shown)


lopopolo:
This PR has been sitting on a branch for a while. https://github.com/artichoke/artichoke/pull/978 and https://github.com/artichoke/artichoke/pull/960 made it not fun to rebase.

PR #1007, opened 2021-01-06
Improve quality of `no_std` crates

lopopolo:
This PR unconditionally enables the `#![no_std]` crate-level attribute for no-std crates in this workspace, removing the `cfg_attr`. The `std` feature links to `std` using an `extern crate` declaration. All dependencies from `std` must always be imported.

This PR ensures all no-std crates build and test successfully with all combinations of feature flags. This involved adjusting some doctests to include `cfg` metas and restructuring test modules to be feature-aware.

### DIFF
--- a/artichoke-core/src/lib.rs
+++ b/artichoke-core/src/lib.rs
@@ -81,7 +81,7 @@
 //! [`Regexp`]: https://ruby-doc.org/core-2.6.3/Regexp.html#class-Regexp-label-Special+global+variables
 //! [convert]: crate::convert
 
-#![cfg_attr(not(feature = "std"), no_std)]
+#![no_std]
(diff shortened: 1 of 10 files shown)


lopopolo:
These failures would not have been detected either before or after the changes in https://github.com/artichoke/artichoke/pull/1004 since tests do not run and are not compiled.

Filed https://github.com/artichoke/artichoke/issues/1009 to catch these bugs in CI with `cargo check --all-targets`.

lopopolo:
> This PR unconditionally enables the `#![no_std]` crate-level attribute for no-std crates in this workspace, removing the `cfg_attr`.

This convention and approach is documented here: https://github.com/rust-lang/cargo/blob/58a961314437258065e23cb6316dfc121d96fb71/src/doc/src/reference/features-examples.md#no_std-support.

PR #1013, opened 2021-01-07
Add a default cargo-deny version to the audit CI job

lopopolo:
The `cargo-deny` version is injected into the `audit` GitHub Actions
workflow using an organization secret. This means that PRs from GitHub
users that are not part of the @artichoke organization do not have
access to the version and the subsequent fetch of the release assets
fails.

This PR adds a fallback version for `cargo-deny` which is set if
`$version` is empty.

This change allows PRs from external contributors to build all green.

### DIFF
--- a/.github/workflows/audit.yaml
+++ b/.github/workflows/audit.yaml
@@ -21,6 +21,7 @@ jobs:
       - name: Setup cargo-deny
         run: |
           version=${RELEASE_VERSION#"version="}
+          version=${version:-"0.8.5"}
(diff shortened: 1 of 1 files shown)


lopopolo:
This patch was also applied in all other repositories which have a `cargo-deny` workflow:

- https://github.com/artichoke/boba/pull/73
- https://github.com/artichoke/intaglio/pull/52
- https://github.com/artichoke/focaccia/pull/52
- https://github.com/artichoke/rand_mt/pull/60
- https://github.com/artichoke/strudel/pull/36
- https://github.com/artichoke/playground/pull/297

PR #1016, opened 2021-01-07
Add caching to GitHub Actions workflows

lopopolo:
This commit adds caching for the cargo registry, git cache, and target
directory of pre-built dependencies for the root workspace, fuzz
workspace, and spec-runner workspace.

These changes are made for the CI and rustdoc workflows.

These changes should speed up CI iterations significantly. In the
playground repo where these actions steps are cribbed from, this led for
a 50% reduction in build times.

PR #1017, opened 2021-01-08
Add Rust compiler artifact caching to GitHub Actions CI

lopopolo:
Add caching using the `Swatinem/rust-cache@v1` GitHub Action for all CI
jobs that use a stable compiler.

`rustdoc`, `fuzz`, and `leak-san` jobs use a nightly compiler which is
updated daily and will not get much reuse out of the `rust-cache`-generated
cache keys which include the rustc compiler hash on nightlies.

This GitHub Action was introduced successfully in the playground in
artichoke/playground#299.

This commit is a reimplementation of the failed PR #1016 that works
around actions/cache#403.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,6 +34,10 @@ jobs:
         with:
           ruby-version: ".ruby-version"
 
+      - uses: Swatinem/rust-cache@v1
+        with:
+          key: v3
+
       - name: Compile
         run: cargo build --workspace --verbose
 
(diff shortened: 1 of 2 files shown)


lopopolo:
Windows runners are failing to create a cache artifact: https://github.com/Swatinem/rust-cache/issues/7.

lopopolo:
This speeds up the longest build (`build` on Windows) from 8m33s to 6m26s.

All other builds come back in half the time.

PR #1022, opened 2021-01-08
Update SmallVec to 1.6.1

lopopolo:
Fixes a buffer overflow: servo/rust-smallvec#252.

`spinoso-array` is not impacted because it does not use the vulnerable
`SmallVec::insert_many` API.

RUSTSEC vulnerability notice: https://rustsec.org/advisories/RUSTSEC-2021-0003.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -531,9 +531,9 @@ checksum = "7fdf1b9db47230893d76faad238fd6097fd6d6a9245cd7a4d90dbd639536bbd2"
 
 [[package]]
 name = "smallvec"
-version = "1.6.0"
+version = "1.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1a55ca5f3b68e41c979bf8c46a6f1da892ca4db8f94023ce0bd32407573b1ac0"
+checksum = "fe0f37c9e8f3c5a4a66ad655a93c74daac4ad00c441533bf5c6e7990bb42604e"
 
 [[package]]
 name = "spinoso-array"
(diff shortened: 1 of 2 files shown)


lopopolo:
This is the report from `cargo-deny`, uncovered in [a GitHub Actions run](https://github.com/artichoke/artichoke/pull/1021/checks?check_run_id=1671033560) for #1021.

```
error[A001]: Buffer overflow in SmallVec::insert_many
   ┌─ /home/runner/work/artichoke/artichoke/Cargo.lock:60:1
   │
60 │ smallvec 1.6.0 registry+https://github.com/rust-lang/crates.io-index
   │ -------------------------------------------------------------------- security vulnerability detected
   │
   = ID: RUSTSEC-2021-0003
   = Advisory: https://rustsec.org/advisories/RUSTSEC-2021-0003
   = A bug in the `SmallVec::insert_many` method caused it to allocate a buffer that was smaller than needed.  It then wrote past the end of the buffer, causing a buffer overflow and memory corruption on the heap.
     
     This bug was only triggered if the iterator passed to `insert_many` yielded more items than the lower bound returned from its `size_hint` method.
      
     The flaw was corrected in smallvec 0.6.14 and 1.6.1, by ensuring that additional space is always reserved for each item inserted.  The fix also simplified the implementation of `insert_many` to use less unsafe code, so it is easier to verify its correctness.
     
     Thank you to Yechan Bae (@Qwaz) and the Rust group at Georgia Tech’s SSLab for finding and reporting this bug.
   = Announcement: https://github.com/servo/rust-smallvec/issues/252
   = Solution: Upgrade to >=0.6.14, <1.0.0 OR >=1.6.1
   = smallvec v1.6.0
     └── spinoso-array v0.4.0
         └── artichoke-backend v0.1.0
             └── artichoke v0.1.0-pre.0

 advisories FAILED: 1 errors, 0 warnings, 0 notes
           bans ok: 0 errors, 0 warnings, 1 notes
       licenses ok: 0 errors, 0 warnings, 85 notes
        sources ok: 0 errors, 0 warnings, 0 notes
Error: Process completed with exit code 1.
```

PR #1032, opened 2021-01-11
Fix interpreter memory leaks in REPL and Ruby frontends

lopopolo:
Neither `airb` nor `artichoke` binary frontends closed the `Artichoke`
interpreter, which leaks the interpreter and its heap. This does not
have many consequences since the program is exiting anyway, but it does
prevent destructors for Ruby values from being called.

This fix restructures the `ruby` and `repl` modules so they create an
interpreter at the top level and pass a `&mut Artichoke` to the
entrypoint implementation. This allows all short circuit `Err` returns
to be caught so the interpreter can be closed before passing the
`Result` back to `main`.

### DIFF
--- a/src/bin/artichoke.rs
+++ b/src/bin/artichoke.rs
@@ -90,7 +90,7 @@ fn main() {
         .with_programfile(matches.value_of_os("programfile").map(PathBuf::from));
 
     let mut stderr = StandardStream::stderr(ColorChoice::Auto);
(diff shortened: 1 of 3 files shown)


lopopolo:
This wasn't caught by LeakSanitizer in https://github.com/artichoke/artichoke/pull/984 because the `run` functions do not have tests :(

Issue #595, opened 2020-03-24
Investigate adding UI tests for artichoke binary frontend


lopopolo:
https://github.com/artichoke/artichoke/pull/594 added backtrace formatting to `artichoke` binary.

It is unclear how to test such a behavior change in Artichoke now.

[`rustc` has a suite of UI tests](https://github.com/rust-lang/rust/tree/master/src/test/ui) that test the output of the compiler in various circumstances.

In addition to testing that stacktraces get formatted correctly, we might want to test that `puts` uses `#to_s` and `pp` uses `#inspect`. Or that ANSI escapes make it to the output properly.

phansch:
Maybe [`insta`](https://github.com/mitsuhiko/insta) is a good fit for this?

lopopolo:
`insta` looks like it will fit the bill.

The tricky part is reworking the Artichoke output state to allow overrides without using cargo features. Otherwise, a unit test would have no way to capture `stdout` and `stderr`.

Capturing `stdout` is required for testing things like `puts`, `print`, and `p`.
Capturing `stdout` is required for `Warning::warn`, `Kernel::warn`, and built in warnings from core types like `Regexp` which are emitted from Rust.

lopopolo:
The regression addressed in https://github.com/artichoke/artichoke/pull/976 might have been prevented if there were UI tests for `airb`. `airb` is a readline shell which I believe requires a tty.

PR #1039, opened 2021-01-12
Add integration tests for ruby frontend

lopopolo:
Add tests that excersise the various `Args` configuration pathways.
These tests do not test the output of the frontend, see #595.

These tests ensure that LeakSanitizer tests the `ruby` frontend.


lopopolo:
6df1215e145ee33fa80fb4c651e369be4454080a should fail LeakSanitizer tests because it does not include #1032.

lopopolo:
GitHub actions is creating merge commits to current trunk which prevent the LeakSan test from running as I intended. https://github.com/actions/checkout/issues/27

I can verify locally that LeakSan does fail on 6df1215.

PR #1040, opened 2021-01-13
Make artichoke-backend tests pass with `--no-default-features`

silathdiir:
Closes #1008

PR #1041, opened 2021-01-13
CMath stdlib package depends on core Math module

lopopolo:
`CMath` includes `Math`, so reflect this dependency in the cargo features of `artichoke-backend`.

https://github.com/artichoke/artichoke/blob/44b60ed9a8f3922cfac6c8e07ecc343e3091bc3e/artichoke-backend/src/extn/stdlib/cmath/vendor/cmath.rb#L23-L28


lopopolo:
This dep was made clear to me in https://github.com/artichoke/artichoke/pull/1040.

Issue #184, opened 2019-08-20
Implement String core class


lopopolo:
Implementing `Encoding` in GH-183 is Artichoke's first significant deviation from the mruby implementation of Ruby Core classes.

To add support for `Encoding` to `String`, we need to implement `String` ourselves.

This issue only concerns implementing the `String` struct representation and does not cover exporting this type to mruby.

Ruby `String`s can be modeled with the following "backends":

## Rust `String`

For Ruby `String`s with UTF-8 encoding that _are_ valid UTF-8, use a Rust `String`. Character-oriented operations can use the `chars` iterator. Byte-oriented methods can use `String::as_bytes`. This representation can also be used for ASCII-encoded Ruby `String`s.

This implementation should lazily cache `String#length` since `s.chars().count()` is `O(n)`.

## `BString`

For Ruby `String`s with UTF-8 encoding that _are not_ valid UTF-8, use a [`BString`](https://docs.rs/bstr). Ruby allows for conventionally UTF-8 `String`s that contain invalid byte sequences. Character-oriented methods try to treat the `String` as UTF-8. Printing such a string prints the valid UTF-8 bytes as UTF-8.

## `Vec<u8>` + `Encoding`

For all other encodings, including `Encoding::BINARY`, store raw bytes and an encoder.

PR #1045, opened 2021-01-14
Add the beginnings of the spinoso-string crate

lopopolo:
This commit adds a WIP spinoso-string crate, which will be used to
implement the `String` Ruby Core class in Artichoke.

Spinoso `String` is modeled as a `Vec<u8>` with an associated
`Encoding`. The `Encoding` is conventional only, but does influence
several read-oriented APIs, like `String#ord` and `String#chr` as well
as new constructors like `String#center`.

Because a Spinoso `String` is meant to be packed into an mruby `RArray`,
it can be converted to raw pointer. `Encoding` can be serialize to a
`u8` bit flag.

Spinoso `String` intentionally only supports three Ruby encodings:

- UTF-8
- ASCII
- Binary / ASCII-8BIT

For UTF-8 support, `spinoso-string` makes use of `bstr`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -594,6 +594,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "spinoso-string"
+version = "0.1.0"
+dependencies = [
+ "bstr",
+ "focaccia",
+ "quickcheck",
+]
+
+[[package]]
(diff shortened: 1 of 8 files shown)


lopopolo:

PR #1047, opened 2021-01-14
Fix String::make_capitalized for invalid UTF-8 byte sequences

### LINKED ISSUE #1046: String::make_capitalized removes invalid UTF-8 byte sequences and the Unicode replacement character from conventionally UTF-8 strings
Introduced in #1045.

Two branches in the `String::make_capitalized` implementation cause both invalid UTF-8 byte sequences and the Unicode replacement character to be stripped from the capitalized string.

https://github.com/artichoke/artichoke/blob/7b867a30a442cde8c3ab8fd0b585432f93f7fa1a/spinoso-string/src/lib.rs#L1326
https://github.com/artichoke/artichoke/blob/7b867a30a442cde8c3ab8fd0b585432f93f7fa1a/spinoso-string/src/lib.rs#L1340-L1346

Despite the comments being correct that such sequences cannot be capitalized, the raw bytes must be preserved in the capitalized `String`.

Refactor the UTF-8 branch of `make_capitalized` to use `bstr::decode_utf8` and a loop.

lopopolo:
For conventionally UTF-8 strings, `String::make_capitalized` stripped
invalid UTF-8 bytes and the Unicode replacement character from the
string. This commit fixes this bug and adds tests for these cases, as
well as the happy path of ASCII and UTF-8 strings.

Fixes #1046.

This commit also removes a bounds check from the binary/ASCII path in
`make_capitalized`.

### DIFF
--- a/spinoso-string/src/lib.rs
+++ b/spinoso-string/src/lib.rs
@@ -32,7 +32,6 @@ extern crate std;
 
 use alloc::boxed::Box;
 use alloc::vec::{self, Vec};
-use core::char::REPLACEMENT_CHARACTER;
 use core::cmp::Ordering;
 use core::convert::TryFrom;
 use core::fmt::{self, Write};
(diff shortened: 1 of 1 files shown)


lopopolo:

lopopolo:
Filed a ticket to maybe move this logic into `bstr`: https://github.com/BurntSushi/bstr/issues/81.

PR #1048, opened 2021-01-14
Implement fmt::Debug for Spinoso String with bstr to render conventionally UTF-8 strings

lopopolo:
This also renders binary content with hex escapes.


lopopolo:

PR #1049, opened 2021-01-14
Add tests for UTF-8 char len and make_capitalize with pathological strings

lopopolo:
Sourced from https://github.com/minimaxir/big-list-of-naughty-strings/blob/894882e7/blns.txt.


lopopolo:

lopopolo:
Filed a ticket to maybe move this logic into `bstr`: https://github.com/BurntSushi/bstr/issues/81.

PR #1050, opened 2021-01-16
Add `--all-targets` flag to `cargo check` calls in CI

### LINKED ISSUE #1009: Add `--all-targets` flag to `cargo check` calls in CI
https://github.com/artichoke/artichoke/pull/1004 transitioned many CI build steps to use `cargo check`.

`cargo check` supports an `--all-targets` flag (see https://github.com/rust-lang/cargo/pull/4592), e.g. `cargo check --all-targets`, which will also run `cargo check` on tests.

Running `cargo check` on tests without default features would have detected the errors uncovered in #1007 and #1008.

silathdiir:
Closes #1009 

Should we also add `--profile=test` to `cargo check` calls?
It seems that it could check unit tests according to this [Cargo doc](https://doc.rust-lang.org/cargo/commands/cargo-check.html#examples).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,13 +80,13 @@ jobs:
           args: --workspace --all-features --all-targets
 
       - name: Check artichoke with locked Cargo.lock
-        run: cargo check --locked --all-targets
+        run: cargo check --locked --all-targets --profile=test
 
       - name: Check artichoke with no default features
-        run: cargo check --verbose --no-default-features --all-targets
+        run: cargo check --verbose --no-default-features --all-targets --profile=test
 
       - name: Check artichoke with all features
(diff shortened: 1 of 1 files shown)


lopopolo (approved):

lopopolo:
> Should we also add `--profile=test` to `cargo check` calls?
> It seems that it could check unit tests according to this [Cargo doc](https://doc.rust-lang.org/cargo/commands/cargo-check.html#examples).

OO good call thank you for for checking the docs. @silathdiir would you mind revising this PR to ensure we check unit tests?

silathdiir:
Hi @lopopolo , I append `--profile=test` to `cargo check` in CI. Please help review if you have time. Thanks.

lopopolo (approved):
🚀

PR #1051, opened 2021-01-18
Implement String#valid_encoding?

lopopolo:
Progress on #184.

This implementation uses SIMD-optimized routines from `bstr` and only does validation.


lopopolo:

Issue #310, opened 2019-10-20
Implement implicit conversions to Array, Integer, and String on Value


lopopolo:
This is a fairly common pattern when preparing args for a function (and this function gets it wrong by not calling `to_str`):

```rust
pub fn extract(interp: &Artichoke, at: Value) -> Result<Self, Box<dyn RubyException>> {
    let name = at.pretty_name();
    if let Ok(index) = at.clone().try_into::<Int>() {
        Ok(Self::Index(index))
    } else if let Ok(name) = at.clone().try_into::<&str>() {
        Ok(Self::Name(name))
    } else if let Ok(index) = at.funcall::<Int>("to_int", &[], None) {
        Ok(Self::Index(index))
    } else {
        Err(Box::new(TypeError::new(
            interp,
            format!("no implicit conversion of {} into Integer", name),
        )))
    }
}
```

Implement functions on `Value` for implicitly converting to `Array`, `Integer`, and `String`. The steps:

- Return `self` if `self` already is the type we want to convert to.
- Call `to_xxx`, e.g. `to_ary` or `to_int` or `to_str`, if it exists:
  - If method raises, propagate exception
  - If method returns the type we are trying to convert to, return it
  - Else raise `TypeError`
- Else raise `TypeError`

lopopolo:
GH-384 added implicit conversions for `Integer` but did not fully migrate the codebase.

lopopolo:
GH-450 fully migrated the codebase for `Value::implicitly_convert_to_int`.

lopopolo:
GH-520 adds `Value::implicitly_convert_to_string` and `Value::implicitly_convert_to_nilable_string`.

lopopolo:
`Kernel::puts` needs to be modified to implicitly convert to `Array` while looping over values.

PR #1065, opened 2021-01-20
Improve spec compliance of implicit conversions

lopopolo:
This PR moves the implicit conversion functions out of the `Value` impl and into their own module.

The implicit conversions are made to be more compatible with MRI, such as propagating exceptions from `to_int`/`to_str`.

The converters are well commented with REPL snippets from MRI corresponding to each branch. Each branch is tested via doctests. Each implicit conversion function is documented.

Implicit conversions to `String` are unsafe due to concerns about lifetime extension.

Implicit conversions to `String` no longer coerce `Symbol`s to `&[u8]`.

See #310.

### DIFF
--- a/artichoke-backend/src/convert.rs
+++ b/artichoke-backend/src/convert.rs
@@ -18,10 +18,12 @@ mod bytes;
 mod fixnum;
 mod float;
 mod hash;
+mod implicit;
 mod nilable;
 mod string;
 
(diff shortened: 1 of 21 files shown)


lopopolo:
Some `Regexp` specs are broken by not implicitly converting `Symbol`s anymore.

Issue #142, opened 2019-08-06
Split out spec-compliance tests to a separate repo


lopopolo:
Split out spec-compliance tests to a separate repo similar to [rust-toolstate](https://rust-lang-nursery.github.io/rust-toolstate/).

Trigger rebuilds of this repo on every commit and track the stability of spec complance by stating when state has changed and whether the specs pass or not.

Track number of failures, successes, skipped, and not implemented for each spec group.

This will speed up build times for the Rust build of Artichoke.

Issue #101, opened 2019-08-04
Generate metrics on spec-compliance during build and output to project website


lopopolo:
https://news.ycombinator.com/item?id=20606476

lopopolo:
This is unlikely to get done. Closing as WONTFIX.

eregon:
Here is what I used to estimate the number of specs Artichoke passes:
```
$ git clone https://github.com/artichoke/artichoke.git
$ cd artichoke
$ dnf install cargo clang llvm-devel

$ cargo run -q -p spec-runner -- spec-runner/enforced-specs.yaml
Passed 1694, skipped 214, not implemented 0, failed 0 specs.
```
I split `enforced-specs.yaml` into core and library:

```
$ cargo run -q -p spec-runner -- spec-runner/core-specs.yaml
Passed 1367, skipped 202, not implemented 0, failed 0 specs.

$ cargo run -q -p spec-runner -- spec-runner/library-specs.yaml
Passed 327, skipped 12, not implemented 0, failed 0 specs.
```

And then we can compare to how many specs CRuby passes by running the specs on CRuby.
That's the approach I used for https://eregon.me/blog/2020/06/27/ruby-spec-compatibility-report.html

What would be useful is if the CI ran the specs. I don't see the specs run in e.g. https://github.com/artichoke/artichoke/runs/962372127?check_suite_focus=true

Separating core & library into different files would be more convenient, but that's easy to do manually too.

lopopolo:
Hey @eregon thanks for sharing. I felt a bit of pride at being included in your roundup.

This approach likely significantly underestimates the number of `it` blocks Artichoke successfully passes.

`enforced-specs.yaml` corresponds to entire spec files that Artichoke passes. This is the granularity of the test harness.

Running the specs on CI is not something I'd like to do in this repository. I'd prefer to do something like `rust-toolstate` where spec pass rates are tested after every merge or daily externally -- much like how nightly and Docker builds are generated in external repos. This would enable building a custom crate and spec harness with more exhaustive testing and output more suitable for building a dashboard (JSON).

This seems to me like a lot of work that I am not interested in taking on right now.

This work is tracked in GH-142.

eregon:
> This approach likely significantly underestimates the number of `it` blocks Artichoke successfully passes.

Do you mean Artichoke might pass spec examples that are not in the `enforced-specs.yaml`, or something else?
[...]

PR #1088, opened 2021-02-14
Add a spec-runner configuration that tests all of Ruby Core

lopopolo:
It runs to completion!

```
Passed 4500, skipped 1097, not implemented 774, failed 5188 specs.
```

Gist with the full error log: https://gist.github.com/lopopolo/d7f758a4482a64ede6bb8a77918baba6.

There are still a few segfaults in there, but as configured, the specs reliably execute to completion.

This change adds a self-referential Hash check to `Hash#inspect` to prevent a stack overflow.

This change modifes `mrb_convert_type` to prevent a recursive attempt to raise a `TypeError` when attempting to convert a type to `String` that doesn't implement `#to_s`. This was fixed upstream in https://github.com/mruby/mruby/commit/01a8d8498fc3c1b107b303661d06b858858fce26.

The `all-core-specs.toml` configuration enables all top-level specs in `spec-runner/vendor/spec/core` except:

- `IO` (attempts to instantiate a binary `Regexp` that aborts the spec-runner)
- `Kernel` (infinite loop somewhere in there)
- `Mutex`
- `ObjectSpace`
- `Process`
- `Thread`
- `ThreadGroup`
- `TracePoint`

This config also copies the permitted spec list from `enforced-specs.toml` for `Array`. `Array#initialize` specs have an exception that crashes the spec-runner.

### DIFF
--- a/artichoke-backend/src/extn/core/hash/hash.rb
+++ b/artichoke-backend/src/extn/core/hash/hash.rb
@@ -1,5 +1,45 @@
 # frozen_string_literal: true
 
+module Artichoke
+  class Hash
+    # rubocop:disable Lint/HashCompareByIdentity
+    def self.trace(obj, seen)
+      id = obj.object_id
+      return true if seen[id]
+
+      seen[obj.object_id] = true
+
+      case obj
+      when Hash
+        obj.each do |key, value|
+          return true if trace(key, seen)
+          return true if trace(value, seen)
+        end
+        obj.instance_variables.each do |iv|
+          iv = obj.instance_variable_get(iv)
+          return true if trace(iv, seen)
+        end
+      when Array
+        obj.each do |elem|
+          return true if trace(elem, seen)
+        end
+        obj.instance_variables.each do |iv|
+          iv = obj.instance_variable_get(iv)
+          return true if trace(iv, seen)
+        end
+      when Object
+        obj.instance_variables.each do |iv|
+          iv = obj.instance_variable_get(iv)
+          return true if trace(iv, seen)
+        end
+      end
+
+      false
+    end
+    # rubocop:enable Lint/HashCompareByIdentity
+  end
+end
+
 class Hash
   include Enumerable
 
@@ -193,25 +233,30 @@ class Hash
   end
 
   def inspect
-    return '{}' if size.zero?
+    return '{}' if empty?
 
-    items = keys.map do |key|
-      val = self[key]
-      key =
-        if object_id == key.object_id
-          '{...}'
-        else
-          key.inspect
-        end
-      val =
-        if object_id == val.object_id
-          '{...}'
-        else
-          val.inspect
-        end
-      "#{key}=>#{val}"
+    id = object_id
+    pairs = []
+    seen = { id => true }
+
+    each do |key, value|
+      if ::Artichoke::Hash.trace(key, seen)
+        key = '{...}'
+      else
+        key = key.inspect
+        key = key.to_s unless key.is_a?(String)
+      end
+      if ::Artichoke::Hash.trace(value, seen)
+        value = '{...}'
+      else
+        value = key.inspect
+        value = key.to_s unless key.is_a?(String)
+      end
+
+      pairs << "#{key}=>#{value}"
     end
-    "{#{items.join(', ')}}"
+
+    "{#{pairs.join(', ')}}"
   end
 
   def invert
--- a/artichoke-backend/vendor/mruby/src/object.c
+++ b/artichoke-backend/vendor/mruby/src/object.c
@@ -338,7 +338,11 @@ mrb_convert_type(mrb_state *mrb, mrb_value val, enum mrb_vtype type, const char 
   if (mrb_type(val) == type) return val;
   v = convert_type(mrb, val, tname, method, TRUE);
   if (mrb_type(v) != type) {
-    mrb_raisef(mrb, E_TYPE_ERROR, "%v cannot be converted to %s by #%s", val, tname, method);
+    if (type == MRB_TT_STRING) {
+      mrb_raisef(mrb, E_TYPE_ERROR, "value cannot be converted to %s by #%s", tname, method);
+    } else {
+      mrb_raisef(mrb, E_TYPE_ERROR, "%v cannot be converted to %s by #%s", val, tname, method);
+    }
   }
   return v;
 }
--- a/spec-runner/all-core-specs.toml
+++ b/spec-runner/all-core-specs.toml
@@ -1,0 +1,260 @@
+# This config file lists the ruby/specs that should pass for Artichoke Ruby.
+#
+# Valid values for `include` are:
+#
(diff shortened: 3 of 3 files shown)


lopopolo:
Hi @eregon 👋 

This PR adds a configuration profile for Artichoke's spec runner that tests a wider swath or Ruby Core APIs.

I was curious if you'd be willing to add a note to https://eregon.me/blog/2020/06/27/ruby-spec-compatibility-report.html that points folks at this PR for a significant update.

eregon:
In the output above, does it mean 4500 spec examples (`it`) ran successfully, and then skipped/not implemented/failed are all added on top, so the sum of all of those is how many specs were attempted (all of core/ minus exclusions)?
There were 20841 core specs when I did the report, so I guess the exclusions above account for about half of the core specs.
A nice gain indeed if Artichoke now passes around 22% of core specs.

It'd be great to have this run in CI if possible, and ideally separate for language/core/library specs.

lopopolo:
> In the output above, does it mean 4500 spec examples (`it`) ran successfully, and then skipped/not implemented/failed are all added on top, so the sum of all of those is how many specs were attempted (all of core/ minus exclusions)?

@eregon yes that's correct.

> There were 20841 core specs when I did the report

Artichoke vendors the `spec` repo as of https://github.com/artichoke/spec/commit/eebd7aba44853c15a5d6c6535f511f9cd247c30c which might change the number of specs as well.

> It'd be great to have this run in CI if possible, and ideally separate for language/core/library specs.

Yes this is something I'd like to do, but haven't gotten around to it. Linking https://github.com/artichoke/artichoke/issues/101 and https://github.com/artichoke/artichoke/issues/142.

lopopolo:
Tweaking the config a bit, I get these numbers

```
Passed 4669, skipped 1125, not implemented 873, failed 5335 specs.
```

but also a reliable segfault on interpreter teardown.

<details>
<summary><code>all-core-specs.toml</code></summary>

```toml
# This config file lists the ruby/specs that should pass for Artichoke Ruby.
#
# Valid values for `include` are:
#
# - all - run all specs. When `include` is set to `all`, the optional `skip`
#   field may list specs to skip.
# - none - run no specs, equivalent to the section not being present in this
#   config file.
# - `set` - run an enumerated set of specs. When `include` is set to `set`, the
#   set of specs must be listed in the required `specs` field, which is a list
#   of strings.

## Ruby Core

[specs.core.argf]
include = "all"

[specs.core.array]
include = "all"
skip = [
  "initialize",
  "new",
  "pack",
]

[specs.core.basicobject]
include = "all"

[specs.core.binding]
include = "all"

[specs.core.builtin_constants]
include = "all"

[specs.core.class]
include = "all"

[specs.core.comparable]
include = "all"

[specs.core.complex]
include = "all"

[specs.core.dir]
include = "all"

[specs.core.encoding]
include = "all"

[specs.core.enumerable]
include = "all"

[specs.core.enumerator]
include = "all"

[specs.core.env]
include = "all"

[specs.core.exception]
include = "all"

[specs.core.false]
include = "all"

[specs.core.fiber]
include = "all"

[specs.core.file]
include = "all"

[specs.core.filetest]
include = "all"

[specs.core.float]
include = "all"

[specs.core.gc]
include = "all"

[specs.core.hash]
include = "all"

[specs.core.integer]
include = "all"

[specs.core.io]
include = "none"

[specs.core.kernel]
include = "none"

[specs.core.main]
include = "all"

[specs.core.marshal]
include = "all"

[specs.core.matchdata]
include = "all"

[specs.core.math]
include = "all"

[specs.core.method]
include = "all"

[specs.core.module]
include = "all"

[specs.core.mutex]
include = "none"

[specs.core.nil]
include = "all"

[specs.core.numeric]
include = "all"

[specs.core.objectspace]
include = "none"

[specs.core.proc]
include = "all"

[specs.core.process]
include = "none"

[specs.core.queue]
include = "all"

[specs.core.random]
include = "all"

[specs.core.range]
include = "all"

[specs.core.rational]
include = "all"

[specs.core.regexp]
include = "all"

[specs.core.signal]
include = "all"

[specs.core.sizedqueue]
include = "all"

[specs.core.string]
include = "all"

[specs.core.struct]
include = "all"

[specs.core.symbol]
include = "all"
skip = [
  "element_reference",
]

[specs.core.systemexit]
include = "all"

[specs.core.thread]
include = "none"

[specs.core.threadgroup]
include = "none"

[specs.core.time]
include = "all"

[specs.core.tracepoint]
include = "none"

[specs.core.true]
include = "all"

[specs.core.unboundmethod]
include = "all"

[specs.core.warning]
include = "all"
```

</details>

FWIW, this is why I don't like prioritizing getting the `spec-runner` in CI. I don't like chasing segfaults and I'd rather spend my time working toward removing Artichoke's dependency on mruby altogether instead of making the FFI bits less segfaulty.

eregon:
I added a sentence in the blog post:
> Update (February 2021): Artichoke now passes an estimated 22% of core library specs when trying to run all core specs with this PR, however those are not run in CI and might regress.

I think all other Rubies actually run specs in CI (important to avoid regressions), and in that blog post I relied on Ruby implementations untagging passing specs rather than trying to run all specs blindly and ignore errors, so I wanted to clarify how these numbers differ from the rest.

lopopolo:
Thanks @eregon!

I know Artichoke doesn't integrate with ruby/spec and MSpec the way it's intended.

It's still early days for Artichoke; when the time is right, these bits of polish will get prioritized.

lopopolo:
> I think all other Rubies actually run specs in CI (important to avoid regressions), and in that blog post I relied on Ruby implementations untagging passing specs rather than trying to run all specs blindly and ignore errors, so I wanted to clarify how these numbers differ from the rest.

Hi @eregon 👋 

I'm excited to share that Artichoke runs ruby/spec on every commit as of #1328, #1329, #1330, #1331, #1332, #1333.

I've filed https://github.com/ruby/spec/pull/844 to reflect this upstream in the ruby/spec README. 🚀 

Artichoke doesn't have `File` so I've kinda had to roll my own tagging. You can see the current set of tags here: https://github.com/artichoke/spec-state/blob/trunk/reports/tagged/latest.json.

eregon:
That's great!
Is there a place I can look at to easily find out how many `language`, `core`, and `library` are passing?
That would be quite useful for the next time I do an update like https://eregon.me/blog/2020/06/27/ruby-spec-compatibility-report.html

lopopolo:
> Is there a place I can look at to easily find out how many `language`, `core`, and `library` are passing?

Not yet. I'm planning on turning the raw data into a webpage at some point.

The suite currently only runs a subset of `core` specs. I'm working on being able to expand this to include some `library` specs #1325, #1327.

I've never run any `language` specs before, but I imagine getting them running wouldn't be too tricky.

Note: this isn't my biggest priority, but I am excited that this bit of infrastructure is in place to do things automatically. The hardest part is done I think 😄

PR #1101, opened 2021-02-21
Integer bitmask methods in rust

### LINKED ISSUE #1084: Implement integer bitmask functions in Rust
Currently, `Integer#allbits?`, `Integer#anybits?`, and `Integer#nobits?` are implemented in Ruby. These bit masking operations will be cheaper to implement in Rust.

https://github.com/artichoke/artichoke/blob/e13799db9c6a5b91c99509e119753297be91eba6/artichoke-backend/src/extn/core/integer/integer.rb#L8-L47

`Integer#div` will be a useful example to template off of.

The Rust bit is implemented in the `extn::integer::trampoline` module:

https://github.com/artichoke/artichoke/blob/e13799db9c6a5b91c99509e119753297be91eba6/artichoke-backend/src/extn/core/integer/trampoline.rs#L18-L22

and exposed to the Ruby VM in an `mruby` trampoline:

https://github.com/artichoke/artichoke/blob/e13799db9c6a5b91c99509e119753297be91eba6/artichoke-backend/src/extn/core/integer/mruby.rs#L49-L59

and injected into the VM:

https://github.com/artichoke/artichoke/blob/e13799db9c6a5b91c99509e119753297be91eba6/artichoke-backend/src/extn/core/integer/mruby.rs#L13

To match the current implementation, you'll want to convert the argument to an integer with [`implicitly_convert_to_int`](https://artichoke.github.io/artichoke/artichoke_backend/convert/fn.implicitly_convert_to_int.html) like `Integer#[]` does:

https://github.com/artichoke/artichoke/blob/e13799db9c6a5b91c99509e119753297be91eba6/artichoke-backend/src/extn/core/integer/trampoline.rs#L13

To make sure this conversion is the right thing to do, try this code in MRI to see what happens:

```
class A; def to_int; 0xFF; end; end
10.anybits?(A.new)
```

If this gives the same answer as `10.anybits?(0xFF)`, then we should be using the implicit converter, which can delegate to `#to_int`.

Remember to remove the definitions of these functions in Ruby.

To ensure the Rust functions you're creating are being called, you can print to stdout by adding something like:

```rust
println!("calling card from Rust code!");
```

sudopluto:
Fixes #1084

### DIFF
--- a/artichoke-backend/src/extn/core/integer/mod.rs
+++ b/artichoke-backend/src/extn/core/integer/mod.rs
@@ -179,16 +179,19 @@ impl Integer {
         }
     }
 
-    pub fn is_allbits(self, mask: Int) -> Result<bool, Error> {
-        Ok(self.as_i64() & mask == mask)
+    #[must_use]
+    pub fn is_allbits(self, mask: Int) -> bool {
+        self.as_i64() & mask == mask
     }
 
-    pub fn is_anybits(self, mask: Int) -> Result<bool, Error> {
-        Ok(self.as_i64() & mask != 0)
+    #[must_use]
+    pub fn is_anybits(self, mask: Int) -> bool {
+        self.as_i64() & mask != 0
     }
 
-    pub fn is_nobits(self, mask: Int) -> Result<bool, Error> {
-        Ok(self.as_i64() & mask == 0)
+    #[must_use]
+    pub fn is_nobits(self, mask: Int) -> bool {
+        self.as_i64() & mask == 0
     }
 
     #[must_use]
--- a/artichoke-backend/src/extn/core/integer/mod.rs
+++ b/artichoke-backend/src/extn/core/integer/mod.rs
@@ -179,6 +179,18 @@ impl Integer {
         }
     }
 
+    pub fn is_allbits(self, mask: Int) -> Result<bool, Error> {
+        Ok(self.as_i64() & mask == mask)
+    }
+
+    pub fn is_anybits(self, mask: Int) -> Result<bool, Error> {
+        Ok(self.as_i64() & mask != 0)
+    }
+
+    pub fn is_nobits(self, mask: Int) -> Result<bool, Error> {
+        Ok(self.as_i64() & mask == 0)
+    }
+
     #[must_use]
     pub const fn size() -> usize {
         mem::size_of::<Int>()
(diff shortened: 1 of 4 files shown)


artichoke-backend/src/extn/core/integer/mod.rs
```diff
@@ -179,6 +179,18 @@ impl Integer {
         }
     }
 
+    pub fn is_allbits(self, mask: Int) -> Result<bool, Error> {
+        Ok(self.as_i64() & mask == mask)
```
lopopolo: This function doesn't need to be fallible:

```suggestion
    pub fn is_allbits(self, mask: Int) -> bool {
        self.as_i64() & mask == mask
```

artichoke-backend/src/extn/core/integer/mod.rs
```diff
@@ -179,6 +179,18 @@ impl Integer {
...
+    }
+
+    pub fn is_anybits(self, mask: Int) -> Result<bool, Error> {
+        Ok(self.as_i64() & mask != 0)
```
lopopolo: This function doesn't need to be fallible:

```suggestion
    pub fn is_anybits(self, mask: Int) -> bool {
        self.as_i64() & mask != 0
```

artichoke-backend/src/extn/core/integer/mod.rs
```diff
@@ -179,6 +179,18 @@ impl Integer {
...
+    }
+
+    pub fn is_nobits(self, mask: Int) -> Result<bool, Error> {
+        Ok(self.as_i64() & mask == 0)
```
lopopolo: This function doesn't need to be fallible:

```suggestion
    pub fn is_nobits(self, mask: Int) -> bool {
        self.as_i64() & mask == 0
```

lopopolo (changes requested):
This looks great! Let's remove the `Result` wrapper from these functions since they are infallible.

sudopluto:
cool, will do. do I need unit tests? probably not right?

lopopolo:
> cool, will do. do I need unit tests? probably not right?

This is simple enough where I think we can land it without unit tests.

lopopolo (approved):
Looks good to me, but I think clippy is upset about something (I'm on mobile and can't see the build log)

lopopolo (approved):

PR #1105, opened 2021-02-24
Remove duplicate impls of ByteLiteral and move it to string escape utils

lopopolo:
Both `spinoso-symbol` and `spinoso-regexp` had impls of a `ByteLiteral`
iterator which wrapped `scolapasta_string_escape::Literal` to more
ergonomically deal with multi-byte UTF-8 decoding failures from `bstr`.

This change extracts the best parts of these diverging impls and
relocates them to `scolapasta-string-escape`. The new iterator has a new
name, fallible construction from slices (instead of panicking in a
`From` impl), public constructors, docs, `FusedIterator` and
`DoubleEndedIterator` impls, and an optional `std::error::Error` impl.

This PR adds an on-by-default `std` feature to
`scolapasta-string-escape` and bumps the crate version to 0.2.0. All
workspace deps are updated and disable default features.

All subworkspaces have their lockfiles rebuilt.

This turns out to be the API that consumers of
`scolapasta-string-escape` really wanted. After this refactor, the
`Literal` type is not referenced anywhere in this repo outside of the
string escape crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -515,7 +515,7 @@ version = "0.1.0"
 
 [[package]]
 name = "scolapasta-string-escape"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "bstr",
 ]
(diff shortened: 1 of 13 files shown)


lopopolo:
cc @tummychow this change ended up being bigger than I intended :sweat_smile:

PR #1119, opened 2021-03-06
Fix nightly build breakage due to rename of broken_intra_doc_links lint

lopopolo:
The `broken_intra_doc_links` lint was made into a `rustdoc` tool lint in https://github.com/rust-lang/rust/pull/80527. This PR also introduced the `rustdoc` tool name. Use of this tool name on releases older than the latest nightly are hard compile time errors.

To work around this, remove `#![warn(broken_intradoc_links)]` from crate-level pragmas and instead add `-D rustdoc::broken_intra_doc_links` to `RUSTDOCFLAGS` in the `rustdoc` GitHub Actions workflow and to `Rakefile` docs tasks.

This PR also removes `-D warnings` from the LeakSanitizer workflow to remove future sources of breakage. `-D warnings` is asserted in other parts of CI and is not necessary for running LeakSan.


lopopolo:
Docs fail to build because of https://github.com/rust-lang/rust/pull/80527#issuecomment-791981435. Not sure what the right way to fix this is.

PR #1121, opened 2021-03-07
Implement mruby value/float numeric conversions in Rust

lopopolo:
Implement `mrb_to_flo` and `mrb_int_value` in Rust and expose to C via `no_mangle` FFI functions. This commit `ifndef`s these functions out in C when `-DARTICHOKE` is given when compiling.

This commit adds several fixnum conversions to `Float`, notably `Float::try_into_fixnum`, which implements the mruby `FIXABLE_FLOAT` and `FIXABLE` macros.

This code has been sitting on a branch for over a year; before I rebased the mainline commit the branch was based on was 0392cec9f45491fa7135dd2ff8b6f762aaad0ed0. Rather than implement all of `Numeric` all at once, landing just these converters feels like a good first step.

### DIFF
--- a/artichoke-backend/src/extn/core/float/mod.rs
+++ b/artichoke-backend/src/extn/core/float/mod.rs
@@ -202,11 +202,42 @@ impl Float {
         Self(0.0)
     }
 
+    /// Construct a new `Float` with a given [`f64`].
+    #[inline]
+    #[must_use]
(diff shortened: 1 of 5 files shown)


lopopolo:
⚠️ This PR modifies the vendored copy of mruby in a way that is not reflected in the Artichoke fork.

PR #1122, opened 2021-03-07
Make Integer and Float wrappers transparent

lopopolo:
- Annotate these types with `#[repr(transparent)]`.
- Implement `fmt::Debug` as a passthrough to the inner primitive type.
- Implement `fmt::Display` as a passthrough to the inner primitive type.
- Move `float::init` to a `float::mruby` module.
- Remove `artichoke_` prefix from `Integer` C mruby trampolines.


lopopolo:
Tagging with `A-performance` because the `repr` annotations may affect how these immediates are passed via registers.

PR #1140, opened 2021-03-21
Specialize `fs` module for loading Ruby sources

lopopolo:
The `artichoke_backend::fs` module was designed for two roles:

- Loading Ruby source code and Rust extensions, aka managing the Ruby
  `$LOAD_PATH` and `$LOADED_FEATURES` globals and VM functionality.
- Being a virtual filesystem abstraction for eventually implementing
  `File` and `Dir`. This function is yet to be used.

The addition of a `RUBYLIB` source loader made it clear that evolving
this module would be easier if it could be specialized to only do code
loading. This commit renames the module from `fs` to `load_path` and
starts implementing the specialization.

- Rust extension hook APIs are removed from the `Rubylib` and `Native`
  loaders.
- `write_file` APIs are removed from the `Rubylib` and `Native` loaders.
- The `Memory` loader enforces that reads and writes occur beneath its
  mount point.
- `Eval::eval_file` and `LoadSources` APIs do not resolve absolute paths
  where possible.

The cargo features for activating the native and `RUBYLIB` loaders are
renamed to reflect that they affect the load path funcionality and that
they are source loaders.


lopopolo:
I'd like to split `load_path` into its own crate as a followup to this PR.

PR #1142, opened 2021-03-21
Add Time::at for seconds and microseconds since EPOCH

michaeltaras:
Adds a partial implementation of `Time::at` to handle passing in an integer `seconds` and optional integer `microseconds` value since the EPOCH to create a new Time object.

### DIFF
--- a/artichoke-backend/src/extn/core/time/mruby.rs
+++ b/artichoke-backend/src/extn/core/time/mruby.rs
@@ -18,7 +18,7 @@ pub fn init(interp: &mut Artichoke) -> InitializeResult<()> {
         .value_is_rust_object()
         // Constructor
         .add_self_method("now", time_self_now, sys::mrb_args_none())?
-        .add_self_method("at", time_self_at, sys::mrb_args_any())?
+        .add_self_method("at", time_self_at, sys::mrb_args_req(1))?
         .add_self_method("utc", time_self_mkutc, sys::mrb_args_any())?
         .add_self_method("gm", time_self_mkutc, sys::mrb_args_any())?
         .add_self_method("local", time_self_mktime, sys::mrb_args_any())?
@@ -112,10 +112,11 @@ unsafe extern "C" fn time_self_now(mrb: *mut sys::mrb_state, _slf: sys::mrb_valu
 }
 
 unsafe extern "C" fn time_self_at(mrb: *mut sys::mrb_state, _slf: sys::mrb_value) -> sys::mrb_value {
-    let args = mrb_get_args!(mrb, *args);
+    let (seconds, microseconds) = mrb_get_args!(mrb, required = 1, optional = 1);
     unwrap_interpreter!(mrb, to => guard);
-    let args = args.iter().copied().map(Value::from);
-    let result = trampoline::at(&mut guard, args);
+    let seconds = Value::from(seconds);
+    let microseconds = Value::from(microseconds);
+    let result = trampoline::at(&mut guard, seconds, microseconds);
     match result {
         Ok(value) => value.inner(),
         Err(exception) => error::raise(guard, exception),
--- a/artichoke-backend/src/extn/core/time/mruby.rs
+++ b/artichoke-backend/src/extn/core/time/mruby.rs
@@ -115,7 +115,7 @@ unsafe extern "C" fn time_self_at(mrb: *mut sys::mrb_state, _slf: sys::mrb_value
     let (seconds, microseconds) = mrb_get_args!(mrb, required = 1, optional = 1);
     unwrap_interpreter!(mrb, to => guard);
     let seconds = Value::from(seconds);
-    let microseconds = Value::from(microseconds);
+    let microseconds = microseconds.map(Value::from);
     let result = trampoline::at(&mut guard, seconds, microseconds);
     match result {
         Ok(value) => value.inner(),
--- a/artichoke-backend/src/extn/core/time/trampoline.rs
+++ b/artichoke-backend/src/extn/core/time/trampoline.rs
@@ -1,5 +1,6 @@
 //! Glue between mruby FFI and `Time` Rust implementation.
 
+use crate::convert::implicitly_convert_to_int;
 use crate::extn::core::time::Time;
 use crate::extn::prelude::*;
 
@@ -11,13 +12,17 @@ pub fn now(interp: &mut Artichoke) -> Result<Value, Error> {
     Ok(result)
 }
 
-pub fn at<T>(interp: &mut Artichoke, args: T) -> Result<Value, Error>
-where
-    T: IntoIterator<Item = Value>,
-{
-    let _ = interp;
-    let _ = args;
-    Err(NotImplementedError::new().into())
+pub fn at(interp: &mut Artichoke, seconds: Value, microseconds: Value) -> Result<Value, Error> {
+    let seconds = implicitly_convert_to_int(interp, seconds)?;
+    let sub_second_nanos = if microseconds.is_nil() {
+        0
+    } else {
+        implicitly_convert_to_int(interp, microseconds)? * 1_000
+    };
+
+    let time = Time::at(seconds, sub_second_nanos);
+    let result = Time::alloc_value(time, interp)?;
+    Ok(result)
 }
 
 pub fn mkutc<T>(interp: &mut Artichoke, args: T) -> Result<Value, Error>
--- a/artichoke-backend/src/extn/core/time/trampoline.rs
+++ b/artichoke-backend/src/extn/core/time/trampoline.rs
@@ -3,6 +3,7 @@
 use crate::convert::implicitly_convert_to_int;
 use crate::extn::core::time::Time;
 use crate::extn::prelude::*;
+use spinoso_time::MICROS_IN_NANO;
 
 // Constructor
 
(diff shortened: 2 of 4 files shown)


michaeltaras:
This is my first PR into the project and I’m still getting familiar with the best way to do things in Artichoke as well as Rust in general. Please let me know if anything could be improved upon here.  I hope it’s not too much of a hassle but I had a couple questions:

Clippy pointed out a casting issue since microseconds comes in as an i64 and it eventually needs to end up as a u32 for sub_second_nanos.  Maybe there’s a more elegant way to handle that?

I also noticed MRI handles negative sub-seconds even though I don’t see any tests for that in ruby/spec.  I matched their behavior but I’m wondering if that is even needed?

Thanks!

artichoke-backend/src/extn/core/time/mruby.rs
```diff
@@ -112,10 +112,11 @@ unsafe extern "C" fn time_self_now(mrb: *mut sys::mrb_state, _slf: sys::mrb_valu
 }
 
 unsafe extern "C" fn time_self_at(mrb: *mut sys::mrb_state, _slf: sys::mrb_value) -> sys::mrb_value {
-    let args = mrb_get_args!(mrb, *args);
+    let (seconds, microseconds) = mrb_get_args!(mrb, required = 1, optional = 1);
     unwrap_interpreter!(mrb, to => guard);
-    let args = args.iter().copied().map(Value::from);
-    let result = trampoline::at(&mut guard, args);
+    let seconds = Value::from(seconds);
+    let microseconds = Value::from(microseconds);
```
lopopolo: since `microseconds` is an optional argument, the `mrb_get_args!` macro returns an `Option<sys::mrb_value>`.

We can preserve this type information and pass an `Option<Value>` into the trampoline, which avoids the `is_nil` check by doing this:

```suggestion
    let microseconds = microseconds.map(Value::from);
```

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -11,13 +12,17 @@ pub fn now(interp: &mut Artichoke) -> Result<Value, Error> {
...
 }
 
-pub fn at<T>(interp: &mut Artichoke, args: T) -> Result<Value, Error>
-where
-    T: IntoIterator<Item = Value>,
-{
-    let _ = interp;
-    let _ = args;
-    Err(NotImplementedError::new().into())
+pub fn at(interp: &mut Artichoke, seconds: Value, microseconds: Value) -> Result<Value, Error> {
```
lopopolo: with the change in the mruby trampoline, this signature will take `Option<Value>` for `microseconds`:

```suggestion
pub fn at(interp: &mut Artichoke, seconds: Value, microseconds: Option<Value>) -> Result<Value, Error> {
```

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -11,13 +12,17 @@ pub fn now(interp: &mut Artichoke) -> Result<Value, Error> {
...
+    let seconds = implicitly_convert_to_int(interp, seconds)?;
+    let sub_second_nanos = if microseconds.is_nil() {
+        0
+    } else {
+        implicitly_convert_to_int(interp, microseconds)? * 1_000
+    };
```
lopopolo: It looks like MRI adds overflow microseconds to the `seconds` parameter, which means I think we'll want something like:

```suggestion
    let seconds = implicitly_convert_to_int(interp, seconds)?;
    let (seconds, sub_second_nanos) = if let Some(microseconds) = microseconds {
        let microseconds = implicitly_convert_to_int(interp, microseconds)?;
        let overflow_seconds = microseconds / 1_000_000;
        let sub_second_nanos = u32::try_from((microseconds % 1_000_000) * 1_000).map_err(|_| ArgumentError::with_message("Time too large"));
        let seconds = seconds.checked_add(overflow_seconds).ok_or(|| ArgumentError::with_message("Time too large"))?;
        (seconds, sub_second_nanos)
    } else {
        (seconds, 0_u32)
    };
```

This setup ensures we return an error if we overflow, ensures `sub_second_nanos` is in an acceptable range, and allows us to avoid an `as` cast which might truncate or otherwise lose precision.

spinoso-time/src/time/chrono/build.rs
```diff
@@ -50,6 +51,38 @@ impl Time {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use spinoso_time::Time;
+    /// let epoch = Time::at(0, 0);
+    /// let epoch_plus_1_nano = Time::at(0, 1);
+    /// ```
+    #[inline]
+    #[must_use]
+    pub fn at(seconds: i64, sub_second_nanos: i64) -> Self {
```
lopopolo: oh I see, I like this better than my suggestion above to keep the `i64` type on `sub_second_nanos`, but if we do that, the conversion is fallible, so I'd like to see us return `Option<Self>`. This way, when we use checked addition and checked conversions to `u32`, we can fail.

In this case, I'd prefer to return `Option` rather than `Result` since the only way constructing a `Time` with `at` can fail is if the failure is of the `Time is too large` case, which we can document.

spinoso-time/src/time/chrono/build.rs
```diff
@@ -50,6 +51,38 @@ impl Time {
...
+        let offset = Offset::Local;
+
+        let timestamp = seconds + (sub_second_nanos / i64::from(NANOS_IN_SECOND));
```
lopopolo: let's used a checked addition here like my suggestion above in the trampoline so we don't accidentally wrap around. (in debug mode this would be a panic, which you can test by doing a `Time.at(1e60, 1e60)`.

spinoso-time/src/time/chrono/build.rs
```diff
@@ -50,6 +51,38 @@ impl Time {
...
+        let sub_second_nanos = sub_second_nanos % i64::from(NANOS_IN_SECOND);
+
+        // Handle negative sub_seconds_nanos
```
lopopolo: In cases like these where the behavior is non obvious, I like to inline sample executions from an irb repl in the comments, similar to the ones here:

https://github.com/artichoke/artichoke/blob/1498e9d99d4313bfecfd98744cc4db162eb27ccf/artichoke-backend/src/convert/implicit.rs#L75-L88

spinoso-time/src/time/chrono/build.rs
```diff
@@ -50,6 +51,38 @@ impl Time {
...
+        let (timestamp, sub_second_nanos) = if sub_second_nanos > 0 {
+            (timestamp, sub_second_nanos)
+        } else {
+            (timestamp - 1, (i64::from(NANOS_IN_SECOND) - sub_second_nanos.abs()))
+        };
+
+        Self {
+            timestamp,
+            sub_second_nanos: sub_second_nanos as u32,
```
lopopolo: let's use a `u32::try_from` conversion here like my suggestion above in the trampoline.

artichoke-backend/src/extn/core/time/trampoline.rs
lopopolo: After reviewing the `Time::at` impl in `spinoso-time`, I like the direction you've taken to accept `i64` in `spinoso-time`, so the suggestions above are best relocated to `Time::at` in `spinoso-time`. Here' all we should be doing is unmarshalling arguments, so this suggestion should probably look like:

```suggestion
    let seconds = implicitly_convert_to_int(interp, seconds)?;
    let sub_second_nanos = if let Some(microseconds) = microseconds {
        implicitly_convert_to_int(interp, microseconds)?.checked_mul(1_000).ok_or_else(|| ArgumentError::with_message("Time too large"))?;
    } else {
        0_i64
    };
```

Can we turn `1_000` into a named constant in `spinoso-time` that we use here?

spinoso-time/src/time/chrono/build.rs
```diff
@@ -73,4 +106,25 @@ mod tests {
...
+        assert_eq!(time.timestamp, 101);
+        assert_eq!(time.sub_second_nanos, 1);
+    }
+
+    #[test]
+    fn time_at_with_negative_sub_second_nanos() {
+        let time = Time::at(100, -1);
+        assert_eq!(time.timestamp, 99);
+        assert_eq!(time.sub_second_nanos, 999_999_999);
+    }
```
lopopolo: These tests are great!

Can we add one more that exercises the "time too large flow" once we add the checked addition?

lopopolo (changes requested):
Hi @michaeltaras 👋 this looks great! ⏳ 

I love the approach you've taken by limiting the logic in the `artichoke-backend` crate to only argument unmarshalling.

I'd like to make sure `Time::at` is safe in the presence of very large numbers so we should look into using "checked arithmetic" whenever we're multiplying and adding: 

- https://doc.rust-lang.org/std/primitive.i64.html#method.checked_add
- https://doc.rust-lang.org/std/primitive.i64.html#method.checked_mul

You look like you found your way around here pretty nicely 🙏

artichoke-backend/src/extn/core/time/mruby.rs
```diff
@@ -112,10 +112,11 @@ unsafe extern "C" fn time_self_now(mrb: *mut sys::mrb_state, _slf: sys::mrb_valu
 }
 
 unsafe extern "C" fn time_self_at(mrb: *mut sys::mrb_state, _slf: sys::mrb_value) -> sys::mrb_value {
-    let args = mrb_get_args!(mrb, *args);
```
lopopolo, line 115: eventually when we add support for all variants of `Time::at`, it will probably be easier to process the arguments as an iterator, but for now, this change is OK to me.

michaeltaras:
Thanks for the great feedback!  It's all super helpful.  I'll run through everything and let you know when it's ready for another look.

spinoso-time/src/time/chrono/build.rs
```diff
@@ -50,11 +53,56 @@ impl Time {
...
+        // => -1
+        // [2.6.3] > Time.at(0, -1).nsec
+        // => 999999000
+        // ```
+        let (timestamp, sub_second_nanos) = if sub_second_nanos > 0 {
+            (timestamp, sub_second_nanos)
+        } else {
+            (
+                timestamp.checked_sub(1)?,
+                i64::from(NANOS_IN_SECOND).checked_sub(sub_second_nanos.abs())?,
```
lopopolo, line 90: I think there's some carry over behavior here when `sub_second_nanos` is negative and greater than 1s.

I think this does the right thing, but I think it would be easier to follow if the overflow checking was done in each branch.

LGTM for now, I'll make a followup PR.

At the very least, I think we're missing a test.

```
[2.6.3] > t = Time.at(10, -1e6); [t.sec, t.nsec]
=> [9, 0]
[2.6.3] > t = Time.at(10, -1e6 - 1); [t.sec, t.nsec]
=> [8, 999999000]
```

spinoso-time/src/time/chrono/build.rs
```diff
@@ -50,11 +53,56 @@ impl Time {
...
+            )
+        };
+
+        Some(Self {
+            timestamp,
+            sub_second_nanos: u32::try_from(sub_second_nanos).ok()?,
```
lopopolo, line 96: nit: I'd like to see the type of `sub_second_nanos` here to already be a `u32` so the construction of the struct doesn't have any complex logic in it.

lopopolo (approved):
I left a couple of nits, but LGTM.

I'll make a followup PR and  tag you for review @michaeltaras

PR #1144, opened 2021-03-24
Add test for negative nanos overflow to `Time::at`

lopopolo:
Followup to https://github.com/artichoke/artichoke/pull/1142.

cc @michaeltaras


michaeltaras:
Thanks for cleaning this up @lopopolo.  I like the idea of very clearly explaining all the math in there.

PR #1148, opened 2021-03-28
implement Array#shuffle in ruby land

KBuild:
Hello!

Thanks to create this project!

I implement `Array#shuffle` and `Array#shuffle!`

I also know that its going to Rust implementation

But I think this logic can be made in Ruby and it's simple

If you think this PR makes sense, please accept it

If you don't, please close this PR

I really have expecting in this project!

Thanks

### DIFF
--- a/artichoke-backend/src/extn/core/array/array.rb
+++ b/artichoke-backend/src/extn/core/array/array.rb
@@ -1105,12 +1105,20 @@ class Array
     replace(result)
   end
 
-  def shuffle(_rng = (not_set = true)) # rubocop:disable Lint/UselessAssignment
-    raise NotImplementedError, 'TODO implement in Rust'
+  def shuffle(_rng = (not_set = true))
+    rng = _rng.nil? ? Random.new : _rng
+    shuffled_orders = (0..self.size - 1).map.with_index {|_n, i| [rng.rand, i]}.sort {|a,b| a.first <=> b.first}
+    shuffled_orders.map {|_n, i| i}.map do |index|
+      self[index]
+    end
   end
 
-  def shuffle!(_rng = (not_set = true)) # rubocop:disable Lint/UselessAssignment
-    raise NotImplementedError, 'TODO implement in Rust'
+  def shuffle!(_rng = (not_set = true))
+    raise FrozenError, "can't modify frozen Array" if frozen?
+    return self if length <= 1
+
+    self[0, length] = shuffle(_rng)
+    self
   end
 
   def slice!(*args)
--- a/artichoke-backend/src/extn/core/array/array.rb
+++ b/artichoke-backend/src/extn/core/array/array.rb
@@ -1105,19 +1105,24 @@ class Array
     replace(result)
   end
 
-  def shuffle(rng = nil)
-    rng_instance = rng.nil? ? Random.new : rng
-    shuffled_orders = (0..size - 1).map.with_index { |_n, i| [rng_instance.rand, i] }.sort { |a, b| a.first <=> b.first }
-    shuffled_orders.map { |_n, i| i }.map do |index|
-      self[index]
+  def shuffle(random: (not_set = true))
+    random = Random::DEFAULT if not_set
+    shuffled_orders = (0...size).map { |idx| [random.rand, idx] }.sort { |a, b| a[0] <=> b[0] }
+    shuffled_orders.map { |_n, idx| idx }.map do |idx|
+      self[idx]
     end
   end
 
-  def shuffle!(rng = nil)
+  def shuffle!(random: (not_set = true))
     raise FrozenError, "can't modify frozen Array" if frozen?
     return self if length <= 1
 
-    self[0, length] = shuffle(rng)
+    self[0, length] =
+      if not_set
+        shuffle
+      else
+        shuffle(random: random)
+      end
     self
   end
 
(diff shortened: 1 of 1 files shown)


lopopolo:
Hi @KBuild thanks for the PR. I'm acking that I've seen it, but I haven't had time to give it a review yet.

I noticed the Ruby linter is failing. You should be able to fix this by running `bundle exec rake` on your local branch.

KBuild:
Thank you for reaction to my PR!

I fix that lint error!

@lopopolo

artichoke-backend/src/extn/core/array/array.rb
```diff
@@ -1105,12 +1105,20 @@ def select!(&block)
     replace(result)
   end
 
-  def shuffle(_rng = (not_set = true)) # rubocop:disable Lint/UselessAssignment
-    raise NotImplementedError, 'TODO implement in Rust'
+  def shuffle(rng = nil)
```
lopopolo: the default assignment initializer is required because passing `nil` to `Array#shuffle` should raise an exception.

When `not_set` is `true`, `Array#shuffle` should use `Random::Default`

artichoke-backend/src/extn/core/array/array.rb
```diff
@@ -1105,12 +1105,20 @@ def select!(&block)
...
+    rng_instance = rng.nil? ? Random.new : rng
+    shuffled_orders = (0..size - 1).map.with_index { |_n, i| [rng_instance.rand, i] }.sort { |a, b| a.first <=> b.first }
```
lopopolo: this is a clever approach.

lopopolo: This suggestion adds support for `[].shuffle` using the default `Random` instance and makes a few performance optimizations. Let me know what you think @KBuild.

```suggestion
  def shuffle(rng = (not_set = true))
    rng = Random::DEFAULT if not_set
    shuffled_orders = (0...size).map { |idx| [rng.rand, idx] }.sort { |a, b| a[0] <=> b[0] }
```

lopopolo (changes requested):
Thanks @KBuild. Sorry this took a while for me to review. This looks great and I'm glad you found a way to do this in pure Ruby. I have one change I'd like to make: MRI uses the `Random::DEFAULT` RNG when a parameter is not given to `Array#shuffle`.

lopopolo:
It looks like the failing Documentation build is due to a nightly regression that was fixed upstream. Once you push a new commit with the `Random::DEFAULT` fix, the build should fix itself.

artichoke-backend/src/extn/core/array/array.rb
KBuild: Thank you for informing suggestion!
I prefer the code keeping consistency rules
I will fix that

KBuild: Oh, I check some document and find `Array#shuffle` in this link
https://docs.ruby-lang.org/en/2.6.0/Array.html#method-i-shuffle

It seems like to using `random` keyword arguments
So I think we should use

```rb
def shuffle(random: (not_set = true))
...
end
def shuffle(random: (not_set = true))
...
end
```

How about this?

I'm going to commit this idea, and please check code 🙏

artichoke-backend/src/extn/core/array/array.rb
```diff
@@ -1105,12 +1105,25 @@ def select!(&block)
     replace(result)
   end
 
-  def shuffle(_rng = (not_set = true)) # rubocop:disable Lint/UselessAssignment
-    raise NotImplementedError, 'TODO implement in Rust'
+  def shuffle(random: (not_set = true))
+    random = Random::DEFAULT if not_set
```
lopopolo:
```suggestion
    random = Random::DEFAULT if not_set # rubocop:disable Lint/DeprecatedConstants
```

artichoke-backend/src/extn/core/array/array.rb
```diff
@@ -1105,12 +1105,25 @@ def select!(&block)
     replace(result)
   end
 
-  def shuffle(_rng = (not_set = true)) # rubocop:disable Lint/UselessAssignment
-    raise NotImplementedError, 'TODO implement in Rust'
+  def shuffle(random: (not_set = true))
```
lopopolo, line 1108: thanks for catching this :)

artichoke-backend/src/extn/core/array/array.rb
```diff
@@ -1105,12 +1105,25 @@ def select!(&block)
...
+  def shuffle!(random: (not_set = true))
+    raise FrozenError, "can't modify frozen Array" if frozen?
+    return self if length <= 1
+
+    self[0, length] =
+      if not_set
+        shuffle
+      else
+        shuffle(random: random)
+      end
```
lopopolo, line 1125: looks good here thank you.

lopopolo (approved):

artichoke-backend/src/extn/core/array/array.rb
lopopolo: Looks good. Thanks for checking the docs.

lopopolo:
Thanks for the updates @KBuild. I've added one commit on top of your branch that addresses a RuboCop deprecation warning. Assigning to myself to make sure this gets merged.

Thanks for contributing to Artichoke! 💎

PR #1175, opened 2021-04-23
Update Rakefile and RuboCop configuration to work on Windows machines

lopopolo:
- Force line feed endings in RuboCop configuration instead of the default "native" endings.
- Fix quotes for prettier globs so they work in Windows Terminal and PowerShell.
- Add docs on how to set up a toolchain on Windows to `BUILD.md`.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -46,6 +46,12 @@ brew install rustup-init
 rustup-init
 ```
 
+On Windows, you can install rustup from the official site and follow the
+prompts: <https://rustup.rs/>. This requires a download of Visual Studio (the
+[Community Edition][vs-community] is sufficient) and several C++ packages
+selected through the VS component installer. (I'm not sure which packages are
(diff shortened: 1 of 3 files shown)


BUILD.md
```diff
@@ -74,6 +80,9 @@ compiler.
 Artichoke specifically requires clang. WebAssembly targets require clang-8 or
 newer.
 
+On Windows, install the latest LLVM distribution from GitHub and add LLVM to
+your PATH: <https://github.com/llvm/llvm-project/releases>.
+
```
dmtaub, line 85: I already had llvm when things didn't work, but after your suggestion also downloaded a ruby install, so that might be necessary as well...unless I did something else that helped...

lopopolo: Yes, Ruby is required to build. I believe this is outlined in `BUILD.md` a bit further down.

PR #1271, opened 2021-07-26
Make `String::center` return `Result`

lopopolo:
`String::center` can fail and raise an `ArgumentError` if called with a zero-width padding bytestring.

This commit makes `String::center` fallible with a new error type, `CenterError`.

This commit fixes a bug where `Center` with an empty padding bytestring would never terminate.


lopopolo:

PR #1274, opened 2021-07-29
Upgrade Rust toolchain to 1.54.0

lopopolo:
https://blog.rust-lang.org/2021/07/29/Rust-1.54.0.html

This PR addresses many new and updated clippy lints.

This PR uses the new function-like macro in attribute position changes to simplify the testing of `README.md` code blocks in doc tests.


lopopolo:
It's likely the doctest changes will result in the artichoke/nightly CI run to fail tonight, which will require bumping the `RUST_VERSION` arg embedded in the `Dockerfile`s.

Let's wait for that run to fail before pushing a fix and link it here.

lopopolo:
> Let's wait for that run to fail before pushing a fix and link it here.

https://github.com/artichoke/docker-artichoke-nightly/pull/42

PR #1299, opened 2021-08-04
Add experimental crates for Ruby feature loading

lopopolo:
Add `artichoke-load-path` and fix some bugs around the `RUBYLIB` loader:

- The `RUBYLIB` loader should skip "explicit relative" paths that start with `./` or `../`.
- Loader APIs that return `bool` now return `Option<bool>` to allow loaders to indicate they do not handle a given path.

`artichoke-load-path` is an extracted and modified version of `Rubylib` loader already in `artichoke_backend::load_path`. `artichoke_load_path::Rubylib` maintains more or less the same API. `artichoke_backend::load_path` no longer exports a `RUBYLIB` loader, moving it to an implementation detail. The `RUBYLIB` loader in `artichoke-backend` has been removed.

Add `mezzaluna-feature-loader` to re-imagine what the feature loading API could look like. `mezzaluna-feature-loader` will not implement all loader components such that they share a source-compatible API. `mezzaluna-feature-loader` will export one loader type whose implementation will be configurable with Cargo features.

This crate is a work in progress. For now, it includes a stub loader interface, a reworked `RUBYLIB` loader, and a single dedicated `LoadedFeatures` ordered, set-like container.


lopopolo:
> A _mezzaluna_ (/ˌmɛtsəˈluːnə/; Italian: [ˌmɛddzaˈluːna]) is a knife consisting of one or more[1] curved blades with a handle on each end, which is rocked back and forth chopping the ingredients below with each movement.[2] They most commonly have a single blade, but are sometimes seen with two or three blades.[3][4][5]
>
> It is typically used for mincing herbs or garlic, but it can be used for chopping other things such as cheese or meat.[2] Very large single blade versions are sometimes used for pizza.[1] Common uses in Italy include preparation of a soffritto or a pesto, etc.

https://en.wikipedia.org/wiki/Mezzaluna

Like the _spinoso_ crate prefix, _mezzaluna_ crates are utility crates, but these are intended to be consumed directly by the interpreter and may be tightly coupled to Ruby things.

PR #1300, opened 2021-08-05
Refactor artichoke-backend build to use `bindgen` executable instead of dev-dependency

lopopolo:
This PR updates the `artichoke-backend` `build.rs` to invoke bindgen via its CLI instead of as a library.

This change removes 10+ deps from `Cargo.lock` and is a workaround for bugs like https://github.com/rust-lang/cargo/issues/9756 which prevent production dependencies from upgrading to their most recent versions.

This PR translates the existing library invocation of bindgen to a command line invocation.

The build script first attempts to use a globally-installed (and available in `$PATH`) `bindgen` (such as is the case on @lopopolo's devbox and GitHub Actions runners). If this fails, the build scripts `cargo install bindgen`s to the `OUT_DIR` and tries again.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -61,6 +61,18 @@ rustup component add rustfmt
 rustup component add clippy
 ```
 
+### Bindgen
+
+Artichoke generates Rust declarations for C code at build time using
+[`bindgen`]. Although not required to build Artichoke, installing `bindgen`
+globally will speed up the compilation cycle.
+
+To install bindgen using `cargo`:
+
+```sh
+cargo install --version 0.59.1 bindgen
+```
+
 ### Rust Crates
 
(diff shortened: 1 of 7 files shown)


lopopolo:
Before:

```console
$ rg '^name = "' Cargo.lock | cut -d' ' -f3 | wc -l
     101
```

After:

```console
$ rg '^name = "' Cargo.lock | cut -d' ' -f3 | wc -l
      85
```

lopopolo:
Excluding workspace-local packages, Artichoke is down to 68 deps.

```console
$ rg '^name = "' Cargo.lock | grep -v artichoke | grep -v mezzaluna | grep -v scolapasta | grep -v spinoso | cut -d' ' -f3 | wc -l
      68
```

PR #1308, opened 2021-08-05
Revendor emsdk headers as of 2.0.9-upstream

lopopolo:
Followup to #1305 and #1306.

    rm -r artichoke-backend/vendor/emscripten/system/*
    cp -r ../playground/emsdk/upstream/emscripten/system/* artichoke-backend/vendor/emscripten/system/

These headers will match the version of emscripten used to compile the
playground with Rust 1.54.0 which requires LLVM 12.

See also rust-lang/rust#85821 for how this emsdk version was chosen.


lopopolo:
This PR is a dependency for merging https://github.com/artichoke/playground/pull/606.

PR #1312, opened 2021-08-06
Fix artichoke-backend build when bindgen is not globally installed

lopopolo:
Regression was uncovered in the nightly docker build and was introduced in artichoke/artichoke#1300.


lopopolo:
This bug surfaced in the nightly docker builder: https://github.com/artichoke/docker-artichoke-nightly/runs/3257931601?check_suite_focus=true

PR #1326, opened 2021-08-16
Improve parsing of radix parameter for `Kernel::Integer`

lopopolo:
- Move validation for proper radix range of `[2, 36]` to `Radix::new` constructor.
- Add tests for `Radix::new` constructor.
- Fix `should_panic` test so it is properly implemented and passes without the `should_panic` annotation.
- Handle -1 radix which should be parsed to `Some(10)`.
- Handle `[-36, -2]` radixes which should be parsed as if they were positive.
- Add tests for `TryConvertMut` impl for `Radix`.
- Add tests for converting `i64::MAX` and `i64::MIN` to check for panics from improper negation.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -31,6 +31,7 @@ GEM
 
 PLATFORMS
   ruby
+  x64-mingw32
 
 DEPENDENCIES
   bundler-audit (~> 0.8)
--- a/artichoke-backend/src/extn/core/kernel/integer.rs
+++ b/artichoke-backend/src/extn/core/kernel/integer.rs
@@ -361,6 +361,26 @@ mod tests {
     use super::{method as integer, Radix};
 
     #[test]
+    fn radix_new_validates_radix_is_nonzero() {
+        let radix = Radix::new(0);
+        assert!(radix.is_none());
+    }
+
+    #[test]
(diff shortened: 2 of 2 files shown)


lopopolo:
The failing test that this PR replaces was always passing. The construction of the `Radix` was not following the same flow as the Ruby code, which has been corrected.

```console
PS C:\Users\rjl\dev\artichoke\artichoke> cargo run --bin airb
   Compiling artichoke-backend v0.1.0 (C:\Users\rjl\dev\artichoke\artichoke\artichoke-backend)
   Compiling artichoke v0.1.0-pre.0 (C:\Users\rjl\dev\artichoke\artichoke)
    Finished dev [unoptimized + debuginfo] target(s) in 5.08s
     Running `target\debug\airb.exe`
artichoke 0.1.0-pre.0 (2021-08-16 revision 4624) [x86_64-pc-windows-msvc]
[rustc 1.54.0 (a178d0322 2021-07-26) on x86_64-pc-windows-msvc]
>>>
=> nil
>>> Integer("0z", 12000)
        2: from (airb):1
        1: from (eval):92:in Integer
ArgumentError (invalid radix 12000)
>>>
```

PR #1327, opened 2021-08-17
Push a GitHub Actions workflow to reproduce #1325 in CI

lopopolo:
See https://github.com/artichoke/artichoke/issues/1325.

PR #1344, opened 2021-08-21
Mark `UnboxedValueGuard::as_inner_mut` as unsafe

### LINKED ISSUE #1324: Mark `UnboxedValueGuard::as_inner_mut` as unsafe and only impl `DerefMut` if `as_inner_mut` has no safety guarantees that need to be upheld
The safety constraint that the code is failing to surface is that `UnboxedValueGuard::as_inner_mut` is unsafe and `UnboxedValueGuard<'_, Array>` should not implement `DerefMut`.

_Originally posted by @lopopolo in https://github.com/artichoke/artichoke/issues/1323#issuecomment-898815406_

lopopolo:
This PR also removes the `AsMut` and `DerefMut` blanket implementations for `UnboxedValueGuard`. `UnboxedValueGuard` is a proxy object that temporarily reconstructs a Rust object owned by the mruby GC.

Accessing a `&mut T` from `UnboxedValueGuard` is unsafe because accessing `&mut` methods on the underlying `T` may invalidate the raw parts of `T` that are packed into the mruby heap. Callers must ensure these raw parts are valid before the next GC.

Fixes #1324.

This PR also includes safety fixes to box empty arrays into uninitialized MRB_TT_ARRAY before doing init.

This ensures that no uninitialized `MRB_TT_ARRAY` `RBasic *` pointers are seen during a potential intervening GC and mark phase as new values are allocated on the mruby heap.

### DIFF
--- a/artichoke-backend/src/convert/boxing.rs
+++ b/artichoke-backend/src/convert/boxing.rs
@@ -54,9 +54,15 @@ impl<'a, T> UnboxedValueGuard<'a, T> {
     }
 
     /// Get a unique reference to the inner `T`.
+    ///
+    /// # Safety
+    ///
+    /// Callers must ensure the raw parts stored in the source `mrb_value` are
+    /// not invalidated OR that the raw parts are repacked before an mruby
+    /// allocation occurs.
     #[inline]
     #[must_use]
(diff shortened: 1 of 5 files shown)


lopopolo:
This PR extracts 2 commits from #1327.

PR #1349, opened 2021-08-23
Implement `String#inspect`

lopopolo:
Implement `String#inspect` as an iterator similar to `Symbol#inspect`.

This commit bumps `spinoso-string` to v0.1.1.

This iterator does not properly format unprintable ASCII characters as
Unicode escape sequences for UTF-8 `String`s like MRI does.

```console
[3.0.1] > puts "abc\0".b.inspect
"abc\x00"
=> nil
[3.0.1] > puts "abc\0".inspect
"abc\u0000"
=> nil
[3.0.1] > puts "abc\1".b.inspect
"abc\x01"
=> nil
[3.0.1] > puts "abc\1".inspect
"abc\u0001"
=> nil
[3.0.1] > puts "abc\x7F".b.inspect
"abc\x7F"
=> nil
[3.0.1] > puts "abc\x7F".inspect
"abc\u007F"
=> nil
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -586,11 +586,12 @@ dependencies = [
 
 [[package]]
 name = "spinoso-string"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "bstr",
  "focaccia",
  "quickcheck",
+ "scolapasta-string-escape",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


lopopolo:
I filed https://github.com/artichoke/artichoke/issues/1350 to followup on the intentional bug in this implementation so this can be merged sooner.

lopopolo:

PR #1351, opened 2021-08-24
Add is_dead checks to funcall and yield_arg

### LINKED ISSUE #1336: Check is dead before calling `funcall` on values and `yield_arg` on blocks
`Value` and `Block` structs held in Rust may be unreachable and garbage collected by mruby. Such values are "dead" and have `tt = MRB_TT_FREE`.

Calling a function on a dead value or yielding to a dead block is not likely to end well.

Add a check to these functions for dead values and return a `Fatal` error.

`Value::is_dead` is here: https://github.com/artichoke/artichoke/blob/2d446d307258436f4f63ac4ef8cd644c16d9960e/artichoke-backend/src/value.rs#L107-L113

Checks for dead values should be inserted early here and here:

https://github.com/artichoke/artichoke/blob/2d446d307258436f4f63ac4ef8cd644c16d9960e/artichoke-backend/src/value.rs#L139-L149

https://github.com/artichoke/artichoke/blob/2d446d307258436f4f63ac4ef8cd644c16d9960e/artichoke-backend/src/block.rs#L164-L165

stuarth:
Fixes https://github.com/artichoke/artichoke/issues/1336

Is this along the lines you were thinking @lopopolo? If so, thoughts on the error messages? What do you think is a reasonable way to test this?

### DIFF
--- a/artichoke-backend/src/block.rs
+++ b/artichoke-backend/src/block.rs
@@ -162,6 +162,9 @@ impl Block {
     }
 
     pub fn yield_arg(&self, interp: &mut Artichoke, arg: &Value) -> Result<Value, Error> {
+        if arg.is_dead(interp) {
+            return Err(Fatal::from("Arg is dead").into());
+        }
         let result = unsafe { interp.with_ffi_boundary(|mrb| protect::block_yield(mrb, self.inner(), arg.inner()))? };
         match result {
             Ok(value) => {
--- a/artichoke-backend/src/block.rs
+++ b/artichoke-backend/src/block.rs
@@ -209,6 +209,6 @@ mod tests {
         let block = Block::default();
 
         let error = block.yield_arg(&mut interp, &dead).unwrap_err();
-        assert_eq!(error.name().as_ref(), "fatal")
+        assert_eq!(error.name().as_ref(), "fatal");
     }
 }
--- a/artichoke-backend/src/block.rs
+++ b/artichoke-backend/src/block.rs
@@ -163,7 +163,7 @@ impl Block {
 
     pub fn yield_arg(&self, interp: &mut Artichoke, arg: &Value) -> Result<Value, Error> {
         if arg.is_dead(interp) {
-            return Err(Fatal::from("Arg is dead").into());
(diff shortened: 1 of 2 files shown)


artichoke-backend/src/block.rs
```diff
@@ -162,6 +162,9 @@ impl Block {
     }
 
     pub fn yield_arg(&self, interp: &mut Artichoke, arg: &Value) -> Result<Value, Error> {
+        if arg.is_dead(interp) {
+            return Err(Fatal::from("Arg is dead").into());
```
lopopolo:
```suggestion
            return Err(Fatal::from("Value yielded to block is dead. This indicates a bug in the mruby garbage collector. Please leave a comment at https://github.com/artichoke/artichoke/issues/1336.").into());
```

artichoke-backend/src/value.rs
```diff
@@ -143,6 +143,9 @@ impl ValueCore for Value {
         args: &[Self::Arg],
         block: Option<Self::Block>,
     ) -> Result<Self::Value, Self::Error> {
+        if self.is_dead(interp) {
+            return Err(Fatal::from("Value is dead").into());
```
lopopolo:
```suggestion
            return Err(Fatal::from("Value receiver for function call is dead. This indicates a bug in the mruby garbage collector. Please leave a comment at https://github.com/artichoke/artichoke/issues/1336.").into());
```

lopopolo (changes requested):
@stuarth What do you think about these error messages?

I think these diffs need to be rust formatted so it might make sense to not use the GitHub UI to apply them.

lopopolo:
> What do you think is a reasonable way to test this?

I'm hoping these fatals encourage bug reports. I think I've squashed all of the segfaults that might trip these errors but I don't know for sure.

One way to do this might be to look at the tests in `value.rs` which create a dead value and using it for block args and a funcall?

lopopolo:
@stuarth I've assigned this ticket to me so it shows up in my queue but the action is on you to address the review comments and see if we can add a test or two.

lopopolo (approved):
This looks great! Thanks for making those changes and making Artichoke more robust to memory corruptions.

PR #1355, opened 2021-08-28
Move `ClassRegistry` and `ModuleRegistry` trait definitions to `artichoke-core`

### LINKED ISSUE #666: Move ClassRegistry and ModuleRegistry trait definitions to artichoke-core
GH-661 introduces a new `ClassRegistry` and `ModuleRegistry` trait. These traits are defined in `artichoke-backend`. Move them into `artichoke-core`.

stuarth:
Fixes https://github.com/artichoke/artichoke/issues/666

@lopopolo assuming this is along the lines you were hoping, I could use some pointers on producing useful docs if you don't mind

### DIFF
--- a/artichoke-backend/src/class_registry.rs
+++ b/artichoke-backend/src/class_registry.rs
@@ -47,7 +47,7 @@ impl ClassRegistry for Artichoke {
             self.with_ffi_boundary(|mrb| {
                 if let Some(mut rclass) = rclass.resolve(mrb) {
                     let value_class = sys::mrb_sys_class_value(rclass.as_mut());
-                    Some(Self::Value::from(value_class))
+                    Some(Value::from(value_class))
                 } else {
                     None
                 }
@@ -67,7 +67,7 @@ impl ClassRegistry for Artichoke {
         } else {
             return Ok(None);
         };
-        let args = args.iter().map(Self::Value::inner).collect::<Vec<_>>();
+        let args = args.iter().map(Value::inner).collect::<Vec<_>>();
         let arglen = if let Ok(len) = sys::mrb_int::try_from(args.len()) {
             len
         } else {
@@ -77,7 +77,7 @@ impl ClassRegistry for Artichoke {
             self.with_ffi_boundary(|mrb| {
                 if let Some(mut rclass) = rclass.resolve(mrb) {
                     let value = sys::mrb_obj_new(mrb, rclass.as_mut(), arglen, args.as_ptr());
-                    Some(Self::Value::from(value))
+                    Some(Value::from(value))
                 } else {
                     None
                 }
--- a/artichoke-backend/src/class_registry.rs
+++ b/artichoke-backend/src/class_registry.rs
@@ -1,9 +1,8 @@
 use std::any::Any;
 use std::convert::TryFrom;
 
-use artichoke_core::class_registry::ClassRegistry;
-
 use crate::class;
+use crate::core::ClassRegistry;
 use crate::error::Error;
 use crate::ffi::InterpreterExtractError;
 use crate::sys;
--- a/artichoke-backend/src/class_registry.rs
+++ b/artichoke-backend/src/class_registry.rs
@@ -48,7 +48,7 @@ impl ClassRegistry for Artichoke {
             self.with_ffi_boundary(|mrb| {
                 if let Some(mut rclass) = rclass.resolve(mrb) {
                     let value_class = sys::mrb_sys_class_value(rclass.as_mut());
-                    Some(Value::from(value_class))
+                    Some(Self::Value::from(value_class))
                 } else {
                     None
                 }
@@ -57,7 +57,7 @@ impl ClassRegistry for Artichoke {
         Ok(value_class)
     }
 
-    fn new_instance<T>(&mut self, args: &[Value]) -> Result<Option<Self::Value>, Self::Error>
+    fn new_instance<T>(&mut self, args: &[Self::Value]) -> Result<Option<Self::Value>, Self::Error>
     where
         T: Any,
     {
@@ -68,7 +68,7 @@ impl ClassRegistry for Artichoke {
         } else {
             return Ok(None);
         };
-        let args = args.iter().map(Value::inner).collect::<Vec<_>>();
+        let args = args.iter().map(Self::Value::inner).collect::<Vec<_>>();
         let arglen = if let Ok(len) = sys::mrb_int::try_from(args.len()) {
             len
         } else {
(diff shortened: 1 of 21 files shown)


lopopolo:
Thanks for picking this up @stuarth.

I'm `ack`ing that I've seen this PR, but it might be a couple of days before I can give feedback on it.

artichoke-core/src/class_registry.rs
```diff
@@ -0,0 +1,40 @@
+use std::any::Any;
```
lopopolo:
```suggestion
use core::any::Any;
```

`artichoke-core` can be built without `std` and `Any` is available in `core`, so let's import it from there.

artichoke-core/src/class_registry.rs
```diff
@@ -0,0 +1,40 @@
...
+
+pub trait ClassRegistry {
+    type Value;
+    type Error;
+    type Spec;
```
lopopolo: What do you think of these docs?

```suggestion
    /// Concrete value type for boxed Ruby values.
    type Value;

    /// Concrete error type for errors encountered when manipulating the class registry.
    type Error;

    /// Type representing a class specification.
    type Spec;
```

artichoke-core/src/class_registry.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// Create a class definition bound to a Rust type `T`.
+    ///
+    /// Class definitions have the same lifetime as the
+    /// [`State`](crate::state::State) because the class def owns the
+    /// `mrb_data_type` for the type, which must be long-lived.
```
lopopolo: These docs belong in `artichoke-backend` since they mention mruby bits.

artichoke-core/src/module_registry.rs
```diff
@@ -0,0 +1,41 @@
+use std::any::Any;
```
lopopolo:
```suggestion
use core::any::Any;
```

artichoke-core/src/module_registry.rs
```diff
@@ -0,0 +1,41 @@
...
+
+pub trait ModuleRegistry {
+    /// Module value type
+    type Value;
+
+    /// Module error type
+    type Error;
+
+    /// Module specification type
+    type Spec;
```
lopopolo: What do you think of these docs?

```suggestion
    /// Concrete value type for boxed Ruby values.
    type Value;

    /// Concrete error type for errors encountered when manipulating the module registry.
    type Error;

    /// Type representing a module specification.
    type Spec;
```

artichoke-core/src/module_registry.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// Create a module definition bound to a Rust type `T`.
+    ///
+    /// Module definitions have the same lifetime as the interpreter because the
+    /// module def owns the `mrb_data_type` for the type, which must be
+    /// long-lived.
```
lopopolo: these docs belong in `artichoke-backend` since they mention mruby bits.

lopopolo (changes requested):
This looks like it has the right direction. Thank you.

Can you make the fixups I suggested and take a stab at some docs? Once you think you're ready, please move the PR out of draft and I'll maybe push a commit to tweak the docs.

Does that sound good @stuarth?

artichoke-backend/src/class_registry.rs
```diff
@@ -76,7 +47,7 @@ impl ClassRegistry for Artichoke {
             self.with_ffi_boundary(|mrb| {
                 if let Some(mut rclass) = rclass.resolve(mrb) {
                     let value_class = sys::mrb_sys_class_value(rclass.as_mut());
-                    Some(Value::from(value_class))
+                    Some(Self::Value::from(value_class))
```
lopopolo: I like using `Self::Value` in the function signatures to match the trait definition, but in the function body, I'd prefer if we use the concrete type instead of indirecting through the trait.

```suggestion
                    Some(Value::from(value_class))
```

artichoke-backend/src/class_registry.rs
```diff
@@ -96,7 +67,7 @@ impl ClassRegistry for Artichoke {
         } else {
             return Ok(None);
         };
-        let args = args.iter().map(Value::inner).collect::<Vec<_>>();
+        let args = args.iter().map(Self::Value::inner).collect::<Vec<_>>();
```
lopopolo: I like using `Self::Value` in the function signatures to match the trait definition, but in the function body, I'd prefer if we use the concrete type instead of indirecting through the trait.

```suggestion
        let args = args.iter().map(Value::inner).collect::<Vec<_>>();
```

artichoke-backend/src/class_registry.rs
```diff
@@ -106,7 +77,7 @@ impl ClassRegistry for Artichoke {
             self.with_ffi_boundary(|mrb| {
                 if let Some(mut rclass) = rclass.resolve(mrb) {
                     let value = sys::mrb_obj_new(mrb, rclass.as_mut(), arglen, args.as_ptr());
-                    Some(Value::from(value))
+                    Some(Self::Value::from(value))
```
lopopolo: I like using `Self::Value` in the function signatures to match the trait definition, but in the function body, I'd prefer if we use the concrete type instead of indirecting through the trait.

```suggestion
                    Some(Value::from(value))
```

artichoke-backend/src/module_registry.rs
```diff
@@ -71,7 +46,7 @@ impl ModuleRegistry for Artichoke {
         let rclass = unsafe { self.with_ffi_boundary(|mrb| rclass.resolve(mrb))? };
         if let Some(mut rclass) = rclass {
             let module = unsafe { sys::mrb_sys_module_value(rclass.as_mut()) };
-            let module = Value::from(module);
+            let module = Self::Value::from(module);
```
lopopolo: I like using `Self::Value` in the function signatures to match the trait definition, but in the function body, I'd prefer if we use the concrete type instead of indirecting through the trait.

```suggestion
            let module = Value::from(module);
```

lopopolo (changes requested):

lopopolo:
@stuarth I added some docs. Let me know what you think.

I think this is ready to merge if you want to move it out of Draft state.

stuarth:
> @stuarth I added some docs. Let me know what you think.
> 
> I think this is ready to merge if you want to move it out of Draft state.

Thanks for picking it up! Been traveling and haven't had a chance to follow up

lopopolo (approved):

Issue #302, opened 2019-10-14
Add SIGINT handler in artichoke-frontend 


lopopolo:
Crate `signal_hook`.

In `airb`, CTRL-C should wipe the current input and write a new empty prompt line.

lopopolo:
From GH-337:

[ctrlc](https://crates.io/crates/ctrlc): Interrupt handling.
[signalhook](https://github.com/vorner/signal-hook): Interrupt handling:

```rust
extern crate crossbeam_channel as channel;
extern crate signal_hook;

fn notify(signals: &[c_int]) -> Result<channel::Receiver<c_int>> {
    let (s, r) = channel::bounded(100);
    let signals = signal_hook::iterator::Signals::new(signals)?;
    thread::spawn(move || {
        for signal in signals.forever() {
            s.send(signal);
        }
    });
    Ok(r)
}
```

lopopolo:
This was fixed a while ago in https://github.com/artichoke/artichoke/commit/08d4b502dfe219df24fb207045075f17e972f70a (it looks like before this issue was filed?)

I'm not sure if the intent of this issue was to support interrupting the VM arbitrarily. Closing.

PR #1356, opened 2021-08-29
Mirror irb behavior on empty inputs

stuarth:
While checking https://github.com/artichoke/artichoke/issues/302 (which, at a quick glance, looks already resolved!) , I noticed that `airb`'s handling of empty input differs from `irb`'s. This PR aligns them. Specifically, the issue is that on empty input, `airb` evaluates, prints nil, and stores history, while `irb` skips it

### DIFF
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -234,37 +234,35 @@ where
 
         let readline = rl.readline(prompt);
         match readline {
+            Ok(line) if line.is_empty() => (),
             Ok(line) => {
-                if !line.is_empty() {
-                    buf.push_str(line.as_str());
-                    parser_state = parser.parse(buf.as_bytes());
-                    if parser_state.is_fatal() {
-                        return Err(Box::new(ParserInternalError::new()));
-                    }
-                    if parser_state.is_code_block_open() {
-                        buf.push('\n');
-                        continue;
-                    }
-                    if parser_state.is_recoverable_error() {
-                        writeln!(error, "Could not parse input")?;
-                        buf.clear();
-                        continue;
-                    }
-                    match interp.eval(buf.as_bytes()) {
-                        Ok(value) => {
-                            let result = value.inspect(interp);
-                            output.write_all(config.result_prefix.as_bytes())?;
-                            output.write_all(result.as_slice())?;
-                            output.write_all(b"\n")?;
-                        }
-                        Err(ref exc) => backtrace::format_repl_trace_into(&mut error, interp, exc)?,
-                    }
-                    for line in buf.lines() {
-                        rl.add_history_entry(line);
-                        interp.add_fetch_lineno(1).map_err(|_| ParserLineCountError::new())?;
-                    }
+                buf.push_str(line.as_str());
+                parser_state = parser.parse(buf.as_bytes());
+                if parser_state.is_fatal() {
+                    return Err(Box::new(ParserInternalError::new()));
                 }
-
+                if parser_state.is_code_block_open() {
+                    buf.push('\n');
+                    continue;
+                }
+                if parser_state.is_recoverable_error() {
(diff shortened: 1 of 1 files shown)


lopopolo (changes requested):
Wow nice catch. Can we rework this diff to skip the evaluation loop using a match guard? I think it will make the diff much cleaner.

Rather than wrap the entire Ok branch in an if, I think we can get the desired behavior by adding this match arm first:

```rust
Ok(line) if line.is_empty() => {}
```

lopopolo:
What does `irb` do if the line is only white space?

stuarth:
> Wow nice catch. Can we rework this diff to skip the evaluation loop using a match guard? I think it will make the diff much cleaner.
> 
> Rather than wrap the entire Ok branch in an if, I think we can get the desired behavior by adding this match arm first:
> 
> ```rust
> Ok(line) if line.is_empty() => {}
> ```

Yep, that's fair! I had written it that way so

```
interp.incremental_gc()?;
buf.clear();
```

would still run, but no reason I can see we can't skip those in this case.

> What does irb do if the line is only white space?

Depends -- for just a newline, it'll behave as we do now and skip evaluation / storing in history. Given some whitespace it'll treat it as a typical expression and eval + store

lopopolo:
> Yep, that's fair! I had written it that way so
> 
> ```
> interp.incremental_gc()?;
> buf.clear();
> ```
> 
> would still run, but no reason I can see we can't skip those in this case.

Yea skipping that seems fine to me. 

> > What does irb do if the line is only white space?
> 
> Depends -- for just a newline, it'll behave as we do now and skip evaluation / storing in history. Given some whitespace it'll treat it as a typical expression and eval + store

Awesome thanks for digging into that.

src/repl.rs
```diff
@@ -234,6 +234,7 @@ where
 
         let readline = rl.readline(prompt);
         match readline {
+            Ok(line) if line.is_empty() => (),
```
lopopolo: Oo I think this changes behavior if the `buf` is non-empty and the REPL is in multi-line editing mode.

Does this look better to you?

```suggestion
            Ok(line) if line.is_empty() && buf.is_empty() => (),
```

One way to test this is to try to put this into the REPL and see if the newline is preserved:

```ruby
def foo(x)
  return 1 if x.nil?

  2
end
```

lopopolo (commented):
@stuarth what do you think about this change for multi-line editing mode?

src/repl.rs
stuarth: OK, so that gets to another issue -- multiline is not behaving at all for me on `trunk`. I'd actually assumed it was unimplemented. 

```sh
~/d/artichoke (trunk)> target/debug/airb
artichoke 0.1.0-pre.0 (2021-08-24 revision 4673) [x86_64-apple-darwin]
[rustc 1.54.0 (a178d0322 2021-07-26) on x86_64-apple-darwin]
>>> def foo(x)
```

lopopolo: Yikes. This was never fully working but it looks like it has regressed. My guess is that broke with the mruby 3 upgrade.

Please make the change to only skip blank lines when buf is empty. We can make a ticket for the broken multiline input mode to fix separately.

@stuarth Would you be able to cut the issue for the broken multiline input mode?

stuarth:
> @stuarth Would you be able to cut the issue for the broken multiline input mode?

Of course! Will get that posted in a few minutes

lopopolo (approved):
Looks good. Thank you!

lopopolo:
> > @stuarth Would you be able to cut the issue for the broken multiline input mode?
> 
> Of course! Will get that posted in a few minutes

This was filed here: https://github.com/artichoke/artichoke/issues/1358.

PR #1222, opened 2021-05-30
Implement `String` in Rust

lopopolo:
**Implement the `String` class from Ruby core in Rust with `spinoso-string` in `artichoke-backend`.** 💎 🎉 💎 🎉 💎

[All 144 APIs available on `String`][ruby-core-string] in MRI 3.0.2 are either implemented or stubbed to raise `NotImplementedError`.

[ruby-core-string]: https://ruby-doc.org/core-3.0.2/String.html

## Design

Artichoke's `String` is based on [`spinoso_string::String`][spinoso-string-doc]. `String` is a byte buffer and associated encoding. `String` supports UTF-8, ASCII, and binary encodings. Encodings are conventional (a UTF-8 `String` may contain invalid UTF-8 byte sequences, an ASCII `String` may contain bytes greater than `0x7F`).

Character oriented APIs like `String::center`, `String::get_char` and `String::char_len` are implemented according to the `String`'s conventional encoding.

Unlike `String` in MRI, Artichoke's `String` does not cache whether a `String`'s encoding is valid, nor does it cache whether the `String` is `ascii_only`. These enhancements are expected to come in a future set of changes. I've [discussed how I would like to restructure `String`'s internals][twitter-string-nirvana] to make this happen on Twitter

`spinoso-string` uses state of the art Rust dependencies for encoding-oriented operations with first-class SIMD implementations. `spinoso-string` is built on top of `bstr` (ASCII and UTF-8 decoding, `find`-oriented APIs), `simdutf8` (UTF-8 validity checks), and `bytecount` (UTF-8 character length).

[spinoso-string-doc]: https://artichoke.github.io/artichoke/spinoso_string/struct.String.html
[twitter-string-nirvana]: https://twitter.com/artichokeruby/status/1464117285002440707

### mruby FFI Compatibility

Much like earlier work on `Array`, the `MRB_API` functions for strings are used heavily in the VM in addition to being exposed to Ruby code. The mruby VM treats strings as memory managed byte buffers and uses them extensively in its internals.

It is not sufficient to `#ifdef` out the C API since it is used outside of the `string.c` compilation unit.

Artichoke includes several reimplementations of mruby C APIs, mostly in Rust and some in the `mruby-sys` C extension.

During the development of these FFI compat functions, I encountered 1 segfault and 1 buffer overrun, both documented in the commits that address them:

- 7a13a30cf7d232fe9f7aefa8ab1d8f26073f9fd1
- 279c698b11ba885843e86561301f4951ecd16641

[...]

PR #1357, opened 2021-08-29
Implement methods in `spinoso-string` to enable packing and unpacking `String`s from Ruby values

lopopolo:
## Add "raw parts" APIs to spinoso-string

Add `String::into_raw_parts` and `String::from_raw_parts`. These APIs
enable deconstructing a `String` into a (pointer, length, capacity)
tuple so they can be packed into a `*mut sys::RString` when implementing
`BoxUnboxVmValue` in `artichoke-backend`.

## Add String::set_encoding method to spinoso-string

This method enables implementing `String#force_encoding` and friends.
This method also enables properly setting the encoding after
reconstructing a `String` from "raw parts".

## Notes

These commits were extracted from https://github.com/artichoke/artichoke/pull/1222.

### DIFF
--- a/spinoso-string/src/lib.rs
+++ b/spinoso-string/src/lib.rs
@@ -712,6 +712,23 @@ impl String {
         self.encoding
     }
 
+    /// Set the [`Encoding`] of this `String`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use spinoso_string::{Encoding, String};
+    ///
+    /// let mut s = String::utf8(b"xyz".to_vec());
+    /// assert_eq!(s.encoding(), Encoding::Utf8);
+    /// s.set_encoding(Encoding::Binary);
+    /// assert_eq!(s.encoding(), Encoding::Binary);
+    /// ```
+    #[inline]
+    pub fn set_encoding(&mut self, encoding: Encoding) {
+        self.encoding = encoding;
+    }
+
     /// Extracts a slice containing the entire byte string.
     ///
     /// Equivalent to `&s[..]`.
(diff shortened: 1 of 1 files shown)


lopopolo:
I meant to include this patch in this PR, but I don't think it's worth splitting out of #1222 on its own:

```patch
commit b348504a364d94b33d0cb608f057dc2c8158f1d5
Author: Ryan Lopopolo <rjl@hyperbo.la>
Date:   Sun Aug 29 06:08:11 2021 -0700

    Fix a typo in `Array::from_raw_parts` doc comment

diff --git a/spinoso-array/src/array/vec/mod.rs b/spinoso-array/src/array/vec/mod.rs
index e33cc6b91b..e9db1b4a1b 100644
--- a/spinoso-array/src/array/vec/mod.rs
+++ b/spinoso-array/src/array/vec/mod.rs
@@ -285,7 +285,7 @@ impl<T> Array<T> {
         self.0.set_len(new_len);
     }

-    /// Creates a `Array<T>` directly from the raw components of another array.
+    /// Creates an `Array<T>` directly from the raw components of another array.
     ///
     /// # Safety
     ///
@@ -333,6 +333,8 @@ impl<T> Array<T> {
     pub fn into_raw_parts(self) -> (*mut T, usize, usize) {
         // TODO: convert to `Vec::into_raw_parts` once it is stabilized.
         // See: https://doc.rust-lang.org/1.48.0/src/alloc/vec.rs.html#399-402
+        //
+        // https://github.com/rust-lang/rust/issues/65816
         let mut me = ManuallyDrop::new(self.0);
         (me.as_mut_ptr(), me.len(), me.capacity())
     }
```

lopopolo:

PR #1360, opened 2021-09-01
repl fix: check for open code block first

### LINKED ISSUE #1358: `airb` does not handle multiline input
`airb` does not currently handle multiline inputs correctly when it's entered line by line. To reproduce

```
>>> def foo
```

The same, in `irb`
```
irb(main):001:0> def foo
irb(main):002:1>
```

Note that pasting a multiline block _does_ work in `airb`
```
>>> def foo
  2
end
=> :foo
```

stuarth:
Reorders checks in the repl to look at whether a code block is open _before_ checking whether it has fatal parsing errors

Tracks behavior in https://github.com/artichoke/artichoke/blob/858b686ed93fa3b23b4e65d0dc64d1ea6e591cb6/artichoke-backend/vendor/mruby/mrbgems/mruby-bin-mirb/tools/mirb/mirb.c#L634-L648

Fixes https://github.com/artichoke/artichoke/issues/1358

### DIFF
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -237,12 +237,13 @@ where
             Ok(line) => {
                 buf.push_str(line.as_str());
                 parser_state = parser.parse(buf.as_bytes());
-                if parser_state.is_fatal() {
-                    return Err(Box::new(ParserInternalError::new()));
-                }
+
                 if parser_state.is_code_block_open() {
                     buf.push('\n');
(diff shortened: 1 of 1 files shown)


lopopolo (approved):
Thanks @stuarth this looks right to me. Thanks for following up on this breakage you observed in the other ticket.

PR #1402, opened 2021-09-06
Keep `spinoso-string` monomorphization instantiations in check to minimize code bloat

lopopolo:
Inspired by: https://matklad.github.io/2021/09/04/fast-rust-builds.html#keeping-instantiations-in-check


lopopolo:

PR #1403, opened 2021-09-06
Fix `String::chr` when string begins with an incomplete UTF-8 sequence

lopopolo:
When a `String` begins with a prefix that could be a well-formed UTF-8
code point, but isn't, `bstr` will return the full multi-byte sequence
when calling `bstr::decode_utf8` according to the substitution of
maximal subparts strategy.

MRI does not follow this strategy and will always return only the first
byte of an invalid UTF-8 prefix for conventionally UTF-8 strings.


lopopolo:

PR #1409, opened 2021-09-11
Convert parser error states to enum to force exhaustive handling

lopopolo:
Follow up to #1360 by converting the if chains to a match statement with
an enum for exhaustive and non-overlapping branching in parser error
handling.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -10,6 +10,28 @@ use std::ptr::NonNull;
 use crate::backend::sys;
 use crate::backend::Artichoke;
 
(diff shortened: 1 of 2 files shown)


lopopolo:
this PR is not a valid approach and regresses on multi-line editing:

```
artichoke 0.1.0-pre.0 (2021-09-10 revision 4788) [x86_64-apple-darwin]
[rustc 1.55.0 (c8dfcfe04 2021-09-06) on x86_64-apple-darwin]
>>> def foo
```

PR #1414, opened 2021-09-18
[WIP] Implement `StringScanner` in Rust

### LINKED ISSUE #135: Implement strscan partially in Rust
Artichoke core implements `strscan` in pure Ruby and does so in terms of the `charpos` pointer. This means that Artichoke behaves incorrectly when combining `get_byte` with non-ASCII `String`s.

Implement `strscan` backed by a `Vec<u8>`.

The output of `StringScanner#inspect` is telling in that it treats the non-ascii `String` as binary content.

It is not necessary to implement all of `strscan` in Rust, just the bits necessary to support `get_byte` + Unicode.

We should write a spec to validate the fix and contribute it back to [ruby/spec](https://github.com/ruby/spec).

## Artichoke

```console
$ cargo run --bin airb
mruby 2.0 [2.0.1-12]
>>> require 'strscan'
=> true
>>> s = StringScanner.new('💎')
=> #<StringScanner 0/1 @ "💎...">
>>> s.get_byte
=> "\xf0"
>>> s.rest
=> ""
```

## MRI

```console
$ irb
[2.6.3] > require 'strscan'
=> true
[2.6.3] > s = StringScanner.new('💎')
=> #<StringScanner 0/4 @ "\xF0\x9F\x92\x8E">
[2.6.3] > s.get_byte
=> "\xF0"
[2.6.3] > s.rest
=> "\x9F\x92\x8E"
```

lopopolo:
Fixes https://github.com/artichoke/artichoke/issues/135.

### DIFF
--- a/spinoso-strscan/src/lib.rs
+++ b/spinoso-strscan/src/lib.rs
@@ -1,0 +1,198 @@
+use core::fmt;
+use std::error::Error;
+
+#[derive(Default, Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct RangeError {
+    message: &'static str,
+}
+
+impl RangeError {
+    pub const fn new() -> Self {
+        Self::with_message("RangeError")
+    }
+
+    pub const fn with_message(message: &'static str) -> Self {
+        Self { message }
+    }
+
+    pub fn message(self) -> &'static str {
+        self.message
+    }
+}
+
+impl fmt::Display for RangeError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.message)
+    }
+}
+
+impl Error for RangeError {}
+
+#[derive(Default, Debug, Clone, Hash)]
+pub struct StringScanner {
+    string: Vec<u8>,
+    scan_pos: usize,
+    previous_scan_pos: Option<usize>,
+    last_match: Option<(usize, usize)>,
+}
+
+impl From<String> for StringScanner {
+    fn from(string: String) -> Self {
+        Self::new(string.into_bytes())
+    }
+}
+
+impl From<Vec<u8>> for StringScanner {
+    fn from(string: Vec<u8>) -> Self {
+        Self::new(string)
+    }
+}
+
+impl StringScanner {
+    pub const fn new(string: Vec<u8>) -> Self {
+        Self {
+            string,
+            scan_pos: 0,
+            previous_scan_pos: None,
+            last_match: None,
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


spinoso-strscan/src/lib.rs
```diff
@@ -0,0 +1,198 @@
+use core::fmt;
+use std::error::Error;
```
lopopolo, line 2: the dep on `std` can probably be optional if we add an `extern crate alloc`.

spinoso-strscan/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+impl RangeError {
+    pub const fn new() -> Self {
+        Self::with_message("RangeError")
+    }
+
+    pub const fn with_message(message: &'static str) -> Self {
+        Self { message }
+    }
+
+    pub fn message(self) -> &'static str {
```
lopopolo, line 18: this function can be const.

spinoso-strscan/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        Some(byte)
+    }
+
+    pub fn get_ch(&mut self) -> Option<&[u8]> {
+        let remaining = &self.string[self.scan_pos..];
+        let (ch, size) = bstr::decode_utf8(remaining);
+        let previous_scan_pos = self.scan_pos;
+        self.scan_pos += size;
+        self.previous_scan_pos = Some(previous_scan_pos);
+        self.last_match = Some((previous_scan_pos, self.scan_pos));
```
lopopolo, line 114: all of this can be short circuited if size is 0.

spinoso-strscan/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        if ch.is_some() {
+            Some(&remaining[..size])
+        } else {
+            None
+        }
```
lopopolo, line 119: this is incorrect. if `ch` is none, we might have decoded an invalid UTF-8 sequence which means we should return `remaining.get(..1)`.

spinoso-strscan/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    pub fn peek(&self, len: usize) -> &[u8] {
+        self.string
+            .get(self.scan_pos..self.scan_pos.saturating_add(len))
+            .or_else(|| self.string.get(self.scan_pos..))
+            .unwrap_or_default()
+    }
+
+    pub fn pos(&self) -> usize {
```
lopopolo, line 143: this can be const

lopopolo:
This PR assumes UTF-8 string encoding and is obsoleted by `spinoso-string` and #1630.

PR #1416, opened 2021-09-18
[WIP] Add function to implicitly convert to `Array`

### LINKED ISSUE #310: Implement implicit conversions to Array, Integer, and String on Value

lopopolo:
Fixes https://github.com/artichoke/artichoke/issues/310.

### DIFF
--- a/artichoke-backend/src/convert/implicit.rs
+++ b/artichoke-backend/src/convert/implicit.rs
@@ -1,11 +1,71 @@
+use core::mem;
+
 use artichoke_core::debug::Debug as _;
 use artichoke_core::value::Value as _;
 use spinoso_exception::TypeError;
 
+use crate::convert::{BoxUnboxVmValue, UnboxedValueGuard};
 use crate::error::Error;
+use crate::extn::core::array::Array;
 use crate::types::Ruby;
 use crate::value::Value;
 use crate::Artichoke;
+
+pub unsafe fn implicitly_convert_to_array<'a>(
+    interp: &mut Artichoke,
+    value: &'a mut Value,
+) -> Result<UnboxedValueGuard<'a, Array>, TypeError> {
+    // This is a hack to work around double borrowing.
+    let val = *value;
+    if let Ok(array) = Array::unbox_from_value(value, interp) {
+        return Ok(array);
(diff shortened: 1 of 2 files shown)


lopopolo:
This has been on a local branch for 1.5 years. I rebased to latest trunk the best I could. Pushing it up so I don't lose it.

artichoke-backend/src/convert/implicit.rs
```diff
@@ -1,12 +1,72 @@
...
 use crate::types::Ruby;
 use crate::value::Value;
 use crate::Artichoke;
 
+pub unsafe fn implicitly_convert_to_array<'a>(
+    interp: &mut Artichoke,
+    value: &'a mut Value,
+) -> Result<UnboxedValueGuard<'a, Array>, TypeError> {
+    // This is a hack to work around double borrowing.
+    let val = *value;
```
lopopolo, line 19: big yikes.

lopopolo:
Obsoleted by https://github.com/artichoke/artichoke/pull/2121.

PR #1346, opened 2021-08-21
Add a GitHub Actions workflow that runs the spec-runner under Pernosco

lopopolo:
This PR branches from 4321b467860419e0718bdac25e8683a9fdc8cf50 which was known to cause a core dump in #1327.

This PR includes an improved version of the `crash.yaml` workflow which splits apart the invocations of the spec runner to one per job step and enables recording execution with Pernosco.

cc @rocallahan @khuey

Opening this PR will kick off a run of the crash workflow which I expect to produce a core dump. Once I confirm the Pernosco GitHub Actions integration is working correctly, I'll rebase to the latest trunk and merge this PR.

PR #1419, opened 2021-09-18
Push a GitHub Actions workflow to trigger crashers with Pernosco

lopopolo:
Now that the Pernosco integration is known to be working as tested in the PR #1346, merge the workflow so we can generate reproduction environments for future crashes.

Pernosco is enabled via environment variable on the runs of the `spec-runner`.

This commit cherry-picks and squashes a couple of commits from #1346:

- e1eea7148076e8a9329a076bb8d0a39b4332dc93
- 91bc825269530a37fff960d448127f0a843f1cee

Closes #1346.

### DIFF
--- a/.github/workflows/crash.yaml
+++ b/.github/workflows/crash.yaml
@@ -1,0 +1,80 @@
+---
+name: Crash
+"on":
+  pull_request:
+    branches:
+      - trunk
+jobs:
+  crash:
+    name: Crash
+    runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: -D warnings
+      RUST_BACKTRACE: 1
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


lopopolo:
This workflow is only triggered on PRs. It is NOT triggered on merges to `trunk`.

lopopolo:
Linking the docs: https://github.com/Pernosco/pernosco/wiki/Using-Pernosco-To-Debug-Github-Actions-Test-Failures

I'm not sure if the `PERNOSCO_ENABLED` env var needs to be enabled on the build step or the run step (the docs do it on `cargo test` which does both).

Enabling it on the run steps worked in #1346 though, so leaving it that way.

lopopolo:
cc @rocallahan @khuey

I've merged the Pernosco workflow. I'll spend some time on a blog post this weekend.

PR #1420, opened 2021-09-20
Add `license-file` key to Cargo.toml for all crates

lopopolo:


lopopolo:
Filed https://github.com/EmbarkStudios/cargo-about/issues/162 upstream to see if `cargo-about` can handle path dependencies.

froydnj:
Is the concern here fixing the `cargo-about` output, or that artichoke crates have their own slightly different MIT-like license?

lopopolo:
The concern is about fixing the `cargo-about` output. `cargo-about` outputs a generic MIT license without my copyright information included.

Moving the license file into the sub crates in the workspace causes `cargo-about` to correctly pick the license text with my copyright.

lopopolo:
Filed https://github.com/sstadick/cargo-bundle-licenses/issues/2 upstream as well.

lopopolo:
@froydnj It appears the template MIT license is used when `cargo-about` either can't find or can't parse the license file, using the chrono crates as a reference:

- https://github.com/chronotope/chrono/blob/3467172c31188006147585f6ed3727629d642fed/LICENSE.txt
- https://github.com/chronotope/chrono-tz/blob/90ed30d5fe42e8b2d80c10ee9d5394e9e3b0f528/LICENSE

lopopolo:
grr it looks like `cargo` doesn't like it if both of these keys are set:

```console
[09:40] [artichoke/scolapasta-string-escape] $ cargo package
   Packaging scolapasta-string-escape v0.2.0 (/Users/lopopolo/dev/artichoke/artichoke/scolapasta-string-escape)
   Verifying scolapasta-string-escape v0.2.0 (/Users/lopopolo/dev/artichoke/artichoke/scolapasta-string-escape)
warning: only one of `license` or `license-file` is necessary
    Updating crates.io index
   Compiling memchr v2.4.1
   Compiling bstr v0.2.16
   Compiling scolapasta-string-escape v0.2.0 (/Users/lopopolo/dev/artichoke/artichoke/target/package/scolapasta-string-escape-0.2.0)
    Finished dev [unoptimized + debuginfo] target(s) in 3.20s
```

PR #1450, opened 2021-10-01
Add a 'perl hash' implementation to `spinoso-string`

lopopolo:
This hash routine is used in MRI Ruby 1.8.7 and mruby 3.0.0.

https://github.com/artichoke/mruby/blob/3.0.0/src/string.c#L1751-L1769
https://github.com/ruby/ruby/blob/v1_8_7/string.c#L869-L903

And the hash routine from Perl 5.8.0 on which the above are based: https://github.com/Perl/perl5/blob/perl-5.8.0/hv.h#L49-L73

This routine is not a cryptographic hash and is not collision-resistant.

This PR was extracted from https://github.com/artichoke/artichoke/pull/1222.

PR #1488, opened 2021-11-06
Add global hasher

### LINKED ISSUE #1404: Add an interpreter-global hasher
To facilitate APIs like `String#hash`, it would help to dangle a [`BuildHasher`] like [`RandomState`] off of the interpreter and expose it with a trait in `artichoke-core`. 

[`BuildHasher`]: https://doc.rust-lang.org/std/hash/trait.BuildHasher.html
[`RandomState`]: https://doc.rust-lang.org/std/collections/hash_map/struct.RandomState.html

I imagine a trait like:

```rust
// This trait is identical to `BuildHasher`.
pub trait Hash: core::hash::BuildHasher {}
```

To implement this trait in `artichoke-backend`, add a `RandomState` to the interpreter state:

https://github.com/artichoke/artichoke/blob/fdc7bb6ef8d2023abaa672040efe902f0ccea030/artichoke-backend/src/state/mod.rs#L24-L35

and implement the trait in a source file at `artichoke-backend/src/hash.rs`, like this example for `Warn`:

https://github.com/artichoke/artichoke/blob/fdc7bb6ef8d2023abaa672040efe902f0ccea030/artichoke-backend/src/warn.rs#L14-L41

This blocks #1222.

masonforest:
> 😄 you might want to take a look at:
>Add a 'perl hash' implementation to spinoso-string #1450
>https://twitter.com/artichokeruby/status/1444089207547330562

Awesome! It's like we're Indiana Jonesing hashing algorithms haha. Let me know if this needs anything else!

Closes #1404


artichoke-core/src/hash.rs
```diff
@@ -0,0 +1,9 @@
+//! Build hashers and hash values
+
+/// A trait for creating instances of RandomState
+pub trait BuildHasher<B: std::hash::BuildHasher> {
+    /// Concrete error type for errors encountered when outputting hash errors.
+    type Error;
```
lopopolo: oh sorry I may have led you down a confusing path here.

Adding a generic parameter to the trait means that `BuildHasher` can be implemented multiple times for a given interpreter with multiple `BuildHasher` type params. This isn't quite the API I was hoping for.

Instead, like we have the associated `Error` type here, implementors of `BuildHasher` should declare exactly 1 `BuildHasher` type they use.

Something like this:

```suggestion
pub trait BuildHasher {
    /// Concrete error type for errors encountered when outputting hash errors.
    type Error;
    
    /// Concrete build hasher type.
    type BuildHasher: core::hash::BuildHasher;
```

This also uses the `BuildHasher` import via `core` so `artichoke-core` continues to be compilable with no-std.

This structure with associated types is used pretty heavily in `artichoke-core`, for example the PRNG in the `Prng` trait:

https://github.com/artichoke/artichoke/blob/1c1702894e6ef237b8d7644c16f6b1bc9ae5f5e7/artichoke-core/src/prng.rs#L10-L11

artichoke-backend/src/hash.rs
```diff
@@ -0,0 +1,14 @@
+use crate::core::BuildHasher;
+use crate::error::Error;
+use std::collections::hash_map::RandomState;
+use crate::ffi::InterpreterExtractError;
+use crate::Artichoke;
+
+impl BuildHasher<RandomState> for Artichoke {
+    type Error = Error;
```
lopopolo: To match the change proposed below:

```suggestion
impl BuildHasher for Artichoke {
    type Error = Error;
    type BuildHasher = RandomState;
```

artichoke-backend/src/hash.rs
```diff
@@ -0,0 +1,14 @@
...
+
+    fn build_hasher(&mut self) -> Result<&RandomState, Self::Error> {
```
lopopolo: style nit, I prefer using the associated types here to mirror the trait definition:

```suggestion
    fn build_hasher(&mut self) -> Result<&Self::BuildHasher, Self::Error> {
```

artichoke-backend/src/lib.rs
```diff
@@ -126,6 +126,7 @@ mod eval;
 pub mod exception_handler;
 pub mod extn;
 pub mod ffi;
+pub mod hash;
```
lopopolo: there's nothing here to export, there's only a trait implementation, so let's not make this `pub` like `globals` and `intern` below.

```suggestion
mod hash;
```

lopopolo (changes requested):
Thanks for the PR @masonforest.

It looks like I communicated poorly around the associated type parameter for the build hasher. A couple of tweaks and I think this will be good to go.

In case you're not familiar with associated types in traits, this is what they refer to:

https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html

artichoke-core/src/hash.rs
lopopolo: another reason to prefer an associated item here:

If I'm building part of the standard library that wants to generically constrain the Artichoke interpreter to include a hasher, most of the time I don't care what `BuildHasher` is used, just that the trait is implemented.

Having the generic parameter in the trait means it must be named, even if all we do is add `where B: core::hash::BuildHasher`. Using an associated type with that same bound improves the ergonomics a bit.

See this example for places where we want to constrain the interpreter we get based on some of these traits in `artichoke-core`:

https://github.com/artichoke/artichoke/blob/1c1702894e6ef237b8d7644c16f6b1bc9ae5f5e7/spinoso-symbol/src/lib.rs#L234-L244

masonforest:
Ah, I should have read more carefully. You did say to use an associated type in your feedback.

I think this should be getting close now.

lopopolo (approved):
This looks great. Thanks!

I'll merge once CI passes.

lopopolo:
It looks like rustfmt failed. If you run this locally it should fix everything up:

```shell
bundle install
bundle exec rake
```

Or if you don't have Ruby, it looks like all it needs is a:

```shell
cargo fmt
```

lopopolo:
Sorry for the back and forth here. I hope you don't mind that I've tweaked your branch to add the docs and clean things up a bit. Once the tests pass, this is good to go.

lopopolo:
this PR bumps `artichoke-core` to version 0.10.0.

PR #1506, opened 2021-11-20
Add GitHub Actions workflow to run markdown-link-check

lopopolo:
Artichoke already incorporates this tool in the `Rakefile` as the
`release:markdown_link_check` task, but it is slow and doesn't get
regularly run.

Incorporate this into CI such that links are checked on every PR that
touches markdown files and on a weekly cron with the same cadence as the
audit workflow.

Ignore docs.rs links since none of the crates in this workspace are published yet.

Fixup all of the outstanding broken markdown links.

### DIFF
--- a/.github/workflows/markdown-link-check.yaml
+++ b/.github/workflows/markdown-link-check.yaml
@@ -26,7 +26,7 @@ jobs:
       - name: Purge vendored sources
         run: |
           shopt -s globstar
-          rm -rf **/vendor/*/
+          rm -rf **/vendor/*/*
       - uses: gaurav-nelson/github-action-markdown-link-check@v1
         # checks all markdown files from /docs including all subfolders
         with:
--- a/.github/workflows/markdown-link-check.yaml
+++ b/.github/workflows/markdown-link-check.yaml
@@ -22,7 +22,7 @@ jobs:
     name: Check links
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
       - name: Purge vendored sources
         run: |
           shopt -s globstar
--- a/.github/workflows/markdown-link-check.yaml
+++ b/.github/workflows/markdown-link-check.yaml
@@ -23,6 +23,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
+      - name: Purge vendored sources
+        run: |
+          shopt -s globstar
+          rm -rf **/vendor/*/
       - uses: gaurav-nelson/github-action-markdown-link-check@v1
         # checks all markdown files from /docs including all subfolders
         with:
--- a/.github/workflows/markdown-link-check.yaml
+++ b/.github/workflows/markdown-link-check.yaml
@@ -1,0 +1,32 @@
+---
+"on":
+  push:
+    branches:
+      - trunk
+    paths:
+      - .github/markdown-link-check.json
+      - .github/workflows/markdown-link-check.yaml
+      - "**/*.md"
+  pull_request:
+    branches:
+      - trunk
+    paths:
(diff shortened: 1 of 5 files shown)


.github/workflows/markdown-link-check.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - "**/*.md"
+  pull_request:
+    branches:
+      - trunk
+    paths:
+      - .github/markdown-link-check.json
+      - .github/workflows/markdown-link-check.yaml
+      - "**/*.md"
+  schedule:
+    - cron: "0 0 * * TUE"
```
lopopolo, line 18: the cron is necessary since sometimes code refactors which don't otherwise touch markdown can break links.

See the refactor from `artichoke_backend::fs` to `artichoke_backend::load_path`.

VISION.md
```diff
@@ -116,7 +116,7 @@ Artichoke will include implementation-agnostic [C APIs][a-c-api] targeting:
 [file system backends]:
   https://github.com/artichoke/artichoke/labels/A-filesystem
 [virtual file system]:
-  https://artichoke.github.io/artichoke/artichoke_backend/fs/index.html
+  https://artichoke.github.io/artichoke/artichoke_backend/load_path/index.html
```
lopopolo, line 119: this breakage was introduced here:

- https://github.com/artichoke/artichoke/pull/1140

spinoso-env/README.md
```diff
@@ -80,7 +80,7 @@ This crate requires [`std`], the Rust Standard Library.
 All features are enabled by default:
 
 - **system-env** - Enable an `ENV` backend that accesses the host system's
-  environment variables via the [`std::env`](module@std::env) module.
```
lopopolo, line 83: this looks like a copy-paste error out of the `lib.rs` documentation.

This syntax is used to differentiate the `std::env` module from the `env!` macro.

.github/workflows/markdown-link-check.yaml
```diff
@@ -0,0 +1,32 @@
...
+name: Markdown Links Check
+jobs:
+  check-links:
+    name: Check links
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: gaurav-nelson/github-action-markdown-link-check@v1
```
lopopolo: this will probably fail the first time up due to markdown in vendored sources.

Maybe we can do this in a preceding step?

```sh
shopt -s globstar
rm -rf **/vendor/*/*
```

I think this will keep all of the `vendor/*.md` files in tact.

lopopolo: It looks like that'll work:

```console
$ /usr/local/bin/bash --version
GNU bash, version 5.1.12(1)-release (x86_64-apple-darwin20.6.0)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
$ /usr/local/bin/bash -c 'shopt -s globstar ; rm -rf **/vendor/*/' # use bash 5
$ ls artichoke-backend/vendor/
README.md
```

lopopolo:
the only failures are 429s from GitHub.

```
FILE: ./spinoso-securerandom/README.md
[✖] https://github.com/artichoke/artichoke/actions → Status: 429
[✖] https://github.com/artichoke/artichoke/workflows/CI/badge.svg → Status: 429

15 links checked.

ERROR: 2 dead links found!
[✖] https://github.com/artichoke/artichoke/actions → Status: 429
[✖] https://github.com/artichoke/artichoke/workflows/CI/badge.svg → Status: 429
```

I don't think this is mergable as is because the cron will spam my email with failures.

PR #1510, opened 2021-11-22
Revert "Add a 'perl hash' implementation to spinoso-string"

lopopolo:
This reverts commit b990622b3fc8a03ab2f6dd23a46e53ac4893636f.

Bump `spinoso-string` to v0.6.0 for this breaking change.

Prefer to use the interpreter-global SipHash hasher introduced in #1488.

See these comments for a discussion on why this commit is dangerous, is no longer necessary, and can be revereted:

- https://github.com/artichoke/artichoke/pull/1450#issuecomment-975414054
- https://github.com/artichoke/artichoke/pull/1450#issuecomment-975593629
- https://github.com/artichoke/artichoke/pull/1450#issuecomment-975595948

cc @rurban thanks for the report.


lopopolo:
mruby 3.1.0-rc now uses FNV1a as its string hashing algorithm.

PR #1511, opened 2021-11-24
Ensure `core-regexp` feature is activated when onig is active

lopopolo:
`core-regexp-oniguruma` is intended to be a derived feature that enables `core-regexp` + a dep on `rust-onig`.


lopopolo:
This bug was found when working on https://github.com/artichoke/artichoke/pull/1512 and extracted into a dedicated PR.

PR #1519, opened 2021-11-26
Add APIs for encoding-aware character indexing and slicing to spinoso-string

lopopolo:
In Ruby, `String#[]` with either integer or range arguments performs substring slicing based on characters in the `String`s declared encoding.

```ruby
"🦀💎"[1]
=> "💎"
"🦀💎"[1, 10]
=> "💎"
```

Introduce `spinoso_string::String::get_char` for encoding-aware single character lookups. `get_char` returns `Option<&'_ [u8]>`.

Introduce `spinoso_string::String::get_char_slice` for encoding-aware character range lookups. `get_char_slice` returns `Option<&'_ [u8]>`. `get_char_slice` handles backwards ranges, ranges that start at `s.length`, and ranges that run beyond `s.length` in the same manner as MRI.

The algorithm for UTF-8 aware slicing is fairly complicated and extensively documented with inline comments and snippets from an MRI Ruby 3.0.2 REPL session. The algorithm is further complicated by a variety of fast paths and short circuiting in attempts to avoid an expensive byte-at-a-time UTF-8 decode.

Bump `spinoso-string` to v0.10.0.

This PR has made me realize that `spinoso-string` would be better structured if a `String`'s encoding were represented in the Rust type system. If `BinaryString` and `AsciiString` were their own types, the `Utf8String` type would be able to bump down the indentation level a lot and the fast pathing would be isolated to the cases that need it. See https://twitter.com/artichokeruby/status/1464117273862443010 for more context and thoughts along these lines. I'd like to defer work here until after #1222 is merged.

This PR was plucked out of #1222 from these commits:

- https://github.com/artichoke/artichoke/pull/1222/commits/e11c4d3e13fcae886bb76d10abd7395fc0d2d7aa
- https://github.com/artichoke/artichoke/pull/1222/commits/93241e0e80b99e7384a96d9be4caf335fa8ec36c
- https://github.com/artichoke/artichoke/pull/1222/commits/b324be1ff878de8fbfe985a1830eb2c354f3e84f

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -680,7 +680,7 @@ dependencies = [
 
 [[package]]
 name = "spinoso-string"
-version = "0.9.0"
+version = "0.10.0"
 dependencies = [
  "bstr",
  "bytecount",
--- a/spinoso-string/Cargo.toml
+++ b/spinoso-string/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "spinoso-string"
-version = "0.9.0"
+version = "0.10.0"
 authors = ["Ryan Lopopolo <rjl@hyperbo.la>"]
 edition = "2021"
 description = """
--- a/spinoso-string/README.md
+++ b/spinoso-string/README.md
@@ -27,7 +27,7 @@ Add this to your `Cargo.toml`:
 
 ```toml
 [dependencies]
-spinoso-string = "0.9"
+spinoso-string = "0.10"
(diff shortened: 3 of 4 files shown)


lopopolo:
This PR could use a lot more testing, but the doctests have reasonable coverage (and captured some bugs!) and this will be much easier to test when `Utf8String` is its own thing.

PR #1570, opened 2021-12-13
Use `cstr!` macro to create argspec literals in `mrb_get_args!`

lopopolo:
Remove some unnecessary casts that are not detected by clippy.


lopopolo:
The following issues were filed upstream against clippy that I think should have detected and suggested these changes:

- https://github.com/rust-lang/rust-clippy/issues/8120
- https://github.com/rust-lang/rust-clippy/issues/8119

PR #1575, opened 2021-12-13
[artichoke-backend]: make `Integer#chr` encoding aware

### LINKED ISSUE #1563: Make `Integer#chr` encoding-aware
Now that `String` + `Encoding` support is landed in Artichoke, update `Integer#chr` to create `String`s with the proper encoding.

Supporting the `encoding` parameter is not yet possible since `Encoding` objects are not available in the VM yet.

https://github.com/artichoke/artichoke/blob/b4f93ee4220a3f785cd53f838bed96986731a21a/artichoke-backend/src/extn/core/integer/mod.rs#L99-L150

Instead of creating `String`s out of `vec![]` and converting them in the VM, allocate spinoso-strings directly with either `spinoso_string::String::{ascii,binary,utf8}` and call `spinoso_string::String::alloc_value`.

See `securerandom` for examples of this:

https://github.com/artichoke/artichoke/blob/b4f93ee4220a3f785cd53f838bed96986731a21a/artichoke-backend/src/extn/stdlib/securerandom/trampoline.rs#L15-L16

kolharsam:
This PR should resolve #1563.

@lopopolo I wasn't sure of all the changes because my `cargo` setup has been acting up lately. The best way to check my changes would've been to run on CI. I would appreciate a little more support :pray:

### DIFF
--- a/artichoke-backend/src/extn/core/integer/mod.rs
+++ b/artichoke-backend/src/extn/core/integer/mod.rs
@@ -101,7 +101,10 @@ impl Integer {
             let mut message = spinoso_string::String::ascii(b"encoding parameter of Integer#chr (given ".to_vec());
             message.extend(encoding.inspect(interp));
             message.extend(b") not supported");
-            spinoso_string::String::alloc_value(message, interp)
+            match spinoso_string::String::alloc_value(message, interp) {
+                Ok(msg) => Ok(msg.inspect(interp)),
+                Err(_) => Err(NotImplementedError::from(message).into()),
+            }
         } else {
             // When no encoding is supplied, MRI assumes the encoding is
             // either ASCII or ASCII-8BIT.
@@ -142,7 +145,10 @@ impl Integer {
                 _ => {
                     let mut message = spinoso_string::String::new();
                     write!(&mut message, "{} out of char range", self.as_i64()).map_err(WriteError::from)?;
-                    spinoso_string::String::alloc_value(message, interp)
+                    match spinoso_string::String::alloc_value(message, interp) {
+                        Ok(msg) => Ok(msg.inspect(interp)),
+                        Err(_) => Err(RangeError::from(message).into()),
+                    }
                 }
             }
         }
--- a/artichoke-backend/src/extn/core/integer/mod.rs
+++ b/artichoke-backend/src/extn/core/integer/mod.rs
@@ -98,10 +98,10 @@ impl Integer {
 
     pub fn chr(self, interp: &mut Artichoke, encoding: Option<Value>) -> Result<Vec<u8>, Error> {
         if let Some(encoding) = encoding {
-            let mut message = b"encoding parameter of Integer#chr (given ".to_vec();
+            let mut message = spinoso_string::String::ascii(b"encoding parameter of Integer#chr (given ".to_vec());
             message.extend(encoding.inspect(interp));
             message.extend(b") not supported");
-            Err(NotImplementedError::from(message).into())
+            spinoso_string::String::alloc_value(message, interp)
         } else {
             // When no encoding is supplied, MRI assumes the encoding is
             // either ASCII or ASCII-8BIT.
@@ -140,9 +140,9 @@ impl Integer {
                     Ok(vec![chr])
                 }
                 _ => {
-                    let mut message = String::new();
+                    let mut message = spinoso_string::String::new();
                     write!(&mut message, "{} out of char range", self.as_i64()).map_err(WriteError::from)?;
-                    Err(RangeError::from(message).into())
+                    spinoso_string::String::alloc_value(message, interp)
                 }
             }
         }
--- a/artichoke-backend/src/extn/core/integer/mod.rs
+++ b/artichoke-backend/src/extn/core/integer/mod.rs
@@ -96,15 +96,12 @@ impl Integer {
         self.0 as f64
     }
 
-    pub fn chr(self, interp: &mut Artichoke, encoding: Option<Value>) -> Result<Vec<u8>, Error> {
+    pub fn chr(self, interp: &mut Artichoke, encoding: Option<Value>) -> Result<spinoso_string::String, Error> {
         if let Some(encoding) = encoding {
-            let mut message = spinoso_string::String::ascii(b"encoding parameter of Integer#chr (given ".to_vec());
+            let mut message = b"encoding parameter of Integer#chr (given ".to_vec();
             message.extend(encoding.inspect(interp));
             message.extend(b") not supported");
-            match spinoso_string::String::alloc_value(message, interp) {
-                Ok(msg) => Ok(msg.inspect(interp)),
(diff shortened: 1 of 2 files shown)


artichoke-backend/src/extn/core/integer/mod.rs
```diff
@@ -140,9 +143,12 @@ impl Integer {
                     Ok(vec![chr])
```
lopopolo: This `Vec<u8>` is what should be allocated a a spinoso string with the proper encoding.

artichoke-backend/src/extn/core/integer/mod.rs
```diff
@@ -98,10 +98,13 @@ impl Integer {
...
         if let Some(encoding) = encoding {
-            let mut message = b"encoding parameter of Integer#chr (given ".to_vec();
+            let mut message = spinoso_string::String::ascii(b"encoding parameter of Integer#chr (given ".to_vec());
             message.extend(encoding.inspect(interp));
             message.extend(b") not supported");
-            Err(NotImplementedError::from(message).into())
+            match spinoso_string::String::alloc_value(message, interp) {
+                Ok(msg) => Ok(msg.inspect(interp)),
+                Err(_) => Err(NotImplementedError::from(message).into()),
+            }
```
lopopolo: This code constructs a `Vec<u8>` which is passed directly to an exception constructor. It does not need to go through a spinoso string.

artichoke-backend/src/extn/core/integer/mod.rs
```diff
@@ -140,9 +143,12 @@ impl Integer {
...
                 }
                 _ => {
-                    let mut message = String::new();
+                    let mut message = spinoso_string::String::new();
                     write!(&mut message, "{} out of char range", self.as_i64()).map_err(WriteError::from)?;
-                    Err(RangeError::from(message).into())
+                    match spinoso_string::String::alloc_value(message, interp) {
+                        Ok(msg) => Ok(msg.inspect(interp)),
+                        Err(_) => Err(RangeError::from(message).into()),
+                    }
```
lopopolo: This code constructs a `Vec<u8>` which is passed directly to an exception constructor. It does not need to go through a spinoso string.

artichoke-backend/src/extn/core/integer/mod.rs
```diff
@@ -98,10 +98,13 @@ impl Integer {
 
     pub fn chr(self, interp: &mut Artichoke, encoding: Option<Value>) -> Result<Vec<u8>, Error> {
```
lopopolo: This is the crux of the change I'd like us to make:

```suggestion
    pub fn chr(self, interp: &mut Artichoke, encoding: Option<Value>) -> Result<spinoso_string::String, Error> {
```

lopopolo (changes requested):
Hi @kolharsam sorry for not being clear enough in the source issue.

The bits in the code you've changed are places where error messages are being constructed. The exception types in Artichoke Ruby take `Vec<u8>` directly and don't need to go through spinoso string to turn those messages into raised exceptions.

The spot I've pointed out below that returns `Ok(vec![])` is the bit that should be turned into a spinoso string with an encoding, since that is what is returned to callers of `chr` as a Ruby `String`.

Let's change the return type of the Rust `chr` function to `Result<spinoso_string::String, Error>` and fixup the code to make that work.

kolharsam:
@lopopolo the [build](https://github.com/artichoke/artichoke/runs/4527594608?check_suite_focus=true#step:6:113) is failing due to an instance of `ConvertMut` missing for the `spinoso_string::String` with the changes from my latest commit. I don't think we'd need such an instance, but if we do, where should I be adding this?

lopopolo:
@kolharsam this looks good, there's one more spot we need to fixup.

`TryConvert` and `TryConvertMut` are Artichoke's versions of `TryFrom` from the Rust standard library for boxing native Rust types into Ruby `Value`s and unboxing Ruby `Value`s to native Rust types. `TryConvert` and `TryConvertMut` are only implemented for types in the Rust standard library like `Vec<u8>`, `i64`, and `Option<bool>`.

For _Spinoso_ types, the way to convert them into Ruby `Value`s (which is refered to as "boxing" them) is to use [`BoxUnboxVmValue::alloc_value`] if we're allocating a new object or [`BoxUnboxVmValue::box_into_value`] if we're mutating or reallocating an existing Ruby `Value`.

[`BoxUnboxVmValue::alloc_value`]: https://artichoke.github.io/artichoke/artichoke_backend/convert/trait.BoxUnboxVmValue.html#tymethod.alloc_value
[`BoxUnboxVmValue::box_into_value`]: https://artichoke.github.io/artichoke/artichoke_backend/convert/trait.BoxUnboxVmValue.html#tymethod.box_into_value

You can see this in action for `securerandom` here - we create a spinoso string and then allocate a Ruby value to store it and export the `String` to the VM:

https://github.com/artichoke/artichoke/blob/2f937cb20f4b493e6753795706a112d8d2b1c0c3/artichoke-backend/src/extn/stdlib/securerandom/trampoline.rs#L15-L16

Methods defined in `extn/core/$TYPE/mod.rs` usually return Rust native types which are then adapted into Ruby by a set of trampolines:

- `extn/core/$TYPE/trampoline.rs` which turns the native types and errors into Artichoke boxed Ruby values and exception values.
- `extn/core/$TYPE/mruby.rs` which turns the Artichoke-level representations into mruby boxes and converts exception values into an unwind.

Because we were previously returning a `Vec<u8>` which can only be converted to a Ruby `String` via `TryConvertMut`, we'll need to update the `integer::trampoline::chr` function to call `alloc_value` instead (line 8 in the snippet below):

https://github.com/artichoke/artichoke/blob/2f937cb20f4b493e6753795706a112d8d2b1c0c3/artichoke-backend/src/extn/core/integer/trampoline.rs#L5-L9

artichoke-backend/src/extn/core/integer/mod.rs
```diff
@@ -132,13 +132,14 @@ impl Integer {
...
-                    Ok(vec![chr])
-                }
+
+                // ASCII encoding - chr[0 - 127]
+                // Binary/ASCII-8BIT encoding - chr[128 - 255]
+
+                // Create a single byte `String` from the character given by `self`.
+                Ok(chr @ 0..=127) => Ok(spinoso_string::String::ascii(vec![chr])),
+                Ok(chr @ 128..=255) => Ok(spinoso_string::String::binary(vec![chr])),
                 _ => {
```
lopopolo: I think this could be converted into an `Err(_)` pattern which would ensure we're exhaustively matching the `Ok` variants above.

```suggestion
                Err(_) => {
```

kolharsam:
@lopopolo, I'm not sure if you had a chance to review, but I think I have the right changes this time around. 
Thanks again for your support on the work!

lopopolo (approved):
This looks great! Thanks for working through this with me.

PR #1582, opened 2021-12-27
Update ruby dep to the latest release

b-n:
A new Christmas time, a new ruby version.

Also update rubocop version since it doesn't hurt.

Tested with just a `rake build` and everything Just Worked:tm: 

Note: Updated the bundled with to 2.3.3 since the 2.2.15 was throwing an error about `DidYouMean.SPELL_CHECKERS` being referenced ([fixed in 2.3.1](https://github.com/rubygems/rubygems/pull/5202))


lopopolo:
It looks like the rebundling pulled in an updated version of `rubocop` which brings in some new lints and new cop violations:

```
Run bundle exec rubocop --format github

Error: Style/CollectionCompact: Use `compact` instead of `reject(&:nil?)`.

Error: Style/CollectionCompact: Use `compact!` instead of `reject!(&:nil?)`.

Error: Style/FileWrite: Use `File.write`.
Error: Process completed with exit code 1.
```

For now, would you mind downgrading RuboCop to `1.23.0` by manually editing the lockfile?

lopopolo:
The windows failure means Ruby 3.1.0 hasn't been built by the RubyInstaller folks and wired up in ruby/setup-ruby yet.

See:

- https://github.com/ruby/setup-ruby/pull/250#issuecomment-1001647707

b-n:
> For now, would you mind downgrading RuboCop to 1.23.0 by manually editing the lockfile?

Sure can, but alternatively, I could just fix those rubocop changes too :smile: (unless you've got a particular opinion on these rules of course)

As for that Windows issue, shall we just leave this open until it's built?

lopopolo:
> Sure can, but alternatively, I could just fix those rubocop changes too

oh sure if you'd like to do that, let's do it. One thing to keep in mind is that RuboCop is running on code that implements the standard library, so the best option might be to disable the warning.

b-n:
I'm just realizing the levels of inception in this project, it's pretty damn cool.

So just so my head is correct here, some methods (e.g. compact) are literally just ruby methods created on other ruby functions which are being interpreted by the rust code? :mage: 

If the above is true, is there actually a way via rust code to implement a method such as `compact`? (not that you would in this case, but maybe in another place if it made sense for speed optimization etc).

lopopolo:
> So just so my head is correct here, some methods (e.g. compact) are literally just ruby methods created on other ruby functions which are being interpreted by the rust code? 🧙

Yup that's right. The classes and modules in Ruby Core (like `String`, `Hash`, and `Regexp`) are defined in native code (either C, in the mruby codebase, or in Rust, in the Artichoke codebase which is slowly eating the mruby code).

Once the types are defined in native code, native code attaches some functions to these types. Then, the source code in these `rb` files in `extn/core` get loaded into the interpreter, which reopens the core class or module and defines the rest of the API.

The approach to what should be defined in Rust vs Ruby is kinda inconsistent, but in general, operations that allocate will be defined in native code. Methods that take blocks are a bit fiddly to implement in Rust correctly (such as handling `return` or `break` inside the block) so those mostly are implemented in Ruby.

> If the above is true, is there actually a way via rust code to implement a method such as `compact`? (not that you would in this case, but maybe in another place if it made sense for speed optimization etc).

Here's what this looks like for `Array#pop`, going from where the function is injected into the VM down to where it is implemented on the `spinoso-array` data structure:

Attach the native function to the type so compiled bytecode can jump to this routine:

https://github.com/artichoke/artichoke/blob/53e96d0a593983316e464eb0e3713a542b06ab45/artichoke-backend/src/extn/core/array/mruby.rs#L32

Extract arguments from the interpreter VM's stack, transform them into boxed values Artichoke Rust code can operate on, and delegate to Rust code that can return results. This function transforms returned errors into an unwinding exception.

https://github.com/artichoke/artichoke/blob/53e96d0a593983316e464eb0e3713a542b06ab45/artichoke-backend/src/extn/core/array/mruby.rs#L243-L256

The Rust trampoline unboxes native types from the boxed Ruby value:

https://github.com/artichoke/artichoke/blob/53e96d0a593983316e464eb0e3713a542b06ab45/artichoke-backend/src/extn/core/array/trampoline.rs#L267-L288

Unboxing routines are implemented here:

https://github.com/artichoke/artichoke/blob/53e96d0a593983316e464eb0e3713a542b06ab45/artichoke-backend/src/extn/core/array/mod.rs#L235-L322

An unboxed value is a proxy object that allows the VM heap to maintain ownership of the object.

`Array::pop` on `spinoso-array`:

https://github.com/artichoke/artichoke/blob/53e96d0a593983316e464eb0e3713a542b06ab45/spinoso-array/src/array/vec/mod.rs#L711-L713

lopopolo:
> As for that Windows issue, shall we just leave this open until it's built?

sounds good to me 👍 

I added a label to mark this ticket as blocked on upstream.

lopopolo:
I know this isn't obvious and I don't think anyone other than me can make this change, but filing a note for myself that once this lands, I'll need to backport these changes to project-infrastructure IaC and mirror the version bump in all other artichoke organization repos.

https://github.com/artichoke/project-infrastructure/blob/45061deddde7a3b1c79095ca1f619c4be66b646c/github/ruby-version.tf

b-n:
Thanks so much for the explanation! It honestly makes a lot of sense now that you explain it all - that's damn clever!

I'm wondering if there's a good place to keep that above information as an "How to understand what's going on - for noobs" sort of thing. It could living in CONTRIBUTING.md, but then it get's massive. If you see these questions coming up often, let me know, I'm happy to transpose this into a Github Wiki and/or contribute there too.

lopopolo:
If you wanted to add a markdown file with this info to the `docs` folder, I'd love to merge that PR. Or maybe `ARCHITECTURE.md` is a good place for it.

Another interesting bit that goes into the approach here is https://hyperbo.la/w/source-level-polymorphism/. There are 3 array data structures that all use different vector types with the same API that can be swapped freely with each other at compile time.

lopopolo:
I cherry-picked 250bf7e13348a4a569efd3e77180f382d6e249b8 into https://github.com/artichoke/artichoke/pull/1587.

lopopolo:
Upstream PR to add a Windows build of Ruby 3.1.0 to `ruby/setup-ruby`:

- https://github.com/ruby/setup-ruby/pull/253

b-n:
@lopopolo it looks like there's a Windows build for ruby upstream now. I've merged in the latest trunk, so we should be good to go!

lopopolo (approved):

PR #1583, opened 2021-12-27
Sync rust versions

### LINKED ISSUE #1581: Add `rust-version` field to all crate manifests and automate upgrading them when bumping the Rust toolchain
I like the idea of adding this bit of metadata, but I'd love to do it to all crates in this workspace and maybe automate bumping these fields along with the version in `rust-toolchain` every time the Rust compiler is released. Maybe with a Ruby script or a Rake task?

Let's explore this in a new PR.

_Originally posted by @lopopolo in https://github.com/artichoke/artichoke/pull/1522#discussion_r760790516_

https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

b-n:
Something to try and help with #1581 

I tried looking for some TOML parsers, but there's nothing that looks like it can easily edit in place - so regex is our best friend I guess. It should be stable, and/or error if there are ever new crates created that don't have the `rust-version` attribute in their `Cargo.toml`.

I'll have a quick think about how to add an action for this. I'm thinking a nightly action that creates a PR (think dependabot?) with the newest version. It should be doable, just need to find a way of programmatically getting the current stable version number (I have some ideas)

Fixes #1581.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0-pre.0"
 authors = ["Ryan Lopopolo <rjl@hyperbo.la>"]
 license = "MIT"
 edition = "2021"
+rust-version = "1.57.0"
 readme = "README.md"
 repository = "https://github.com/artichoke/artichoke"
 documentation = "https://artichoke.github.io/artichoke/artichoke/"
--- a/Rakefile
+++ b/Rakefile
@@ -164,7 +164,7 @@ namespace :pkg do
     next_rust_version = "rust-version = \"#{rust_version}\""
 
     failures = Dir.glob("#{__dir__}/{,*/}Cargo.toml").map do |file|
-      contents = File.open(file).read
+      contents = File.read(file)
 
       if (existing_version = contents.match(regexp))
         File.write(file, contents.gsub(regexp, next_rust_version)) if existing_version != rust_version
--- a/Rakefile
+++ b/Rakefile
@@ -159,7 +159,7 @@ end
 namespace :pkg do
   desc 'Sync the root rust-toolchain version to all crates'
   task :'rust_version:sync' do
-    rust_version = File.open('rust-toolchain').read.chomp
+    rust_version = File.read('rust-toolchain').chomp
     regexp = /^rust-version = "(.*)"$/
     next_rust_version = "rust-version = \"#{rust_version}\""
 
--- a/Rakefile
+++ b/Rakefile
@@ -163,7 +163,9 @@ namespace :pkg do
     regexp = /^rust-version = "(.*)"$/
     next_rust_version = "rust-version = \"#{rust_version}\""
 
-    failures = Dir.glob("#{__dir__}/{,*/}Cargo.toml").map do |file|
+    pkg_files = FileList.new('*/Cargo.toml').include('Cargo.toml')
+
+    failures = pkg_files.map do |file|
       contents = File.read(file)
 
       if (existing_version = contents.match(regexp))
--- a/Rakefile
+++ b/Rakefile
@@ -155,3 +155,26 @@ namespace :release do
     end
   end
 end
+
+namespace :pkg do
+  desc 'Sync the root rust-toolchain version to all crates'
+  task :'rust_version:sync' do
+    rust_version = File.open('rust-toolchain').read.chomp
+    regexp = /^rust-version = "(.*)"$/
+    next_rust_version = "rust-version = \"#{rust_version}\""
+
+    failures = Dir.glob("#{__dir__}/{,*/}Cargo.toml").map do |file|
+      contents = File.open(file).read
+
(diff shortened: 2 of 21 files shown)


b-n:
I've done some searching about how to get the latest rust version based on stable (now asking on discord), and have at least this as a sort of proxy: https://api.github.com/repos/rust-lang/rust/releases?per_page=1&page=1&sort=created&direction=desc (the latest created github release).

It should "work" in terms of making a PR at least, so maybe thats sufficient for now.

lopopolo:
> I've done some searching about how to get the latest rust version based on stable

This looks pretty nifty but I'm not sure if it is too clever. Let's leave this change out of this PR and maybe land it as a follow up change.

Cargo.toml
```diff
@@ -16,6 +16,7 @@ Windows, macOS, and Linux, as well as WebAssembly via Emscripten.
 """
 keywords = ["artichoke", "artichoke-ruby", "mri", "cruby", "ruby"]
 categories = ["command-line-utilities"]
+rust-version = "1.57.0"
```
lopopolo: Sorry to nitpick here, but I consider the `rust-version` field semantically linked to `edition`. Would you mind moving this line to line 7 directly below the `edition = "2021"` line?

Can we make this change for all of the `Cargo.toml`s?

Rakefile
```diff
@@ -155,3 +155,26 @@ namespace :release do
     end
   end
 end
+
+namespace :pkg do
+  desc 'Sync the root rust-toolchain version to all crates'
+  task :'rust_version:sync' do
```
lopopolo, line 161: this is awesome! Can you update the list of rake tasks in `CONTRIBUTING.md`?

Please run `bundle exec rake --tasks` and paste the output here:

https://github.com/artichoke/artichoke/blob/53e96d0a593983316e464eb0e3713a542b06ab45/CONTRIBUTING.md#L61-L86

Rakefile
```diff
@@ -155,3 +155,26 @@ namespace :release do
...
+    rust_version = File.open('rust-toolchain').read.chomp
```
lopopolo: We can avoid possibly leaking a file handle by doing this:

```suggestion
    rust_version = File.read('rust-toolchain').chomp
```

Rakefile
```diff
@@ -155,3 +155,26 @@ namespace :release do
...
+    regexp = /^rust-version = "(.*)"$/
+    next_rust_version = "rust-version = \"#{rust_version}\""
+
+    failures = Dir.glob("#{__dir__}/{,*/}Cargo.toml").map do |file|
```
lopopolo: the `release` namespace uses a `FileList` to pick a list of files relevant to the task. Can we be consistent and use that here?

I think it simplifies this expressions to:

```ruby
  link_check_files = FileList.new('**/Cargo.toml') do |f|
    f.exclude('node_modules/**/*')
    f.exclude('**/target/**/*')
    f.exclude('**/vendor/**/*')
    f.include('Cargo.toml')
  end
```

I find this to be a bit more readable than the glob expansion style.

Rakefile
```diff
@@ -155,3 +155,26 @@ namespace :release do
...
+      contents = File.open(file).read
```
lopopolo:
```suggestion
      contents = File.read(file)
```

lopopolo (changes requested):
Thanks for the PR @b-n. I like the approach you've taken with the Rake task. I have a couple of asks for cleanups but this is mostly good to go as is.

lopopolo:
@b-n I edited the PR description to add a magic GitHub _Fixes #xxx_ annotation to get the autoclose linking wired up.

Rakefile
b-n: Can do!

Just curious, would you expect this rake task to go more than one sub directory deep? From what I can see so far, all `Cargo.toml` files are one directory deep at most (and one at the root level).

If it's only one level deep, then its "safer" to just go single `*` expansion instead of recursive `**` expansion `FileList.new('*/Cargo.toml').include('Cargo.toml')` ?

Edit: I've pushed an update to the code to show you what I mean.

lopopolo: that sounds good to me.

lopopolo (approved):
Thanks @b-n this looks great. Once the build is green I'll smash the merge button. Happy New Year and thanks for contributing 💎

lopopolo:
> I'll have a quick think about how to add an action for this. I'm thinking a nightly action that creates a PR (think dependabot?) with the newest version.

oh wow I missed this! that would be amazing. I'd love to get this automated to that degree.

One other thing to think about is that when the `rust-toolchain` is updated, there is a corresponding job that sometimes fails in artichoke/docker-artichoke-nightly which requires another PR:

- https://github.com/artichoke/docker-artichoke-nightly/pull/69
- https://github.com/artichoke/docker-artichoke-nightly/pull/60

PR #1584, opened 2021-12-28
First attempt at adding coverage for binaries

### LINKED ISSUE #595: Investigate adding UI tests for artichoke binary frontend

### LINKED ISSUE #1579: `artichoke` returns different errors than MRI when failing to load a program file
```console
$ ruby bar.rb a b c
ruby: No such file or directory -- bar.rb (LoadError)
$ cargo run -q --bin artichoke -- bar.rb a b c
No such file or directory (os error 2) (IOError)
$ chmod -r foo.rb
$ ruby foo.rb a b c
ruby: Permission denied -- foo.rb (LoadError)
$ cargo run -q --bin artichoke -- foo.rb a b c
Permission denied (os error 13) (IOError)
```

It looks like `io::Error` during file loading need to be translated into `LoadError` and the file path needs to be included in the error message. This error needs to be properly transformed in `Eval::eval_file`:

https://github.com/artichoke/artichoke/blob/3fbea81564844e45f57c4c2bf8b7c7815ec88f09/artichoke-backend/src/eval.rs#L54-L61

b-n:
Fixes #595 

First cut at trying to get binary coverage from within the test suite.


lopopolo:
I think this ticket is also related too:

- https://github.com/artichoke/artichoke/issues/595

Cargo.toml
```diff
@@ -182,6 +182,9 @@ stdlib-strscan = ["artichoke-backend/stdlib-strscan"]
 stdlib-time = ["artichoke-backend/stdlib-time"]
 stdlib-uri = ["artichoke-backend/stdlib-uri"]
 
+[dev-dependencies]
+insta = "1.9.0"
```
lopopolo: let's also add a dev dep on `bstr` so we can use `Vec<u8>` for snapshot outputs with more readable debug impls

tests/integration/main.rs
```diff
@@ -0,0 +1,48 @@
...
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+#[allow(dead_code)]
+#[derive(Debug)]
+struct CommandOutput<'a> {
+    call_args: Vec<&'a str>,
+    status: i32,
+    stdout: String,
+    stderr: String,
```
lopopolo: Let's make this change after adding a `use bstr::BString` above.

```suggestion
    stdout: BString,
    stderr: BString,
```

tests/integration/main.rs
```diff
@@ -0,0 +1,48 @@
+mod artichoke;
+
+use std::env;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+#[allow(dead_code)]
+#[derive(Debug)]
+struct CommandOutput<'a> {
```
lopopolo: can we add a builder pattern style set of constructors for this type?

I'm thinking `CommandOutput::new()` which creates an empty struct, `CommandOutput::with_call_args`, `CommandOutput::with_status`, etc. all of which take `self` and return `Self`.

If we make the `with_stdout` and `with_stderr` methods take `Vec<u8>` and wrap to `BString` internally, I think it will make the test assertion less line-noisy.

tests/integration/main.rs
```diff
@@ -0,0 +1,48 @@
...
+    stdout: String,
+    stderr: String,
+}
+
+fn manifest_path() -> String {
+    env::var_os("CARGO_MANIFEST_DIR").unwrap().into_string().unwrap()
+}
+
+fn binary_path<'a>(name: &'a str) -> Result<PathBuf, String> {
+    let path = Path::new(&manifest_path()).join("target").join("debug").join(name);
```
lopopolo:
```suggestion
    let manifest_path = env::var_os("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR environment variable not set by cargo");
    let path = PathBuf::from(manifest_path).join("target").join("debug").join(name);
```

tests/integration/main.rs
```diff
@@ -0,0 +1,48 @@
...
+    status: i32,
+    stdout: String,
+    stderr: String,
+}
+
+fn manifest_path() -> String {
+    env::var_os("CARGO_MANIFEST_DIR").unwrap().into_string().unwrap()
+}
+
+fn binary_path<'a>(name: &'a str) -> Result<PathBuf, String> {
```
lopopolo: I don't think the named lifetime is required here and can be elided because:

- there is only one lifetime in the arg list
- the lifetime does not appear in the output type.

tests/integration/main.rs
```diff
@@ -0,0 +1,48 @@
...
+    let path = Path::new(&manifest_path()).join("target").join("debug").join(name);
+    match path.exists() {
+        true => Ok(path),
+        false => Err(format!("Can't find binary {} in ./target/debug/", name)),
+    }
```
lopopolo: I don't think clippy likes matching on a `bool`, but if this passes the linter it looks readable to me.

tests/integration/main.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+fn run<'a>(binary_name: &'a str, call_args: Vec<&'a str>) -> Result<CommandOutput<'a>, String> {
+    let binary = binary_path(binary_name)?;
+
+    let output = Command::new(&binary)
```
lopopolo: The deref here is not necessary since `Command::new` takes `AsRef<OsStr>` and `binary is not used elsewhere in this function.

```suggestion
    let output = Command::new(binary)
```

tests/integration/main.rs
```diff
@@ -0,0 +1,48 @@
...
+        .args(call_args.clone())
+        .output()
+        .expect(format!("Failed to run ruby app {}", binary_name).as_str());
```
lopopolo: this function is already fallible, can we use `map_err` and `?` to turn this expect into a returned error?

Or is this a fatal error?

tests/integration/main.rs
```diff
@@ -0,0 +1,48 @@
...
+    let status = match output.status.code() {
+        Some(code) => code,
+        None => -1,
+    };
+
+    Ok(CommandOutput {
+        call_args,
+        status,
+        stdout: std::str::from_utf8(&output.stdout).expect("Invalid Utf-8 string").to_string(),
+        stderr: std::str::from_utf8(&output.stderr).expect("Invalid Utf-8 string").to_string(),
```
lopopolo: With the changes proposed above, we'll be able to pass stdout and stderr directly, which would allow testing console output of invalid UTF-8 bytes on non-Windows platforms.

tests/integration/snapshots/integration__artichoke__ruby_app__fizz_buzz.rb.snap
```diff
@@ -0,0 +1,16 @@
...
+
+---
+Ok(
+    CommandOutput {
+        call_args: [
+            "./tests/integration/apps/fizz_buzz.rb",
+        ],
+        status: 1,
+        stdout: "",
+        stderr: "\u{1b}[0m\u{1b}[0m\u{1b}[1mTraceback\u{1b}[0m (most recent call last):\n\t3: from ./tests/integration/apps/fizz_buzz.rb:3\n\t2: from (eval):590:in each\n./tests/integration/apps/fizz_buzz.rb:4: \u{1b}[0m\u{1b}[1mundefined method '%'\u{1b}[0m\u{1b}[1m (\u{1b}[0m\u{1b}[1m\u{1b}[4mNoMethodError\u{1b}[0m\u{1b}[1m)\n\u{1b}[0m",
```
lopopolo: with `BString` debug impl, I think this will be a bit more readable.

tests/integration/main.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    let output = Command::new(&binary)
+        .args(call_args.clone())
+        .output()
+        .expect(format!("Failed to run ruby app {}", binary_name).as_str());
+
+    let status = match output.status.code() {
+        Some(code) => code,
+        None => -1,
+    };
```
lopopolo: can you restructure this match as an `unwrap_or_else`?

tests/integration/main.rs
```diff
@@ -0,0 +1,48 @@
+mod artichoke;
+
+use std::env;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+#[allow(dead_code)]
```
lopopolo: why is this warning silencing attribute required?

tests/integration/artichoke.rs
```diff
@@ -0,0 +1,24 @@
+use insta;
+use crate::run;
+use std::fs;
+use std::path::Path;
+
+#[test]
+fn test_ruby_apps() -> Result<(), String> {
```
lopopolo: I think I'd prefer to see this harness factored out a bit and each app instantiated as a separate `#[test]` (no automatic discovery) which would let the Rust test harness parallelize execution.

tests/integration/snapshots/integration__artichoke__ruby_app__hello_world.rb.snap
```diff
@@ -0,0 +1,16 @@
+---
```
lopopolo: how do these snaps get updated if there is an intentional change? Do you think we can or should add docs to `CONTRIBUTING.md` for this?

Cargo.toml
lopopolo: I think we should be able to get by with `default-features = false, features = ["std"]`.

tests/integration/apps/fizz_buzz.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+[0..100].each do |i|
```
lopopolo: I think you meant to add parens around this range.

```suggestion
(0..100).each do |i|
```

tests/integration/main.rs
b-n: What's your thoughts on a CommandRequest DAO sort of thing, and then doing something like `CommandOutput::from_request(request: &CommandRequest) -> Self` sort of thing too?

Reason, it'd be nice to store some of this in the CommandOutput (e.g. call args), since it might help understand how the snapshot was created etc (note, the snapshots don't auto clean up)

tests/integration/main.rs
b-n: Ah yes, that was due to an old change I had there. Will fix!

tests/integration/main.rs
b-n: Since the execution will just throw things on stderr, I'm 99% sure this is a fatal error. e.g. Rust just failed spawn/execute entirely (I imagine this would never happen, but never say never I guess)

tests/integration/main.rs
b-n: Fancy rust is fancy. I've made the change, however your thoughts, should it be fatal or not?

From my understanding, all applications should return an integer status code. Setting it to -1 is just because I've always seen positive codes (although I believe there's nothing preventing a negative), and it's just a slight testing signal "something went wrong". Thoughts?

tests/integration/main.rs
b-n: Weirdly enough, `insta::assert_debug_snapshot!` doesn't actually count as reading the variables in this struct :shrug:. Speaking of which, there are other strategies for output, I just used the "easiest" one. Alternative options are things such as `asset_toml_snapshot!` are available - which IMO would maybe be better for human readable structured data - so I'll give that a go and see if this warning goes with it.

tests/integration/artichoke.rs
b-n: Yeah, good call, totally agree!

tests/integration/snapshots/integration__artichoke__ruby_app__hello_world.rb.snap
b-n: Reading my mind. I was/am planning on adding something to `CONTRIBUTING.md`

Just so you know though, all test runs will test whether the output matches. If it doesn't, the test execution fails (nice), but also a draft/review snap is created. Those can be viewed via `cargo insta review` which is just a small CLI that allows for easy review.

They can all be forced to be accepted via a command, but snapshot tests should be done consciously. e.g. I'm accepting this snapshot, because I know it is changing, and I accept said changes. They can almost be seen as a reflection of the execution. This does mean you don't want to test rand functions etc, since the output isn't predictable. They're just a safe guard at the end of the day.

tests/integration/main.rs
b-n: Weirdly enough, it works! Resolving

tests/integration/snapshots/integration__artichoke__ruby_app__fizz_buzz.rb.snap
b-n: Still lots of `\n\n\n\n` everywhere on FizzBuzz. I'll play with the toml formatter, see if it does multiline in this case :crossed_fingers:

tests/integration/main.rs
lopopolo: > Weirdly enough, `insta::assert_debug_snapshot!` doesn't actually count as reading the variables in this struct 🤷.

this is a change from recent rustc. Derived `Debug` impls don't count as used fields:

- https://github.com/rust-lang/rust/issues/88900

tests/integration/main.rs
lopopolo: sounds good, the `expect` makes sense here.

tests/integration/snapshots/integration__artichoke__ruby_app__fizz_buzz.rb.snap
lopopolo: Yea I think we might want a custom debug impl that calls `lines()` on the `Vec<u8>` and then renders each line with `writeln!(f, "{:?}", line.as_bstr())?`.

lopopolo:
this looks like it's close. I've removed the WIP label, please convert from draft whenever you're feeling ready @b-n

lopopolo:
one way to avoid the complexities of the minimum versions CI is to make the UI tests a sub-workspace like `fuzz`. Then UI tests can be run as a separate rake task and separate CI job.

tests/integration/main.rs
```diff
@@ -0,0 +1,93 @@
...
+impl Serialize for CommandOutput {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut s = serializer.serialize_struct("parameters", 4)?;
+        s.serialize_field("call_args", &self.call_args)?;
+        s.serialize_field("status", &self.status)?;
+        s.serialize_field("stdout", &format!("{}", self.stdout))?;
+        s.serialize_field("stderr", &format!("{}", self.stderr))?;
```
lopopolo: Formatting the `BString` with `Display` is not quite right since the impl replaces invalid UTF-8 with '�'.

I think we'll probably want something like this:

```suggestion
        let mut buf = String::new();
        for line in self.stdout.lines() {
            write!(&mut buf, "{:?}", line.as_bstr())?;
        }
        s.serialize_field("stdout", &buf)?;

        buf = String::new();
        for line in self.stderr.lines() {
            write!(&mut buf, "{:?}", line.as_bstr())?;
        }
        s.serialize_field("stderr", &buf)?;
```

This will turn non-printable control characters into ASCII escape sequences and invalid UTF-8 byte sequences into hex escape sequences, which keeps everything lossless but also human readable.

tests/integration/main.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn new() -> Self {
+        Self {
+            call_args: vec![],
+            status: -1,
+            stdout: BString::from(""),
+            stderr: BString::from(""),
+        }
+    }
+
+    fn with_args(&mut self, call_args: &Vec<&str>) -> &mut Self {
```
lopopolo: Passing `&Vec` and `&String` is not idiomatic. These types coerce to slices which are one word smaller in stack size and have all of the same methods.

```suggestion
    fn with_args(&mut self, call_args: &[&str]) -> &mut Self {
```

tests/integration/main.rs
```diff
@@ -0,0 +1,93 @@
+mod artichoke;
+
+use std::env;
+use std::path::PathBuf;
+use std::process::Command;
+use serde::ser::{Serialize, Serializer, SerializeStruct};
+use bstr::BString;
```
lopopolo: I think for the changes below we'll also need to pull in `ByteSlice`.

```suggestion
use bstr::{BString, ByteSlice};
```

tests/integration/main.rs
```diff
@@ -0,0 +1,93 @@
+mod artichoke;
+
+use std::env;
+use std::path::PathBuf;
+use std::process::Command;
+use serde::ser::{Serialize, Serializer, SerializeStruct};
```
lopopolo: If the fmt job is finicky, I think it might ask for an empty line between `std` and 3p import groups.

`bundle exec rake fmt` will run nightly rustfmt which should fix this up.

tests/integration/artichoke.rs
```diff
@@ -0,0 +1,18 @@
+use crate::run;
+
+const BINARY: &str = "artichoke";
+const FIXTURES_ROOT: &str = "./tests/integration/apps/";
+
+#[test]
+fn test_hello_world() {
```
lopopolo: usually when I write unit tests in Rust in these `test/integration` crates, I skip the `test_` prefix in the function name, like this:

https://github.com/artichoke/artichoke/blob/5f70f7ee62b903bcb62ef03f0e994771da9744f3/artichoke-backend/tests/integration/gc.rs#L16-L23

lopopolo (commented):
this is an awesome change, we're getting close.

The TOML formatting is nifty, but I think the minimal versions CI job will be tricky to get happy with a chunky dep like `insta`. Can we move the UI tests to a sub-workspace?

tests/integration/main.rs
b-n: Magic! I fixed this with the linting errors - since they agree too haha.

tests/integration/main.rs
b-n: I had to change a couple of things on this. At first I was confused as to why to split on lines, to convert to `Bstr`, and reassemble (e.g. just convert the BString, instead of line by line), but then I saw that `\n` was being output and combining it all into one string.

I took an alternative than a loop, since `write!` would need to be `writeln!` if we still wanted it human readable. However, in doing that, we might be adding an additional `\n` in the output, which might not be 100% correct - thus the `join` approach. Iter vs. for loop doesn't worry me, so if you prefer a for loop, let me know!

lopopolo: Yup that works for me. This is similar to how another test helper for printing backtraces in functional tests works:

https://github.com/artichoke/artichoke/blob/5f70f7ee62b903bcb62ef03f0e994771da9744f3/artichoke-backend/src/test/prelude.rs#L81-L89

tests/integration/main.rs
```diff
@@ -0,0 +1,111 @@
+mod artichoke;
+
+use std::env;
+use std::path::PathBuf;
+use std::process::Command;
```
lopopolo:
```suggestion
use std::process::{Command, Output};
```

Can we import this type which gets used below to avoid using a fully qualified item path?

tests/integration/main.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    fn with_args(&mut self, call_args: &[&str]) -> &mut Self {
+        self.call_args
+            .append(&mut (*call_args).iter().map(|x| x.to_string()).collect());
+        self
+    }
+
+    fn with_command_output(&mut self, output: std::process::Output) -> &mut Self {
```
lopopolo: This is mostly a style nit, I generally only use zero or one level of modules in item paths (except in macros).

```suggestion
    fn with_command_output(&mut self, output: Output) -> &mut Self {
```

lopopolo: I love that you've passed this in by value so we can deconstruct it and take the outputs from it without cloning.

tests/integration/main.rs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        let mut s = serializer.serialize_struct("parameters", 4)?;
+        s.serialize_field("call_args", &self.call_args)?;
+        s.serialize_field("status", &self.status)?;
+        let stdout = self
+            .stdout
+            .lines()
+            .map(|line| format!("{:?}", line.as_bstr()))
+            .collect::<Vec<String>>()
+            .join("\n");
```
lopopolo: This is great. Thank you!

tests/integration/main.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+fn binary_path(name: &str) -> Result<PathBuf, String> {
+    let executable = binary_name(name);
+    let manifest_path =
+        env::var_os("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR environment variable not set by cargo");
+    let path = PathBuf::from(manifest_path)
+        .join("target")
+        .join("debug")
+        .join(executable.clone());
```
lopopolo: The `PathBuf` we have here is derefing to `Path`. `Path::join` takes `AsRef<Path>`, which `&str` satisfies. We can avoid the clone here by passing a borrow of `executable`.

```suggestion
        .join(&executable);
```

tests/integration/snapshots/integration__artichoke__fizz_buzz.snap
```diff
@@ -0,0 +1,111 @@
+---
+source: tests/integration/artichoke.rs
+assertion_line: 29
+expression: "run(BINARY, &[&path]).unwrap()"
+
+---
+call_args = ['./tests/integration/fixtures/fizz_buzz.rb']
+status = 0
+stdout = '''
+"FizzBuzz"
```
lopopolo: Ahh that's right, `BStr`'s debug impl quotes the string. I deal with this elsewhere in artichoke test harnesses by slicing away the quotes

https://github.com/artichoke/artichoke/blob/5f70f7ee62b903bcb62ef03f0e994771da9744f3/artichoke-backend/src/test/prelude.rs#L85-L86

It's still not perfect because `"` in the interior will render as `\"`, but for now I think it is the best we can do.

lopopolo (approved):
This looks great. I left a couple of comments of tweaks we can make, but as of now I'd be happy merging this as is so I'll approve.

Let me know in the comments here if you want to address these last couple of comments. If I don't hear from you, I'll aim to have this merged by the end of the week.

Don't worry about the failing _CI / Compile with minimum dependency versions_ CI job. We can handle fixing this in a followup PR. If you'd like to take this let me know, we can collaborate either in a new GitHub Issue or on Discord.

Thanks so much for this contribution. I didn't know where to start here and I appreciate you running with this @b-n. 📈 💯

tests/integration/snapshots/integration__artichoke__fizz_buzz.snap
b-n: There is one upshot to having the quotes, and that's end of line whitespace (something my editor was trimming and causing random test failures (ask me how I know haha). So I guess you could look at it as a happy side effect of letting you know where the end of line whitespace is? :joy:

b-n:
@lopopolo I attempted to split the tests into another workspace, but that's a whole bunch of headaches. I thought splitting it into test_utils would help - but of course it doesn't because it's just including that dependency later. I'm happy to remove that last commit which is that change, or I can leave it there for the next iteration where someone with more know how figures it out?

Apart from that, I'm pretty sure most is covered! (Apart from contributing updates, but I'll maybe make a PR for that also :smile:)

b-n:
My brain worked about the test isolation stuff finally. https://github.com/b-n/artichoke/commit/efcc7601215a4803d68e4ed8bd836a3d8fdabbc0 <-- I can push this commit onto this branch and/or make a new PR - whatever is easier to review :+1:

lopopolo:
Let's push this up in a new PR

lopopolo (approved):
Woohoo!

PR #1585, opened 2021-12-29
Keep integration tests isolated from the rest (fix min dep issue)

b-n:
#1584 breaks CI because of minimum version workflows.

This PR breaks the integration tests into a separate crate to fix that issue and also tries to make the `rake test:` namespace a bit more full for these separate crates.


integration-tests/src/lib.rs
```diff
@@ -84,6 +84,7 @@ fn binary_path(name: &str) -> Result<PathBuf, String> {
     let manifest_path =
         env::var_os("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR environment variable not set by cargo");
     let path = PathBuf::from(manifest_path)
+        .join("..")
```
b-n: So this is a little dodgy. The `integration-tests` package doesn't build the binaries. The only way to really achieve this would be to copy/paste the entire contents of `src/*` into this folder, since then it'd be building the binary. This hacky way relies on the binary being built (which is why there is a build for this test run) :shrug:

lopopolo:
I pushed a few commits to get the integration tests (which I've renamed to `ui` tests) wired up in CI

lopopolo:
@b-n dodgy is ok. I've encoded the required setup in the CI workflow. I think this is a good kludge 😄

lopopolo:
the current setup requires a debug build of the artichoke binaries

PR #1619, opened 2021-12-31
Add casing tests

b-n:
Adding tests for [upper|lower]case. Brings test coverage in spinoso-string (specifically only tests in `lib.rs`) from 48% => 54%

Happy to make those test helper functions (capitalize/lowercase/uppercase) functions on test module itself (would make the tests a little more condensed, but less explicit)


lopopolo (approved):
❤️ �� 🚀

PR #1634, opened 2022-01-12
Try to reduce complexity in spinoso-string/lib.rs

b-n:
Just a lift and shift of the iter definitions into `iter.rs` in an attempt to reduce the size of `lib.rs` - and to collect the `iter` methods in one place from `impls.rs` too.

I understand why having the Iter as part of the base crate. This change didn't take any real time, so happy if the opinion is: "No, this should still be in `lib.rs`" of course

Actual new code: 0


spinoso-string/src/lib.rs
```diff
@@ -1220,6 +867,34 @@ impl String {
...
+    /// let s = String::from("abc");
+    ///
+    /// let mut iterator = s.into_iter();
+    ///
+    /// assert_eq!(iterator.next(), Some(b'a'));
+    /// assert_eq!(iterator.next(), Some(b'b'));
+    /// assert_eq!(iterator.next(), Some(b'c'));
+    /// assert_eq!(iterator.next(), None);
+    /// ```
+
```
lopopolo, line 893: nit to fix the whitespace

```suggestion

    /// Returns an iterator that moves over the remaining bytes of a slice
    ///
    /// # Examples
    ///
    /// ```
    /// use spinoso_string::String;
    ///
    /// let s = String::from("abc");
    ///
    /// let mut iterator = s.into_iter();
    ///
    /// assert_eq!(iterator.next(), Some(b'a'));
    /// assert_eq!(iterator.next(), Some(b'b'));
    /// assert_eq!(iterator.next(), Some(b'c'));
    /// assert_eq!(iterator.next(), None);
    /// ```
```

spinoso-string/src/iter.rs
```diff
@@ -0,0 +1,589 @@
...
+/// # }
+/// # example().unwrap();
+/// ```
+///
+/// [`String`]: crate::String
+/// [`center`]: crate::String::center
+/// [encoding-aware]: crate::Encoding
+/// [Conventionally UTF-8]: crate::Encoding::Utf8
+#[derive(Debug, Clone)]
+pub struct Center<'a, 'b> {
```
lopopolo: `Center` is a bit more complex of an iterator (with its error type and all). Let's move `Center` and its associated impls and types to a new `center.rs` file, like how `Chars` has its own file.

lopopolo (changes requested):
love this idea. let's also move `Center` to its own source file since it is a chunky set of types and impls.

spinoso-string/src/center.rs
```diff
@@ -0,0 +1,174 @@
...
+///
+/// This error corresponds to the [Ruby `ArgumentError` Exception class].
+///
+/// When the **std** feature of `spinoso-string` is enabled, this struct
+/// implements [`std::error::Error`].
+///
+/// [`String::center`]: crate::String::center
+/// [Ruby `ArgumentError` Exception class]: https://ruby-doc.org/core-2.6.3/ArgumentError.html
+/// [`std::error::Error`]: https://doc.rust-lang.org/std/error/trait.Error.html
+#[allow(clippy::module_name_repetitions)]
```
b-n, line 20: You can imagine why this is needed. It could be renamed `Error`, but it feels less descriptive.

If it was to be renamed to just Error, then we'd need to do `pub use center::Error as CenterError` in lib.rs to keep the public api for spinoso-string the same. Not sure how idiomatic that is though.

lopopolo (approved):

PR #1636, opened 2022-01-14
Use Ruby 3.1

MSP-Greg:
See https://github.com/MSP-Greg/artichoke/runs/4812551354.  Windows passed, but the linter had some problem...


MSP-Greg:
The changes to ci.yaml use [MSP-Greg/setup-ruby-pkgs@v1](https://github.com/MSP-Greg/setup-ruby-pkgs), which knows which clang package to install (mingw64 or ucrt64).  So, if one switched to Ruby 3.0.1, the correct package would be installed.  Note that the package is not installed on the windows-2022 image.

If you'll stick to Ruby 3.1 or later, the changes can be reverted and the following step added:
```yml
      - name: Windows - install clang
        if: startsWith(matrix.os, 'windows')
        run: pacman -S mingw-w64-ucrt-x86_64-clang
```

lopopolo:
This PR is obsoleted by:

- https://github.com/artichoke/artichoke/pull/1694

PR #1678, opened 2022-02-06
Refactor `String` internals to be closer to MRI's encoding API

b-n:
# Description

As per a previous conversation: "Strings don't have encoding, Strings are encoded"[^1] - and with that comes this change. Specifically, each type of encoded string now has it's own methods to operate on it's own internal buffer.

Some things I attempted here:
- Never make the inner implementation public
- Try to keep Ruby API methods as much in the main `String` lib as possible. Why? Trying to keep the rust implementation/expectation in one place, with the ruby expectation elsewhere. (see `chomp` for example, no specific encoding implementation etc.)

:warning: Important to note, there is already another breaking change in the form of `set_encoding` no longer existing - and maybe it shouldn't exist in general (ruby `force_encoding` is gong to be .... fun).  This method was only being used in `artichoke-backend`, and the addition of a `from_raw_parts_with_encoding` covers the use case at least.

Things TODO (maybe for a future PR since this is mostly just migration, no "new" features so to speak):
- [Now?] - should the implementations be in the `mod.rs` file (I see this is how it's done in `artchoke-backend` at least.
- `chars` and `codepoints` can likely use a more generic iterator now. They still have switches on `Encoding`, but I believe these should be able to be implemented directly on the encoded types.
- Of course many more unit tests on the encoded types. I only copied across the relevant ones (and threw in a default empty constructor test for a start)
- Documentation in the encoded specific types?
- `make_binary` needs to return a new instance, not reassign (according to cruby implementation at least)
- some optimizations in the `chr` and `ord` functions in each encoded type. My gut feeling is that these functions can be generalized with other functions on the encoded types
- likewise for `ends_with`, maybe that should just be `as_slice().ends_with(separator)` instead of having their own repeated fn etc.

[^1]: https://twitter.com/artichokeruby/status/1464117275137511427


lopopolo:
Hey @b-n thanks for putting this PR together. Reaching out to acknowledge that I've gotten the notification, but it will likely be a bit before I can get to this to review. Thanks for your patience.

b-n:
@lopopolo no worries! Hope all is well!

I'll do some double due diligence tomorrow and make sure everything is still doing things the same way it was done previously.

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
+use alloc::vec::Vec;
```
lopopolo: I think this source file should be one of these two:

- `spinoso-string/src/ascii_string.rs`
- `spinoso-string/src/ascii_string/mod.rs`

But now that I have given this some thought, I think I'd like all of the encoding-specific impls to be placed in the source tree like this:

- `spinoso-string/src/enc/ascii/{mod,iter,..}.rs`
- `spinoso-string/src/enc/binary/{mod,iter,..}.rs`

And the enum of all possible encodings and the other dispatching bits should live in `spinoso-string/stc/enc/mod.rs`.

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+// Constructors
+impl AsciiString {
+    pub fn new(buf: Vec<u8>) -> Self {
+        Self { inner: buf }
+    }
+}
+
+impl fmt::Debug for AsciiString {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Utf8String")
```
lopopolo:
```suggestion
        f.debug_struct("AsciiString")
```

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+    }
+}
+
+// Debug
+impl AsciiString {
+    #[inline]
+    #[must_use]
+    pub fn as_bstr(&self) -> &BStr {
+        self.inner.as_bstr()
+    }
```
lopopolo: `BStr` I think is a presentation and implementation detail. I'd prefer for code that wants to treat slices as `BStr`s to have to do the conversion explicitly. I don't want `BStr`s to be part of the public API for any types in this crate.

There should be no loss of expressivity here. calling `s.as_slice().as_bstr()` gets you the same thing.

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+impl AsciiString {
+    pub fn new(buf: Vec<u8>) -> Self {
+        Self { inner: buf }
+    }
+}
+
+impl fmt::Debug for AsciiString {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Utf8String")
+            .field("buf", &self.as_bstr())
```
lopopolo:
```suggestion
            .field("buf", &self.inner.as_bstr())
```

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+}
+
+// Raw Accessors
+impl AsciiString {
+    #[inline]
+    #[must_use]
+    pub fn as_vec(&self) -> &Vec<u8> {
+        &self.inner
+    }
+
```
lopopolo: is this ever used? I imagine `AsciiString::as_slice` is the method that will get called. `&Vec<_>` is not a very useful type in Rust. There's not much you can do that you can't do on `&[_]`.

```suggestion
```

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+    #[inline]
+    #[must_use]
+    pub fn into_vec(self) -> Vec<u8> {
+        self.inner
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn as_slice(&self) -> &[u8] {
+        &self.inner
```
lopopolo:
```suggestion
        self.inner.as_slice()
```

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn as_mut_slice(&mut self) -> &mut [u8] {
+        &mut self.inner
```
lopopolo:
```suggestion
        self.inner.as_mut_slice()
```

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+
+// Memory management
+impl AsciiString {
+    #[inline]
+    pub fn reserve(&mut self, additional: usize) {
+        self.inner.reserve(additional);
+    }
+
+    #[inline]
+    pub fn try_reserve(&mut self, additional: usize) -> Result<(), alloc::collections::TryReserveError> {
```
lopopolo: I know this is how the code looked before, but since we're doing this from scratch, I'd love it if we did a `use alloc::collections::TryReserveError;` at the top of this file and the `E` type was only `TryReserveError` here.

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+        self.inner.try_reserve(additional)
+    }
+
+    #[inline]
+    pub fn reserve_exact(&mut self, additional: usize) {
+        self.inner.reserve_exact(additional);
+    }
+
+    #[inline]
+    pub fn try_reserve_exact(&mut self, additional: usize) -> Result<(), alloc::collections::TryReserveError> {
```
lopopolo: ditto

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+
+    #[inline]
+    pub fn try_push_codepoint(&mut self, codepoint: i64) -> Result<(), InvalidCodepointError> {
+        if let Ok(byte) = u8::try_from(codepoint) {
+            self.push_byte(byte);
+            Ok(())
+        } else {
+            Err(InvalidCodepointError::codepoint_out_of_range(codepoint))
+        }
+    }
```
lopopolo: i love how much simpler this is without the `match` on the encoding.

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+mod tests {
+    use alloc::vec::Vec;
+
+    use quickcheck::quickcheck;
+
+    use super::AsciiString;
+
+    quickcheck! {
+        #[allow(clippy::needless_pass_by_value)]
+        fn fuzz_char_len_utf8_contents_ascii_string(contents: alloc::string::String) -> bool {
```
lopopolo: since the types don't have the same name anymore, I think we can `use alloc::string::String` at the top of the `mod tests` and make this method parameter type just `String`.

spinoso-string/src/ascii_string/ascii_string.rs
```diff
@@ -0,0 +1,516 @@
...
+            let expected = contents.len();
+            let s = AsciiString::new(contents.into_bytes());
+            s.char_len() == expected
+        }
+
+        #[allow(clippy::needless_pass_by_value)]
+        fn fuzz_len_utf8_contents_ascii_string(contents: alloc::string::String) -> bool {
```
lopopolo: ditto

spinoso-string/src/ascii_string/impls.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+impl<'a> From<Cow<'a, [u8]>> for AsciiString {
+    #[inline]
+    fn from(content: Cow<'a, [u8]>) -> Self {
+        Self::new(content.into_owned())
+    }
+}
+
+impl From<alloc::string::String> for AsciiString {
```
lopopolo: ditto, types don't shadow, let's import this.

spinoso-string/src/ascii_string/impls.rs
```diff
@@ -0,0 +1,99 @@
...
+        s.to_vec()
+    }
+}
+
+impl Deref for AsciiString {
+    type Target = [u8];
+
+    #[inline]
+    fn deref(&self) -> &[u8] {
+        &*self.as_slice()
```
lopopolo:
```suggestion
        self.as_slice()
```

spinoso-string/src/ascii_string/impls.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+impl DerefMut for AsciiString {
+    #[inline]
+    fn deref_mut(&mut self) -> &mut [u8] {
+        &mut *self.as_mut_slice()
```
lopopolo:
```suggestion
        self.as_mut_slice()
```

spinoso-string/src/ascii_string/mod.rs
```diff
@@ -0,0 +1,6 @@
+#[allow(clippy::module_inception)]
```
lopopolo: I think the docs here answer one of your open questions in the PR description:

- https://rust-lang.github.io/rust-clippy/master/#module_inception

nesting modules like this is confusing. I'd expect the contents of the inner `mod ascii_string` to be in this `mod.rs` file.

spinoso-string/src/codepoints.rs
```diff
@@ -74,6 +75,73 @@ impl fmt::Display for CodepointsError {
...
+
+    #[inline]
+    #[must_use]
+    pub const fn is_out_of_range(self) -> bool {
+        matches!(self.0, CodePointRangeError::OutOfRange(_))
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn message(self) -> alloc::string::String {
```
lopopolo, line 116: let's import the type above and use a short name since there's no type shadowing now.

spinoso-string/src/codepoints.rs
```diff
@@ -74,6 +75,73 @@ impl fmt::Display for CodepointsError {
...
+        // The longest error message is 27 bytes + a hex-encoded codepoint
+        // formatted as `0x...`.
+        const MESSAGE_MAX_LENGTH: usize = 27 + 2 + mem::size_of::<u32>() * 2;
+        let mut s = alloc::string::String::with_capacity(MESSAGE_MAX_LENGTH);
```
lopopolo, line 120: ditto

spinoso-string/src/codepoints.rs
```diff
@@ -125,7 +193,7 @@ impl<'a> TryFrom<&'a String> for Codepoints<'a> {
     fn try_from(s: &'a String) -> Result<Self, Self::Error> {
         let state = match s.encoding() {
             Encoding::Utf8 => {
-                if let Ok(s) = s.buf.to_str() {
+                if let Ok(s) = s.inner.as_vec().to_str() {
```
lopopolo: `to_str` is `ByteSlice::to_str`. As long as `ByteSlice` is in scope, this method is available on `&[u8]`.

```suggestion
                if let Ok(s) = s.inner.as_slice().to_str() {
```

spinoso-string/src/impls.rs
```diff
@@ -9,6 +9,18 @@ use std::io;
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Clone for String {
+    fn clone(&self) -> Self {
+        Self::with_bytes_and_encoding(self.inner.as_vec().clone(), self.encoding())
+    }
+}
```
lopopolo: Why can't we `#[derive(Clone)]`?

spinoso-string/src/impls.rs
```diff
@@ -9,6 +9,18 @@ use std::io;
 
 use crate::String;
 
+impl Default for String {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
lopopolo: ditto, why can't we `#[derive(Default)]`?

The default for `EncodedString` should be an empty UTF-8 string.

spinoso-string/src/impls.rs
```diff
@@ -172,35 +184,35 @@ impl From<&str> for String {
 impl From<String> for Vec<u8> {
     #[inline]
     fn from(s: String) -> Self {
-        s.buf
+        s.inner.as_vec().clone()
```
lopopolo: This conversion consumes the original string, so we don't need to clone here.

```suggestion
        s.into_vec()
```

spinoso-string/src/impls.rs
```diff
@@ -209,42 +221,42 @@ impl Deref for String {
 
     #[inline]
     fn deref(&self) -> &[u8] {
-        &*self.buf
+        &*self.inner.as_slice()
```
lopopolo:
```suggestion
        self.inner.as_slice()
```

spinoso-string/src/impls.rs
```diff
@@ -209,42 +221,42 @@ impl Deref for String {
...
     }
 }
 
 impl DerefMut for String {
     #[inline]
     fn deref_mut(&mut self) -> &mut [u8] {
-        &mut *self.buf
+        &mut *self.inner.as_mut_slice()
```
lopopolo:
```suggestion
        self.inner.as_mut_slice()
```

spinoso-string/src/impls.rs
```diff
@@ -209,42 +221,42 @@ impl Deref for String {
...
-        self.buf.as_mut_slice()
+        self.inner.as_mut_slice()
     }
 }
 
 impl Borrow<Vec<u8>> for String {
     #[inline]
     fn borrow(&self) -> &Vec<u8> {
-        &self.buf
+        self.inner.as_vec()
```
lopopolo: This will help us get rid of `as_vec`.

```suggestion
        self.inner.borrow()
```

lopopolo (changes requested):
I only took a look at `AsciiString`. I did not look at `BinaryString` and `Utf8String`, but I imagine a lot of the same feedback applies. I did not yet look at the changes to `String`.

This change looks mostly faithful to the discussion we had. Thanks for working on this! Let me know when you're ready for me to take another look.

lopopolo:
> ⚠️ Important to note, there is already another breaking change in the form of `set_encoding` no longer existing

if this is the case, please bump the version in `Cargo.toml` and wherever else it needs it. You'll have to run `cargo check` in all of the workspaces in this repo to update the lockfiles.

spinoso-string/src/encoded_string.rs
```diff
@@ -0,0 +1,481 @@
...
+use crate::encoding::Encoding;
+use crate::iter::{Bytes, IntoIter, Iter, IterMut};
+use crate::ord::OrdError;
+use crate::utf8_string::Utf8String;
+
+pub enum EncodedString {
+    Ascii(AsciiString),
+    Binary(BinaryString),
+    Utf8(Utf8String),
+}
```
lopopolo: > ruby `force_encoding` is gong to be .... fun

`String#force_encoding` can be implemented like this:

```rust
impl EncodedString {
    pub fn force_encoding(&mut self, encoding: Encoding) {
        if encoding == self.encoding() {
            return;
        }
        let buf = mem::replace(self.as_mut_vec(), vec![]);
        match encoding {
            Encoding::Ascii => *self = Self::Ascii(AsciiString::new(buf)),
            Encoding::Ascii => *self = Self::Binary(BinaryString::new(buf)),
            Encoding::Ascii => *self = Self::Utf8(Utf8String::new(buf)),
        }
    }
}
```

spinoso-string/src/encoded_string.rs
```diff
@@ -0,0 +1,481 @@
...
+    }
+
+    #[inline]
+    pub fn make_uppercase(&mut self) {
+        match self {
+            EncodedString::Ascii(n) => n.make_uppercase(),
+            EncodedString::Binary(n) => n.make_uppercase(),
+            EncodedString::Utf8(n) => n.make_uppercase(),
+        }
+    }
```
lopopolo: this all looks great and is exactly what I was hoping for.

b-n:
@lopopolo Thanks for the reivew! And thanks for being so thorough and patient on this, +3500 new lines is not how I'd usually operate, and i ended up losing context being so close in the changes. I'll make a pass through and get back to you as soon as I've got all the things updated!

spinoso-string/src/ascii_string/ascii_string.rs
b-n: ~~Hrrrmm, this one is tripping up on the String impls of `Borrow<Vec<u8>>` and `AsRef<Vec<u8>>`. By removing the `as_vec()` there is no access to the underlying vec from what I can see, and using `as_slice()` with any `to_vec()` or `into_vec()` is creating temporary values that are dropping (of course). Ref: https://github.com/b-n/artichoke/blob/spinoso-string-enum/spinoso-string/src/impls.rs#L249-L261~~

~~Are there ways around this without using `as_vec()` on the encoded types?~~

~~Additionally, this makes me think, should `EncodedString` also have a impls for the same as the encoded types? I imagine it's impossible to get to the borrow of `AsciiString` etc as present :thinking:. If I did this, I could actually just call `self.inner.borrow()` and remove the reliance on `as_vec()`..... I might give this a try and update this comment if it works.~~

Update: Okay yes, we do need `EncodedString` `impls` and their appropriate routing. A weird way to find this one out. You'll see what I mean in https://github.com/artichoke/artichoke/pull/1678/commits/f84b69d82209534823e26dad1e6b69f017c696e7 :+1:

spinoso-string/src/ascii_string/ascii_string.rs
b-n: Good call. Since removed!

I just recently realised that the `fmt::Debug` on the encoded implementations is likely to never be called (apart from maybe in testing etc). I'm curious, do you think the encoded implementations should be part of the public API? I can't currently forsee any real use case TBH.

spinoso-string/src/ascii_string/ascii_string.rs
b-n: Yes! There are definitely some places I was going along going: "ooooo noice"

spinoso-string/src/ascii_string/ascii_string.rs
b-n: This brings up an interesting point. Should the encoded implementaitons be called `AsciiString`? I've thrown in a `clippy::module_name_repetition` allow due to this. In theory all these implementations could be called `String` also, and then imported into `enc/mod.rs` as `use ascii::String as AsciiString`. IMO, it's better as `AsciiString` in this case, which is why I've updated the import for `mod tests` - but food for thought whilst we're here :smile:

spinoso-string/src/ascii_string/impls.rs
b-n: How do you feel about `fn deref(&self) -> &Self::Target`? (matches the rust documentation on deref)

spinoso-string/src/codepoints.rs
b-n: Usage of `spinoso::String` is still done on L189 by the looks: `impl<'a> TryFrom<&'a String> for Codepoints<'a> {`

I was thinking about this file too, I think there might be a way to use the new encoding string stuff in this file, but would rather tackle that in another PR I think.

spinoso-string/src/codepoints.rs
b-n: As above, `spinoso::String` is still in use in this file.

spinoso-string/src/impls.rs
b-n: :thinking: I think I did this because rust was complaining that `EncodedString` wasn't implementing Clone. However since the encoded types support Clone, that means EncodedString can support Clone, so I'll just do that instead! :+1:

spinoso-string/src/impls.rs
b-n: Original reason: Because EncodedString didn't implement default :grimacing: 

However, thinking about it now, I don't think EncodedString should be defining this. If someone is asking for a default from `String`, then shouldn't `String` be making that decision about what the default is? EncodedString is just a wrapper at present, and doesn't contain any real decision about what is "the default" so to speak on anything (e.g. should an Enum have a default value?)

Edit: I've pushed the change you've asked so it's easy to keep/revert, since this might just be a style thought hah.

spinoso-string/src/encoded_string.rs
b-n: Ah magic!

I'll leave this for a separate PR as to not add more to the scope of this massive PR as it :grimacing:

artichoke-backend/src/extn/core/string/mod.rs
```diff
@@ -59,12 +59,14 @@ impl BoxUnboxVmValue for String {
...
-        });
-        s.set_encoding(encoding);
+        let s = String::from_raw_parts_with_encoding(
+            RawParts {
+                ptr: ptr.cast::<u8>(),
+                length,
+                capacity,
+            },
+            encoding,
+        );
```
b-n, line 69: @lopopolo Come to think of it, does it make sense to implement `force_encoding` purely so there isn't the breaking change of removing `set_encoding`? Question along the same lines, is there much overhead on that `mem::replace` call - I assume it's cheap, but not free.

I had originally removed `set_encoding` until I had discovered `String#force_encoding` existed, so this might kill two birds with one stone.

b-n:
@lopopolo Sorry for taking so long to get around to parsing the feedback - had a vacation, got distracted by another side project for a week or so. I think this is ready for another round. I resolve everything that was updated to match, and left some replies on the unresolved ones to figure out what we should do etc :rocket:

spinoso-string/src/impls.rs
```diff
@@ -35,48 +35,48 @@ impl fmt::Write for String {
...
     #[inline]
     fn write_vectored(&mut self, bufs: &[io::IoSlice<'_>]) -> io::Result<usize> {
-        self.buf.write_vectored(bufs)
+        self.inner.as_mut_slice().write_vectored(bufs)
     }
 
     #[inline]
     fn flush(&mut self) -> io::Result<()> {
-        self.buf.flush()
+        self.inner.as_mut_slice().flush()
```
lopopolo: These want to use `as_mut_vec` to allow the String to grow.

spinoso-string/src/ord.rs
```diff
@@ -0,0 +1,97 @@
+#[cfg(feature = "std")]
+use std::fmt;
```
lopopolo: This should import `core::fmt` and not require the feature to impl Display.

lopopolo (changes requested):
This was a quick review on my mobile. I'll do a more in depth one later today.

spinoso-string/src/impls.rs
b-n: Alternatively, should `io::Write` be implemented on each encoded type?

lopopolo: I like the sound of that. It seems consistent with the other delegation we're doing

b-n: Super!

Also, sorry on the lack of attention to detail on my behalf in this PR. I really should have checked my changes a bit more for this sort of stuff. On the right track to splitting this all up though!

b-n: Done!

spinoso-string/src/enc/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+    pub fn new(buf: Vec<u8>, encoding: Encoding) -> Self {
+        match encoding {
+            Encoding::Ascii => Self::Ascii(AsciiString::new(buf)),
+            Encoding::Binary => Self::Binary(BinaryString::new(buf)),
+            Encoding::Utf8 => Self::Utf8(Utf8String::new(buf)),
+        }
+    }
+}
+
+impl EncodedString {
```
b-n, line 95: Note: The encoded types don't have a `encoding()` method. It felt like a bit of a circular reference. Theory is that those types would only really be used internally by String

Interestingly enough, this is the only method that isn't delegated from EncodedString to the encoded versions :shrug:

spinoso-string/src/enc/ascii/eq.rs
```diff
@@ -0,0 +1,75 @@
+use alloc::vec::Vec;
+
+use super::AsciiString;
+
+impl PartialEq<Vec<u8>> for AsciiString {
```
b-n, line 5: I'm curious whether we'd ever need PartialEq between the encoded types.

```ruby
a = "hello"
b = "hello".force_encoding(Encoding::ASCII)
a == b # true
```

^-- this is covered in String by checking the slices against each other, but in theory that could just `self.inner == other.inner` if these classes were implementing PartialEq to each other :shrug:

spinoso-string/src/impls.rs
```diff
@@ -172,35 +172,35 @@ impl From<&str> for String {
 impl From<String> for Vec<u8> {
     #[inline]
     fn from(s: String) -> Self {
-        s.buf
+        s.inner.into_vec()
```
b-n, line 175: I understand that this would consume `s`, which the previous code should also be doing right?

spinoso-string/src/impls.rs
```diff
@@ -253,13 +253,13 @@ impl<I: SliceIndex<[u8]>> Index<I> for String {
 
     #[inline]
     fn index(&self, index: I) -> &Self::Output {
-        Index::index(&self.buf, index)
+        Index::index(self.inner.as_slice(), index)
```
b-n, line 258: These haven't been implemented on the sub types, but they are simply relying on the slice. Worth delegating?

spinoso-string/src/lib.rs
```diff
@@ -54,190 +53,33 @@ pub use raw_parts::RawParts;
 mod center;
 mod chars;
 mod codepoints;
+mod enc;
 mod encoding;
 mod eq;
 mod impls;
 mod inspect;
 mod iter;
+mod ord;
```
b-n, line 62: `OrdError` and `InvalidCodepointsError` were both pulled into their respective mods (mostly just to keep `lib` clean of any Trait implementations. Everything here is direct methods now.

spinoso-string/src/lib.rs
```diff
@@ -256,12 +98,12 @@ impl Hash for String {
         // [3.0.2] > t.hash
         // => 3398383793005079442
         // ```
-        self.buf.hash(hasher);
+        self.inner.as_slice().hash(hasher);
```
b-n: Was a vec, is now a slice. Has the same effect from what I see. Alternatively, this could be delegated like other impls

spinoso-string/src/lib.rs
```diff
@@ -272,21 +114,21 @@ impl PartialEq for String {
         // [3.0.2] > s == t
         // => true
         // ```
-        self.buf[..] == other.buf[..]
+        *self.inner.as_slice() == *other.inner.as_slice()
```
b-n: So, i was doubting whether this is correct, but i'm 99% sure it is. You were previously doing direct slice comparison, and since `as_slice` returns a ref, then we needed to deref it.

However, since the encoded types now support Deref (returns `&[u8]`, we could make this `*self.inner == *other.inner` right?

spinoso-string/src/lib.rs
```diff
@@ -303,17 +145,18 @@ impl String {
...
-    /// const S: String = String::new();
+    /// let S: String = String::new();
     /// assert_eq!(S.encoding(), Encoding::Utf8);
     /// ```
     ///
     /// [conventionally UTF-8]: crate::Encoding::Utf8
     #[inline]
     #[must_use]
-    pub const fn new() -> Self {
+    pub fn new() -> Self {
```
b-n: Can no longer be `const` due to relying on the `EncodedString` constructor

spinoso-string/src/lib.rs
```diff
@@ -453,25 +301,8 @@ impl String {
...
-    /// let mut s = String::utf8(b"xyz".to_vec());
-    /// assert_eq!(s.encoding(), Encoding::Utf8);
-    /// s.set_encoding(Encoding::Binary);
-    /// assert_eq!(s.encoding(), Encoding::Binary);
-    /// ```
-    #[inline]
-    pub fn set_encoding(&mut self, encoding: Encoding) {
-        self.encoding = encoding;
+    pub fn encoding(&self) -> Encoding {
+        self.inner.encoding()
```
b-n, line 257: I mentioned elsewhere, but I removed the `set_encoding` since it didn't feel like forcing one EncodedString struct into the other made sense - however I think I should re-implement due to at least `String#force_encoding`. Thoughts?

spinoso-string/src/lib.rs
```diff
@@ -655,6 +486,32 @@ impl String {
...
+    ///
+    /// Violating these may cause problems like corrupting the allocator's
+    /// internal data structures.
+    ///
+    /// The ownership of `ptr` is effectively transferred to the `String` which
+    /// may then deallocate, reallocate or change the contents of memory pointed
+    /// to by the pointer at will. Ensure that nothing else uses the pointer
+    /// after calling this function.
+    #[must_use]
+    pub unsafe fn from_raw_parts_with_encoding(raw_parts: RawParts<u8>, encoding: Encoding) -> Self {
```
b-n, line 463: This exists because RawParts isn't going to ever support our specific use case. However, with `set_encoding`, the reason that this exists can go away (although, it will do more memcpy shifting around, not sure if that's an issue, and worthwhile to keep this anyway)

spinoso-string/src/lib.rs
```diff
@@ -671,7 +528,7 @@ impl String {
     /// [`from_raw_parts`]: String::from_raw_parts
     #[must_use]
     pub fn into_raw_parts(self) -> RawParts<u8> {
-        RawParts::from_vec(self.buf)
+        RawParts::from_vec(self.inner.into_vec())
```
b-n, line 483: Once again, `into_vec` will consume it, so I assume this is the "most correct" way to do this.

spinoso-string/src/lib.rs
```diff
@@ -1762,13 +1496,13 @@ impl String {
     #[inline]
     #[must_use]
     pub fn chop(&mut self) -> bool {
-        if self.buf.is_empty() {
+        if self.is_empty() {
             return false;
         }
-        let bytes_to_remove = if self.buf.ends_with(b"\r\n") {
+        let bytes_to_remove = if self.inner.as_slice().ends_with(b"\r\n") {
```
b-n, line 1454: maybe a new `ends_with()` fn on the encoded implementations would make this even nicer

spinoso-string/src/lib.rs
```diff
@@ -2373,183 +2007,7 @@ impl String {
...
-                        // decoding.
-                        (None, size) => {
-                            slice = &slice[size..];
-                            remaining -= size;
-                        }
-                    }
-                }
-            }
-        }
+        self.inner.get_char_slice(range)
```
b-n, line 1962: It feels like the encoded impls should be doing the above (early exit logic etc), however it is pretty irrelevant to the encoding.

b-n:
@lopopolo I've done some due diligence checks, and thrown a bunch of comments in with ideas and/or options. If you're happy with some of the comments, just resolve them. The biggest :question: for me is around the re-implementation of set_encoding, but it feels like it's needed no matter how I look at it.

The failing CI/CD parts I'll fix after the answers etc. One of them is coming because a `use` is not used during `no-default-features`. I'm assuming this can only be explicitly ignored etc.

artichoke-backend/src/extn/core/string/mod.rs
lopopolo: No this API is great! it lets us construct the string directly from its FFI representation. Let's keep this

spinoso-string/src/enc/ascii/eq.rs
lopopolo: Yup for `String`, `==` is only based on the underlying byte contents. Given that, I don't think it's necessary for us to implement the full matrix of `PartialEq` on all of the internal types.

spinoso-string/src/enc/ascii/impls.rs
```diff
@@ -0,0 +1,180 @@
...
+    #[inline]
+    fn from(s: &str) -> Self {
+        Self::new(s.as_bytes().to_vec())
+    }
+}
+
+impl From<AsciiString> for Vec<u8> {
+    #[inline]
+    fn from(s: AsciiString) -> Self {
+        s.to_vec()
```
lopopolo: should this be `s.into_vec()`? It looks like `self.to_vec()` is coming from the deref to `&[u8]`. This means we're allocating when we don't need to be since the `From` impl consumes `s`.

```suggestion
        s.into_vec()
```

spinoso-string/src/enc/binary/impls.rs
```diff
@@ -0,0 +1,180 @@
+use alloc::borrow::Cow;
+use alloc::string::String;
+use alloc::vec::Vec;
+use core::borrow::{Borrow, BorrowMut};
+use core::fmt::Arguments;
```
lopopolo: It looks like we need one of these:

```suggestion
#[cfg(feature = "std")]
use core::fmt::Arguments;
```

spinoso-string/src/enc/ascii/impls.rs
```diff
@@ -0,0 +1,180 @@
+use alloc::borrow::Cow;
+use alloc::string::String;
+use alloc::vec::Vec;
+use core::borrow::{Borrow, BorrowMut};
+use core::fmt::Arguments;
```
lopopolo: It looks like we need one of these:

```suggestion
#[cfg(feature = "std")]
use core::fmt::Arguments;
```

spinoso-string/src/enc/utf8/impls.rs
```diff
@@ -0,0 +1,180 @@
+use alloc::borrow::Cow;
+use alloc::string::String;
+use alloc::vec::Vec;
+use core::borrow::{Borrow, BorrowMut};
+use core::fmt::Arguments;
```
lopopolo: It looks like we need one of these:

```suggestion
#[cfg(feature = "std")]
use core::fmt::Arguments;
```

spinoso-string/src/enc/binary/impls.rs
```diff
@@ -0,0 +1,180 @@
...
+    #[inline]
+    fn from(s: &str) -> Self {
+        Self::new(s.as_bytes().to_vec())
+    }
+}
+
+impl From<BinaryString> for Vec<u8> {
+    #[inline]
+    fn from(s: BinaryString) -> Self {
+        s.to_vec()
```
lopopolo: same:

```suggestion
        s.into_vec()
```

spinoso-string/src/enc/impls.rs
```diff
@@ -0,0 +1,204 @@
+//use alloc::borrow::Cow;
+use alloc::vec::Vec;
+use core::borrow::{Borrow, BorrowMut};
+use core::fmt::Arguments;
```
lopopolo: It looks like we need one of these:

```suggestion
#[cfg(feature = "std")]
use core::fmt::Arguments;
```

spinoso-string/src/enc/impls.rs
```diff
@@ -0,0 +1,204 @@
...
+use std::io::{IoSlice, Result, Write};
+
+use super::EncodedString;
+
+#[cfg(feature = "std")]
+impl Write for EncodedString {
+    #[inline]
+    fn write(&mut self, buf: &[u8]) -> Result<usize> {
+        match self {
+            EncodedString::Ascii(n) => n.write(buf),
```
lopopolo: Sorry to be nit-picky, but can we replace all uses of `n` for the inner bit of the enum with `inner`?

spinoso-string/src/enc/mod.rs
lopopolo: sounds good to me

spinoso-string/src/enc/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+}
+
+// Defer to Encoded Implementation
+impl EncodedString {
+    #[inline]
+    #[must_use]
+    pub fn into_vec(self) -> Vec<u8> {
+        match self {
+            EncodedString::Ascii(n) => n.into_vec(),
```
lopopolo: Sorry to be nit-picky, but can we replace all uses of `n` for the inner bit of the enum with `inner`?

spinoso-string/src/enc/ascii/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+use core::slice::SliceIndex;
+
+use bstr::{ByteSlice, ByteVec};
+
+use crate::codepoints::InvalidCodepointError;
+use crate::iter::{Bytes, IntoIter, Iter, IterMut};
+use crate::ord::OrdError;
+
+#[allow(clippy::module_name_repetitions)]
+#[derive(Default, Clone)]
```
lopopolo:
```suggestion
#[derive(Default, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
```

spinoso-string/src/enc/binary/mod.rs
```diff
@@ -0,0 +1,492 @@
...
+use core::slice::SliceIndex;
+
+use bstr::{ByteSlice, ByteVec};
+
+use crate::codepoints::InvalidCodepointError;
+use crate::iter::{Bytes, IntoIter, Iter, IterMut};
+use crate::ord::OrdError;
+
+#[allow(clippy::module_name_repetitions)]
+#[derive(Default, Clone)]
```
lopopolo:
```suggestion
#[derive(Default, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
```

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -0,0 +1,1060 @@
...
+use core::slice::SliceIndex;
+
+use bstr::{ByteSlice, ByteVec};
+
+use crate::codepoints::InvalidCodepointError;
+use crate::iter::{Bytes, IntoIter, Iter, IterMut};
+use crate::ord::OrdError;
+
+#[allow(clippy::module_name_repetitions)]
+#[derive(Default, Clone)]
```
lopopolo:
```suggestion
#[derive(Default, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
```

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -0,0 +1,1060 @@
...
+    }
+}
+
+impl PartialEq for Utf8String {
+    fn eq(&self, other: &Self) -> bool {
+        *self.inner.as_slice() == *other.inner.as_slice()
+    }
+}
+
+impl Eq for Utf8String {}
```
lopopolo:
```suggestion
}
```

spinoso-string/src/enc/utf8/impls.rs
```diff
@@ -0,0 +1,180 @@
...
+    #[inline]
+    fn from(s: &str) -> Self {
+        Self::new(s.as_bytes().to_vec())
+    }
+}
+
+impl From<Utf8String> for Vec<u8> {
+    #[inline]
+    fn from(s: Utf8String) -> Self {
+        s.to_vec()
```
lopopolo: same:

```suggestion
        s.into_vec()
```

spinoso-string/src/enc/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+use ascii::AsciiString;
+use binary::BinaryString;
+use utf8::Utf8String;
+
+use crate::codepoints::InvalidCodepointError;
+use crate::encoding::Encoding;
+use crate::iter::{Bytes, IntoIter, Iter, IterMut};
+use crate::ord::OrdError;
+
+#[derive(Clone)]
```
lopopolo, line 21: Can we impl `Hash`, `PartialEq`, `Eq`, `PartialOrd`, and `Ord` on this enum so we can derive these on `String`?

Something like `Hash` and `PartialEq` in terms of inner byte slice. `PartialOrd` and `Ord` in terms of inner byte slice.

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -0,0 +1,1060 @@
...
+        //
+        // If the string is either all ASCII or all ASCII for a prefix
+        // of the string that contains the range we wish to slice,
+        // fallback to byte slicing as in the ASCII and binary fast path.
+        let consumed = match self.inner.find_non_ascii_byte() {
+            None => return self.inner.get(index..=index),
+            Some(idx) if idx > index => return self.inner.get(index..=index),
+            Some(idx) => idx,
+        };
+        let mut slice = &self.inner[consumed..];
```
lopopolo, line 221: can we add a TODO that we can use `get_unchecked` like we do on line 160?

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -0,0 +1,1060 @@
...
+                    remaining -= size;
+                }
+            }
+        }
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn get_char_slice(&self, range: Range<usize>) -> Option<&'_ [u8]> {
+        let Range { start, end } = range;
```
lopopolo, line 283: can we add a TODO that we can do the same fast path check as line 222 if `start > self.len()`?

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -0,0 +1,1060 @@
...
+                    slice = &slice[size..];
+                    remaining -= size;
+                }
+            }
+        }
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn get_char_slice(&self, range: Range<usize>) -> Option<&'_ [u8]> {
```
lopopolo, line 281: this is so much easier to follow since we're only worried about the UTF-8 path here.

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -0,0 +1,1060 @@
...
+impl Utf8String {
+    #[inline]
+    #[must_use]
+    pub fn is_ascii_only(&self) -> bool {
+        self.inner.is_ascii()
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn is_valid_encoding(&self) -> bool {
```
lopopolo, line 525: 🎉

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -0,0 +1,1060 @@
...
+        if self.is_ascii_only() {
+            return true;
+        }
+
+        simdutf8::basic::from_utf8(&self.inner).is_ok()
+    }
+}
+
+// Casing
+impl Utf8String {
```
lopopolo, line 535: Use `roe` for case changing operation in `spinoso-string`.

UTF-8 case changing needs to be parameterized on the casefolding strategy to account for e.g. Turkic or ASCII-only modes.

lopopolo: (no action needed in this PR. I left this comment to make it easier to file a ticket.)

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -0,0 +1,1060 @@
...
+            (None, _) => &self.inner[..1],
+        }
+    }
+
+    #[inline]
+    pub fn ord(&self) -> Result<u32, OrdError> {
+        let (ch, size) = bstr::decode_utf8(self.inner.as_slice());
+        match ch {
+            // All `char`s are valid `u32`s
+            // https://github.com/rust-lang/rust/blob/1.48.0/library/core/src/char/convert.rs#L12-L20
```
lopopolo: this comment is a holdover from when this `From` impl didn't exist. Let's delete it now since the `u32::from` call makes this super clear.

```suggestion
```

spinoso-string/src/enc/ascii/impls.rs
```diff
@@ -0,0 +1,180 @@
...
+        self.inner.write(buf)
+    }
+
+    #[inline]
+    fn write_all(&mut self, buf: &[u8]) -> Result<()> {
+        self.inner.write_all(buf)
+    }
+
+    #[inline]
+    fn write_fmt(&mut self, fmt: Arguments<'_>) -> Result<()> {
```
lopopolo: can we refer to this type as `fmt::Arguments<'_>` for consistency elsewhere in this crate?

spinoso-string/src/enc/binary/impls.rs
```diff
@@ -0,0 +1,180 @@
...
+        self.inner.write(buf)
+    }
+
+    #[inline]
+    fn write_all(&mut self, buf: &[u8]) -> Result<()> {
+        self.inner.write_all(buf)
+    }
+
+    #[inline]
+    fn write_fmt(&mut self, fmt: Arguments<'_>) -> Result<()> {
```
lopopolo: can we refer to this type as `fmt::Arguments<'_>` for consistency elsewhere in this crate?

spinoso-string/src/enc/impls.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn write_all(&mut self, buf: &[u8]) -> Result<()> {
+        match self {
+            EncodedString::Ascii(n) => n.write_all(buf),
+            EncodedString::Binary(n) => n.write_all(buf),
+            EncodedString::Utf8(n) => n.write_all(buf),
+        }
+    }
+
+    #[inline]
+    fn write_fmt(&mut self, fmt: Arguments<'_>) -> Result<()> {
```
lopopolo: can we refer to this type as `fmt::Arguments<'_>` for consistency elsewhere in this crate?

spinoso-string/src/enc/utf8/impls.rs
```diff
@@ -0,0 +1,180 @@
...
+        self.inner.write(buf)
+    }
+
+    #[inline]
+    fn write_all(&mut self, buf: &[u8]) -> Result<()> {
+        self.inner.write_all(buf)
+    }
+
+    #[inline]
+    fn write_fmt(&mut self, fmt: Arguments<'_>) -> Result<()> {
```
lopopolo: can we refer to this type as `fmt::Arguments<'_>` for consistency elsewhere in this crate?

spinoso-string/src/impls.rs
lopopolo: yup that's right. `From` impls are consuming so we can destruct `s` to avoid an allocation.

spinoso-string/src/impls.rs
lopopolo: this seems fine to me.

spinoso-string/src/lib.rs
```diff
@@ -54,190 +53,33 @@ pub use raw_parts::RawParts;
...
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(self.message())
-    }
-}
-
-#[cfg(feature = "std")]
-impl std::error::Error for OrdError {}
-
-#[derive(Default, Clone)]
+#[derive(Clone, Default)]
```
lopopolo: Elsewhere in this workspace, I always put `Default` before `Clone` in the derives list. Can we revert this change to keep things consistent?

```suggestion
#[derive(Default, Clone)]
```

spinoso-string/src/lib.rs
```diff
@@ -54,190 +53,33 @@ pub use raw_parts::RawParts;
...
 pub struct String {
-    buf: Vec<u8>,
-    encoding: Encoding,
+    inner: EncodedString,
```
lopopolo, line 75: if we derive `Hash`, `PartialEq`, `Eq`, `PartialOrd`, and `Ord` on `EncodedString`, we can derive them on `String` too.

spinoso-string/src/lib.rs
lopopolo: yup yup. I'd prefer we push this impl and its docs down into `EncodedString`.

spinoso-string/src/lib.rs
lopopolo: either works, but I'd prefer to push this down into `EncodedString`. The impl in `EncodedString` can basically be `*self.as_slice() == *other.has_slice()`.

The reason to do the deref on the slice is to use `PartialEq<[u8]> for [u8]` impl rather than an extra indirection with an additional ref.

spinoso-string/src/lib.rs
```diff
@@ -272,21 +114,21 @@ impl PartialEq for String {
...
     }
 }
 
 impl Ord for String {
-    fn cmp(&self, other: &String) -> Ordering {
-        self.buf[..].cmp(&other.buf[..])
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.inner.as_slice().cmp(other.inner.as_slice())
     }
 }
```
lopopolo: let's push these into the inner `EncodedString` too.

spinoso-string/src/lib.rs
lopopolo: I believe the `EncodedString` constructor can be `const` too. Can we try to preserve `const` here?

spinoso-string/src/lib.rs
lopopolo: Yup `set_encoding` will have to be reimplemented to `String#force_encoding`, but we can do that after landing this. I'll file a ticket.

spinoso-string/src/lib.rs
lopopolo: this is great. Thanks for adding this.

spinoso-string/src/lib.rs
lopopolo: yup that's right.

spinoso-string/src/lib.rs
```diff
@@ -1415,7 +1250,8 @@ impl String {
     /// [`String#b`]: https://ruby-doc.org/core-2.6.3/String.html#method-i-b
     #[inline]
     pub fn make_binary(&mut self) {
-        self.encoding = Encoding::Binary;
+        let old = self.inner.as_slice().to_vec();
+        self.inner = EncodedString::new(old, Encoding::Binary);
```
lopopolo, line 1206: I believe you mentioned that this implementation is incorrect given the docs for `String#b`? Do you remember the details enough to file a ticket?

spinoso-string/src/lib.rs
```diff
@@ -1440,149 +1276,47 @@ impl String {
...
     }
 
     #[inline]
     #[must_use]
     #[cfg(feature = "casecmp")]
     #[cfg_attr(feature = "docsrs", doc(cfg(feature = "casecmp")))]
     pub fn unicode_casecmp(&self, other: &String, options: CaseFold) -> Option<bool> {
-        let left = self.buf.as_slice();
+        let left = self;
         let right = other;
```
lopopolo: I think there's some deref and type inference magic happening here. Can we change these lines to the following to make it a bit more clear what is happening?

```suggestion
        let left = self.as_slice();
        let right = other.as_slice();
```

spinoso-string/src/lib.rs
lopopolo: 👍 Let's address this in a followup.

spinoso-string/src/lib.rs
```diff
@@ -2373,183 +2007,7 @@ impl String {
             _ => {}
         }
 
-        match self.encoding {
```
lopopolo, line 2376: yay for big red diffs 🔥

spinoso-string/src/lib.rs
lopopolo: I agree with your instinct here. It'd also allow us to only pay the branching cost once, for example we branch in `char_len` above in the short circuiting logic.

Happy for this to be addressed in a followup.

I think the short circuiting logic is most useful for UTF-8 case since that is the expensive one to compute.

spinoso-string/src/lib.rs
```diff
@@ -2596,38 +2054,10 @@ impl String {
...
-            Encoding::Utf8 => simdutf8::basic::from_utf8(&self.buf).is_ok(),
-            Encoding::Ascii => self.buf.is_ascii(),
-            Encoding::Binary => true,
-        }
+        self.inner.is_valid_encoding()
     }
 }
 
-#[must_use]
-fn conventionally_utf8_byte_string_len(mut bytes: &[u8]) -> usize {
```
lopopolo, line 2609: oh awesome! this moved directly to `Utf8String::char_len`

lopopolo (changes requested):
Ok this is my full review. Sorry it's taken me so long to get to it.

This PR is looking great! I have mostly style nits at this point and it looks like we have a lot of followup tickets to file.

I'll take filing the tickets (except the `String#b` one which I hope you can take since I don't remember the details).

This is close! I'm excited for this to land soon.

spinoso-string/src/lib.rs
b-n: I think I mostly follow what you mean :grimacing:. By pushing this down to `EncodedString`, are you meaning `impl PartialEq for EncodedString` (e.g. not on the layer further down right?) and then just using the redirected `as_slice()` methods?

Just checking my understanding on that last part - `rather than an extra indirection with an additional ref`. What happens when comparing two `&[u8]`'s ? I'm not sure I following with the statement `extra indirection with an additional ref` part :sweat_smile:

spinoso-string/src/lib.rs
b-n: not only in the `char_len`, but also in `len()` too (double cost). So yeah, #future for sure. It feels like `String` in an ideal world would just redirect all calls to `EncodedString`, and if the logic is duplicated in the encoded type, then so be it. This will be much easier/clearer in a follow up PR though as you said :+1:

spinoso-string/src/enc/ascii/impls.rs
b-n: Ooo, good catch. Now I'm super curious how rust even knows this. e.g. `to_vec` could be implemented on any type, so does it just check the `Deref::Target` to see if it implements something that has a `to_vec` function, and then uses that if all else fails? Or am I missing something here?

spinoso-string/src/enc/mod.rs
b-n: This is a very good thing to be nit-picky about. As I mentioned on discord, I have 0 ideas why I used `n` instead of something like `inner` or a shortened version like `i` or `e` etc haha. Easy update!

spinoso-string/src/enc/utf8/mod.rs
b-n: Oof, This was no doubt accidentally included. thanks for catching!

spinoso-string/src/enc/mod.rs
b-n: Update: Ignore the below. You were one step ahead :sweat_smile:.

~~Rust seems to not complain about just using the default impl traits here (e.g. `#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]`). From my understanding, this only works since I added the same to the encoded types - which works because `Vec<u8>` support these traits too. Is there any advantages to doing a custom `impl` on these are redirecting to `as_slice()` instead or is the default traits good enough?~~

spinoso-string/src/enc/utf8/mod.rs
b-n: Added as a TODO too

spinoso-string/src/enc/utf8/mod.rs
b-n: I assume you still wanted the first line of the comment, however let me know if you want that gone too!

spinoso-string/src/lib.rs
b-n: Here's a fun question. Should the `EncodedString` be housing the logic of calling `as_slice()` or redirecting to the encoded implementation of said trait.

e.g.
```rust
impl Hash for EncodedString {
    fn hash<H: Hasher>(&self, hasher: &mut H) {
        match self {
            EncodedString::Ascii(inner) => inner.hash(hasher),
            EncodedString::Binary(inner) => inner.hash(hasher),
            EncodedString::Utf8(inner) => inner.hash(hasher),
        }
    }
}
```

or

```rust
impl Hash for EncodedString {
    fn hash<H: Hasher>(&self, hasher: &mut H) {
        self.as_slice().hash(hasher);
    }
}
```

The original comment for Hash is talking about only comparing the byte vectors directly - which leads me a bit more to the latter `impl` (and what i have implemented) - however the first `impl` puts it in the encoded implementations hands "define your own hash" kind of thing - but that feels not what ruby is really trying to do here. :thinking:

spinoso-string/src/lib.rs
b-n: Yes, I'm hoping this is the issue I was thinking of:

```irb
irb(main):001:0> a = "a"
=> "a"
irb(main):002:0> b = a.b
=> "a"
irb(main):003:0> a.object_id
=> 18440
irb(main):004:0> b.object_id
=> 25120
```

Specifically: https://ruby-doc.org/core-3.1.1/String.html#method-i-b <-- Says it'll return a copied string, where as this Rust implementation is just changing the encoding still. If that still feels like the implementation isn't "correct", then I'll raise the issue!

b-n:
@lopopolo I think I've parsed all the feedback. There are some questions above, but honestly, it's just a flavour/taste thing, I believe everything actually works now!

One thing though, the `set_encoding` method disappearing feels like a breaking change. The question is, should this mean spinoso-string `1.0.0`, or just bump to `0.12.0` (I assume the latter because of the warnings of "don't use in prod" etc). Also, is it just as simple as bumping `spinoso-string` and changing the dep version in `artichoke-backend` to match? (e.g. no other publishing or particular order of things that need to happen?)

spinoso-string/src/lib.rs
lopopolo: yup yup I like the way you've done this. thank you for being so thoughtful as we work through the API surface.

spinoso-string/src/lib.rs
```diff
@@ -303,17 +97,18 @@ impl String {
     /// ```
     /// use spinoso_string::{Encoding, String};
     ///
-    /// const S: String = String::new();
+    /// let S: String = String::new();
```
lopopolo:
```suggestion
    /// const S: String = String::new();
```

spinoso-string/src/lib.rs
```diff
@@ -303,17 +97,18 @@ impl String {
...
     /// assert_eq!(S.encoding(), Encoding::Utf8);
     /// ```
     ///
     /// [conventionally UTF-8]: crate::Encoding::Utf8
     #[inline]
     #[must_use]
-    pub const fn new() -> Self {
+    pub fn new() -> Self {
```
lopopolo:
```suggestion
    pub const fn new() -> Self {
```

spinoso-string/src/lib.rs
```diff
@@ -362,8 +157,9 @@ impl String {
     #[must_use]
     pub fn with_capacity(capacity: usize) -> Self {
         let buf = Vec::with_capacity(capacity);
-        let encoding = Encoding::Utf8;
-        Self { buf, encoding }
+        Self {
+            inner: EncodedString::new(buf, Encoding::Utf8),
```
lopopolo, line 161: As a nit, I'd rather add a `EncodedString::utf8` constructor, since we pay to branch on the encoding in `EncodedString::new` when we don't need to.

spinoso-string/src/lib.rs
lopopolo: Thank you! See:

- https://github.com/artichoke/artichoke/issues/1720

spinoso-string/src/lib.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/1719

spinoso-string/src/lib.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/1721

spinoso-string/src/lib.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/1722

spinoso-string/src/enc/utf8/mod.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/1723

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -0,0 +1,1049 @@
...
+        // If the string is either all ASCII or all ASCII for a prefix
+        // of the string that contains the range we wish to slice,
+        // fallback to byte slicing as in the ASCII and binary fast path.
+        let consumed = match self.inner.find_non_ascii_byte() {
+            None => return self.inner.get(index..=index),
+            Some(idx) if idx > index => return self.inner.get(index..=index),
+            Some(idx) => idx,
+        };
+        let mut slice = &self.inner[consumed..];
+        // TODO: See if we can use `get_unchcked` as implmented in`fn char_len`
```
lopopolo:
```suggestion
        // TODO: See if we can use `get_unchecked` as implemented in`fn char_len`
```

spinoso-string/src/enc/utf8/impls.rs
```diff
@@ -0,0 +1,181 @@
...
+#[cfg(feature = "std")]
+use core::fmt;
+use core::ops::{Deref, DerefMut};
+#[cfg(feature = "std")]
+use std::io::{IoSlice, Result, Write};
+
+use super::Utf8String;
+
+#[cfg(feature = "std")]
+impl Write for Utf8String {
```
lopopolo, line 14: Split `io::Write` implementations for all encoded string types into an `io` module

This will make the feature flagging a bit clearer.

lopopolo (approved):
Woohoo! THANK YOU SO MUCH! 🚀

spinoso-string/src/enc/utf8/impls.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/1724

spinoso-string/src/lib.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/1725

spinoso-string/src/enc/mod.rs
```diff
@@ -0,0 +1,512 @@
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.as_slice().cmp(other.as_slice())
+    }
+}
+
+// Constructors
+impl EncodedString {
+    #[inline]
+    #[must_use]
+    pub fn new(buf: Vec<u8>, encoding: Encoding) -> Self {
```
lopopolo:
```suggestion
    pub const fn new(buf: Vec<u8>, encoding: Encoding) -> Self {
```

lopopolo:
> One thing though, the `set_encoding` method disappearing feels like a breaking change. The question is, should this mean spinoso-string `1.0.0`, or just bump to `0.12.0` (I assume the latter because of the warnings of "don't use in prod" etc). Also, is it just as simple as bumping `spinoso-string` and changing the dep version in `artichoke-backend` to match? (e.g. no other publishing or particular order of things that need to happen?)

Yes please let's bump to `0.12.0`. To do that:

- Bump the `version` field in `spinoso-string/Cargo.toml` to `0.12.0`.
- Update the usage section in `spinoso-string/README.md` to `0.12.0`.
- Update the dependency declaration in `artichoke-backend/Cargo.toml` to `0.12.0`.
- Run `bundle exec rake build:all` and check in all of the updated `Cargo.lock` files (there should be 4 that change).

It looks like that's the last bit that needs to get added to this branch before landing this PR. I eagerly await a commit from you to bump the version to `0.12.0` and then I'll merge this branch.

spinoso-string/src/encoded_string.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/1726

lopopolo:
> * `chars` and `codepoints` can likely use a more generic iterator now. They still have switches on `Encoding`, but I believe these should be able to be implemented directly on the encoded types.

Filed:

- https://github.com/artichoke/artichoke/issues/1727
- https://github.com/artichoke/artichoke/issues/1728

b-n:
@lopopolo 0d9cc75 contains the version bump! (funnily enough, there isn't a lock file per sub module, so it's only one lock file, not 4 - unless I missed something of course!)

🚀  This has been a fun piece of work and I've learnt a lot - thank you so much for your patience with my learning process here too!

lopopolo:
Thank you @b-n! So happy to see this merged into trunk.

PR #1753, opened 2022-03-23
Upgrade Rust toolchain to 1.59.0

lopopolo:


lopopolo:
I'm looking to use destructuring assignment (which was stabilized in 1.59.0: https://github.com/rust-lang/rust/issues/71126) in a branch I'm working on.

Given that upstream reports this fixed by an XCode update from Apple, removing WIP and I'll get this ready to merge.

PR #1756, opened 2022-03-27
Remove `ruby` dependency from `artichoke-backend` build script

lopopolo:
`artichoke-backend` crate has a `build.rs` script that builds a static library for mruby and `artichoke-backend`'s `mruby-sys` C extension.

mruby's build process is driven by Ruby and Rake. This build script dynamically generates shims for loading mrbgems into the mruby interpreter.

This PR removes the Rake invocation and essentially checks in the dynamically code in a new `mrbgems` C extension.

This PR significantly refactors the build script. All C sources and header include dirs are explicitly enumerated in the source and yielded to `cc` build configuration via iterators.

The `mruby-sys` C extension has been exploded and moved. All C extensions live in `artichoke-backend/cext`. To avoid clobbering C sources from mruby with identically named sources in mrbgems, mruby, mrgems, and mrbsys static libraries are all built separately. Artichoke already compiles and links with LTO enabled, so hopefully this doesn't result in perf regressions.

The header that drives bindgen is now standalone and aggregates all cexts + mruby.

Update `CONTRIBUTING.md` and `BUILD.md` to reflect that Ruby is no longer required to build Artichoke.

On my machine, this PR reduces debug build time from 28s to 24s.

TODO:

- Remove the MRI build step from the nightly Dockerfiles: https://github.com/artichoke/docker-artichoke-nightly/issues/82
- Document C extensions by adding a README to each cext directory.


lopopolo:
The approach I've taken with this PR comes from what I understand about Bazel, whose approach to codegen during build time is basically "don't; do codegen separately and check it in", so that's what the `mrbgems` cext is.

PR #1764, opened 2022-03-28
Optimize bounds checks in `math::gamma` for integer-ish args

lopopolo:
`math::gamma` looks up items in a fixed array for int values as a fast
path. The prior code used a `len()` check in a match guard and then did
a potentially panicking `[]` index.

Coalesce these two bounds checks into one with `[f64; N]::get`. This
results in the fast path being much more obvious since it does an early
return.


lopopolo:
I discovered this by running `rake lint:clippy:restriction`.

PR #1771, opened 2022-03-30
Replace exhaustive `match` with a lookup table in `hex_escape`

lopopolo:
Replace the 250+ line exhaustive `match` expression in `Literal::hex_escape` with an exhaustive `[&str; 256]` const lookup table. This should result in similar codegen, but is much easier to read. This also removes an `allow(clippy::to_many_lines)`.


lopopolo:
I first discovered this approach by peeking at the `bstr` source for `ByteSlice::find_not_byteset` and have used this same approach in `boba` in https://github.com/artichoke/boba/pull/138.

PR #1772, opened 2022-03-30
Replace exhaustive `match` with lookup table in escape literal

lopopolo:
Replace the exhaustive `match` expression over `u8` with a `[&str; 256]` lookup table. This should result in similar codegen, but is much easier to read. This also removes two `allow(clippy::too_many_lines)`.

See also #1771.


lopopolo:
I first discovered this approach by peeking at the `bstr` source for `ByteSlice::find_not_byteset` and have used this same approach in `boba` in https://github.com/artichoke/boba/pull/138.

Issue #1733, opened 2022-03-14
add readmes to all top level directories


lopopolo:
I think some workspace crates are missing them

lopopolo:
After https://github.com/artichoke/artichoke/pull/1776 lands, the only directories that are missing READMEs are:

- `artichoke-load-path` (pending removal once I can bring `mezzaluna-feature-loader` to parity)
- `mezzaluna-feature-loader`
- `ui-tests`

lopopolo:
For `ui-tests`, I think the README should looks something like:

- What this workspace is
- How to run the tests
- Do you need to `cargo install cargo-insta`?
- How to add a new test (and why you'd add one here)

PR #1810, opened 2022-04-02
Information around ui-tests, writing, and running

b-n:
Relates to #1733 

Specifically, adding some documentation around using ui-tests, what it tries to achieve, and how to add new tests.

I explicitly tried to avoid adding any "how to use insta" comments since that's mostly documented on their page/site etc. I took some inspiration from `artichoke/fuzz` also.

If you want me to be more/or less specific, fire away!


lopopolo (approved):
thanks @b-n. I pushed a couple of tweaks to the language and ran a `rake fmt:text`. LGTM

PR #1821, opened 2022-04-05
Attempt to fix LeakSan linker errors in CI

lopopolo:
From CI logs, it looks like leaksan isn't liking resolving symbols from libmruby.a for
libmrbgems.a, specifically in the parser and codegen.

Move those sources to libmruby.a.

build logs:

```
  = note: /usr/bin/ld: /home/runner/work/artichoke/artichoke/target/x86_64-unknown-linux-gnu/debug/build/artichoke-backend-dbabfb710aa0de76/out/libmrbgems.a(y.tab.o): in function `parser_palloc':
          /home/runner/work/artichoke/artichoke/artichoke-backend/mrbgems/mruby-compiler/core/parse.y:106: undefined reference to `mrb_pool_alloc'
          /usr/bin/ld: /home/runner/work/artichoke/artichoke/target/x86_64-unknown-linux-gnu/debug/build/artichoke-backend-dbabfb710aa0de76/out/libmrbgems.a(y.tab.o): in function `composite_string_node':
          /home/runner/work/artichoke/artichoke/artichoke-backend/mrbgems/mruby-compiler/core/parse.y:1027: undefined reference to `mrb_pool_realloc'
          /usr/bin/ld: /home/runner/work/artichoke/artichoke/target/x86_64-unknown-linux-gnu/debug/build/artichoke-backend-dbabfb710aa0de76/out/libmrbgems.a(y.tab.o): in function `mrb_parser_new':
          /home/runner/work/artichoke/artichoke/artichoke-backend/mrbgems/mruby-compiler/core/parse.y:6571: undefined reference to `mrb_pool_open'
          /usr/bin/ld: /home/runner/work/artichoke/artichoke/artichoke-backend/mrbgems/mruby-compiler/core/parse.y:6573: undefined reference to `mrb_pool_alloc'
          /usr/bin/ld: /home/runner/work/artichoke/artichoke/target/x86_64-unknown-linux-gnu/debug/build/artichoke-backend-dbabfb710aa0de76/out/libmrbgems.a(y.tab.o): in function `mrb_parser_free':
          /home/runner/work/artichoke/artichoke/artichoke-backend/mrbgems/mruby-compiler/core/parse.y:6613: undefined reference to `mrb_pool_close'
          /usr/bin/ld: /home/runner/work/artichoke/artichoke/target/x86_64-unknown-linux-gnu/debug/build/artichoke-backend-dbabfb710aa0de76/out/libmrbgems.a(y.tab.o): in function `mrb_load_exec':
[...]

PR #1822, opened 2022-04-05
Attempt to fix LeakSan linker errors in CI

lopopolo:
From CI logs, it looks like leaksan isn't liking resolving symbols from libmruby.a for
libmrbgems.a, specifically in the parser and codegen.

Use relative paths when feeding files to `cc` so all C sources, even if
they have the same filename, can be built into a single static lib.

When given a relative path, `cc` will preserve directory structure
rather than simply copying the filename to the compile directory.

This is a better way to fix #1821.

### DIFF
--- a/artichoke-backend/build.rs
+++ b/artichoke-backend/build.rs
@@ -207,6 +207,11 @@ mod libs {
 
         for source in sources {
             println!("cargo:rerun-if-changed={}", source.to_str().unwrap());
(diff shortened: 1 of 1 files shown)


lopopolo:
woohoo! trunk is green https://github.com/artichoke/artichoke/runs/5826491954?check_suite_focus=true

PR #1630, opened 2022-01-02
Implement `strscan` in Ruby with byte-wise `String` operations

lopopolo:
These changes are enabled by the spinoso-string implementation of String in Ruby Core.

Obsoletes:

- https://github.com/artichoke/artichoke/pull/1414

Fixes #135.

PR #1823, opened 2022-04-05
Encoding-aware `String#inspect`

lopopolo:
Add encoding support to `String#inspect` and the iterators that implement it.

Binary and ASCII strings always treat non-ASCII bytes as `\xYY` escapes.

This does not address https://github.com/artichoke/artichoke/issues/1350, but moves closer to being able to fix this because UTF-8 iter is separate from bytes as hex codes in ASCII and Binary.

Screen Shot 2022-04-04 at 9 52 30 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -697,7 +697,7 @@ dependencies = [
 
 [[package]]
 name = "spinoso-string"
-version = "0.14.0"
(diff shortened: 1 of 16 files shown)


lopopolo:
unblocks https://github.com/artichoke/artichoke/pull/1630

PR #1824, opened 2022-04-06
Align spinoso-string#make_binary to be more alike the String#b

### LINKED ISSUE #1720: `String::make_binary` should clone instead of mutate; rename to `String::to_binary`
Yes, I'm hoping this is the issue I was thinking of:

```irb
irb(main):001:0> a = "a"
=> "a"
irb(main):002:0> b = a.b
=> "a"
irb(main):003:0> a.object_id
=> 18440
irb(main):004:0> b.object_id
=> 25120
```

Specifically: https://ruby-doc.org/core-3.1.1/String.html#method-i-b <-- Says it'll return a copied string, where as this Rust implementation is just changing the encoding still. If that still feels like the implementation isn't "correct", then I'll raise the issue!

_Originally posted by @b-n in https://github.com/artichoke/artichoke/pull/1678#r820242310_

Also make changes to the implementation of `String#b` in `artichoke-backend` to do a clone and make a newly allocated `Value`. Relevant code:

https://github.com/artichoke/artichoke/blob/f7067c2dc0a80339dce95cd99b5ad6c7a27a89d3/artichoke-backend/src/extn/core/string/trampoline.rs#L374-L379

b-n:
Fixes #1720 

As talked about in the issue, the implementation in spinoso-string can be a little confusing when looking at how it is implemented in artichoke-backend. Specifically, spinoso-string is changing the encoding, however `String#b` is actually cloning the String before actually changing, so we might as well just do this directly and return a new value.

### DIFF
--- a/spinoso-string/src/lib.rs
+++ b/spinoso-string/src/lib.rs
@@ -1199,7 +1199,7 @@ impl String {
     /// assert_eq!(s.encoding(), Encoding::Utf8);
     /// let b = s.to_binary();
     /// assert_eq!(b.encoding(), Encoding::Binary);
-    /// assert_eq!(s.as_slice(), b.as_slice())
+    /// assert_eq!(s.as_slice(), b.as_slice());
     /// ```
     ///
     /// [`String#b`]: https://ruby-doc.org/core-2.6.3/String.html#method-i-b
--- a/spinoso-string/src/lib.rs
+++ b/spinoso-string/src/lib.rs
@@ -1186,7 +1186,7 @@ impl String {
         self.inner.is_ascii_only()
     }
 
-    /// Change the [encoding] of this `String` to [`Encoding::Binary`].
+    /// Return a newly encoded `String` with [`Encoding::Binary`] [encoding].
     ///
     /// This function can be used to implement the Ruby method [`String#b`].
     ///
@@ -1195,18 +1195,18 @@ impl String {
     /// ```
     /// use spinoso_string::{Encoding, String};
     ///
-    /// let mut s = String::utf8(b"xyz".to_vec());
+    /// let s = String::utf8(b"xyz".to_vec());
     /// assert_eq!(s.encoding(), Encoding::Utf8);
-    /// s.make_binary();
-    /// assert_eq!(s.encoding(), Encoding::Binary);
+    /// let b = s.to_binary();
+    /// assert_eq!(b.encoding(), Encoding::Binary);
     /// ```
     ///
     /// [encoding]: crate::Encoding
     /// [`String#b`]: https://ruby-doc.org/core-2.6.3/String.html#method-i-b
     #[inline]
-    pub fn make_binary(&mut self) {
-        let old = self.inner.as_slice().to_vec();
-        self.inner = EncodedString::new(old, Encoding::Binary);
+    #[must_use]
+    pub fn to_binary(&self) -> Self {
+        String::binary(self.inner.as_slice().to_vec())
     }
 
     /// Returns the length of this `String` in bytes.
--- a/spinoso-string/src/lib.rs
+++ b/spinoso-string/src/lib.rs
@@ -1186,7 +1186,7 @@ impl String {
         self.inner.is_ascii_only()
     }
 
-    /// Return a newly encoded `String` with [`Encoding::Binary`] [encoding].
+    /// Return a newly encoded `String` with [`Encoding::Binary`] encoding.
     ///
     /// This function can be used to implement the Ruby method [`String#b`].
     ///
@@ -1199,9 +1199,9 @@ impl String {
     /// assert_eq!(s.encoding(), Encoding::Utf8);
     /// let b = s.to_binary();
     /// assert_eq!(b.encoding(), Encoding::Binary);
+    /// assert_eq!(s.as_slice(), b.as_slice())
     /// ```
     ///
-    /// [encoding]: crate::Encoding
     /// [`String#b`]: https://ruby-doc.org/core-2.6.3/String.html#method-i-b
     #[inline]
     #[must_use]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -697,7 +697,7 @@ dependencies = [
 
 [[package]]
 name = "spinoso-string"
-version = "0.15.0"
+version = "0.16.0"
 dependencies = [
  "bstr",
  "bytecount",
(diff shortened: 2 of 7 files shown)


spinoso-string/src/lib.rs
```diff
@@ -1195,18 +1195,18 @@ impl String {
     /// ```
     /// use spinoso_string::{Encoding, String};
     ///
-    /// let mut s = String::utf8(b"xyz".to_vec());
+    /// let s = String::utf8(b"xyz".to_vec());
     /// assert_eq!(s.encoding(), Encoding::Utf8);
-    /// s.make_binary();
-    /// assert_eq!(s.encoding(), Encoding::Binary);
+    /// let b = s.to_binary();
+    /// assert_eq!(b.encoding(), Encoding::Binary);
```
lopopolo, line 1201: Can we add another line to this doc test that asserts the byte content of the e old and new strings are the same?

lopopolo (changes requested):
Looks good, I'd love to see one additional assert in the doc tests now that we're cloning the bytes

spinoso-string/src/lib.rs
b-n: What's your favourite way to check bytes in this case? I imagine there's many ways:
- `assert_eq!(s, b)` works because the EncodedString PartialOrd is comparing the slices
- `assert_eq(*s.as_slice(), *b.as_slice())` to me is more explicit and doesn't rely on the above knowledge

lopopolo: I think you're right. I like the slice way better since it doesn't rely on folks to know the partial eq behavior of `String`.

When you add this, can you remove the prefix `*`. It'll compile fine without it and it will make the test a bit clearer to read

b-n:
@lopopolo I believe it's unrelated to this specific change, but am I correct in assuming we can't actually check anything with relation to encoding on the `String` ruby class? https://github.com/artichoke/artichoke/blob/trunk/artichoke-backend/src/extn/core/string/string.rb#L514

Specifically, trying to test this above change like so:

```
artichoke 0.1.0-pre.0 (2022-04-02 revision 5484) [x86_64-unknown-linux-gnu]
[rustc 1.58.1 (db9d1b20b 2022-01-20) on x86_64-unknown-linux-gnu]
>>> "a".b.encoding
=> #<Encoding:UTF-8>
```

^-- not what I was expecting, and dug into `String#encode` and found that statement.

spinoso-string/src/lib.rs
b-n: Just a newb question for clarity, is rust doing automatic deref in these cases?

Thought process: `as_slice` returns `&[u8]`, "I'll need to deref that surely"

lopopolo: I think it actually impls partial eq for the slice type with the reference

lopopolo:
Yea `Encoding` as a Ruby object isn't exposed from Artichoke right now. The comments in the Ruby stubs appear to be outdated tho.

Artichoke does have an `Encoding` class, but it is essentially a stub to get Ruby spec to run. It so not wired up to the `String` internals.

I should file some tickets.

b-n:
Would you believe that Documentation Workflow error run can't be reproduced locally?

```shell
➜  artichoke git:(make-binary-to-binary) rustup -Vv    
rustup 1.24.3 (ce5817a94 2021-05-31)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.58.1 (db9d1b20b 2022-01-20)`
➜  artichoke git:(make-binary-to-binary) rustc -Vv 
rustc 1.58.1 (db9d1b20b 2022-01-20)
binary: rustc
commit-hash: db9d1b20bba1968c1ec1fc49616d4742c1725b4b
commit-date: 2022-01-20
host: x86_64-unknown-linux-gnu
release: 1.58.1
LLVM version: 13.0.0
➜  artichoke git:(make-binary-to-binary) cargo version --verbose       
cargo 1.58.0 (f01b232bc 2022-01-19)
release: 1.58.0
commit-hash: f01b232bc7f4d94f0c4603930a5a96277715eb8c
commit-date: 2022-01-19
host: x86_64-unknown-linux-gnu
libgit2: 1.3.0 (sys:0.13.23 vendored)
libcurl: 7.80.0-DEV (sys:0.4.51+curl-7.80.0 vendored ssl:OpenSSL/1.1.1l)
os: Ubuntu 20.04 (focal) [64-bit]
```

And `cargo +nightly doc --workspace` appears to "just work". I see you had some failure of this recently, any tips?

lopopolo:
Oh sorry I ran into this in #1753 too. The current nightly rust doc has an ICE in spinoso-array. It's fixed in master. I'll run the workflow again tonight when the next nightly is out.

https://github.com/rust-lang/rust/issues/95712#issuecomment-1090020369

lopopolo:
> And `cargo +nightly doc --workspace` appears to "just work". I see you had some failure of this recently, any tips?

In case you are curious about this and how Artichoke uses nightly Rust, this command uses whatever version of nightly you happen to have installed. The breakage is only on the current nightly though, which you can get by running `rustup update nightly`.

`rake doc` and `rake fmt:rust` will automatically pull down a nightly if you don't have one, but they won't update it.

`rake doc` uses nightly rust because `docs.rs` renders docs with nightly and Artichoke makes use of some docsrs attributes which are only available on nightly like feature callouts:

https://github.com/artichoke/artichoke/blob/19ff1ef4ee73f3910b4c099bdc1195ba35fc52de/artichoke-core/src/lib.rs#L14-L19

`rake fmt:rust` uses nightly because this rustdoc config, which does intellij style import grouping, is nightly only:

https://github.com/artichoke/artichoke/blob/19ff1ef4ee73f3910b4c099bdc1195ba35fc52de/rustfmt.toml#L14-L15

In CI, Artichoke uses nightly for the fuzz workflows because:

- `cargo-fuzz` requires nightly.
- Using sanitizers like LeakSan requires passing an unstable `-Z sanitizer=leak` flag to `rustc`.

In CI, Artichoke uses nightly for the rustdoc workflow and for `cargo fmt` for the same reasons the `Rakefile` does.

spinoso-string/src/lib.rs
```diff
@@ -1195,18 +1195,18 @@ impl String {
...
+    /// assert_eq!(s.as_slice(), b.as_slice())
```
lopopolo:
```suggestion
    /// assert_eq!(s.as_slice(), b.as_slice());
```

lopopolo (approved):

lopopolo:
new nightly is out, I rebuilt the docs workflow and build is green. we're ready to merge

PR #1831, opened 2022-04-10
Use namespaced and weak dependency features in `Cargo.toml`

lopopolo:
This feature was rolled out in Rust 1.60:

- https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#new-syntax-for-cargo-features


lopopolo:
This PR intentionally skips `spinoso-random` because the features in that crate need a lot of work.

PR #1902, opened 2022-06-02
Add `tz-rs` backend to `spinoso-time`

### LINKED ISSUE #1833: Add a `tz-rs` backend to `spinoso-time`
- https://github.com/x-hgg-x/tz-rs
- https://github.com/Kijewski/tzdb

`tz-rs` describes itself as:

> A pure Rust reimplementation of libc functions [localtime](https://en.cppreference.com/w/c/chrono/localtime), [gmtime](https://en.cppreference.com/w/c/chrono/gmtime) and [mktime](https://en.cppreference.com/w/c/chrono/mktime).
>
> This crate allows to convert between a [Unix timestamp](https://en.wikipedia.org/wiki/Unix_time) and a calendar time exprimed in the [proleptic gregorian calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar) with a provided time zone.
>
> Time zones are provided to the library with a [POSIX TZ string](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html) which can be read from the environment.

This is basically exactly what is needed to implement the `Time` core class.

`tz-rs` combined with `tzdb` should make a full `Time` implementation easier to achieve. There are no panicking APIs which should help too.

Once this is rolled out, consider removing the `chrono` backend, which would allow addressing:

https://github.com/artichoke/artichoke/blob/914bea3c8209ff9ac3299151aa220cdf384afc27/deny.toml#L6-L8

b-n:
# Description

First cut/attempt at implementing a tz-rs backend for spinoso-time

I'm 90% sure this implements enough for 100% of ruby `Time` (except `strftime`) [.](https://www.youtube.com/watch?v=pjvQFtlNQ-M)

Note: There are some small differences in the chrono vs. tzrs implementations. Specifically I didn't add `is_monday()` functions, since they can just be derived from `day_of_week()` etc. Also some small naming differences in the parts. I tried to be consistent with ruby and idiomatic rust (instead of with chrono). I can fairly easily standarise these, but maybe a fun discussion in an issue for posterity instead?

TODO:
- Exceptions/Errors: (see https://github.com/b-n/artichoke/pull/171 for an intial attempt at this)
  - OutsideRange (e.g. 0..23 for hours when constructing)
  - ProjectionError - Essentially re-exporting the [`tz::error::ProjectDateTimeError`](https://docs.rs/tz-rs/0.6.9/tz/error/struct.ProjectDateTimeError.html). But first, check when these can happen, they might already be caught.
- Add a tz_by_name feature and leverage [`tzdb::tz_by_name`](https://docs.rs/tzdb/latest/tzdb/fn.tz_by_name.html)
- Add `Hash` impl for `spinoso_time::tzrs::Offset` and include this in `spinoso_time::tzrs::Time` Hash
- Add `strftime` functionality :grimacing:

Fixes: #1833


spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,222 @@
...
+    #[must_use]
+    pub fn fixed(offset: i32) -> Self {
+        let sign = if offset.is_negative() { '-' } else { '+' };
+        let hours = offset / 3600;
+        let minutes = (offset / 60) % 60;
+        let offset_name = format!("GMT {}{}{}", sign, hours, minutes);
+        let local_time_type =
+            LocalTimeType::new(offset, false, Some(offset_name.as_bytes())).expect("Couldn't create fixed offset");
+        Self::Fixed([local_time_type])
+    }
```
b-n, line 97: ~This should be correct, however writing a test for these values is difficult because of the whole borrowing temporary values etc.~

~This still needs some work, it should actually be doing what `Time::time_zone()` does~

Fixed

spinoso-time/src/time/tzrs/to_a.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// for the source _time_.
+    pub wday: u8,
+    /// An integer representing the day of the year, `1..=366` for the source
+    /// _time_.
+    pub yday: u16,
+    /// Whether the source _time_ occurs during Daylight Saving Time in its time
+    /// zone.
+    pub isdst: bool,
+    /// The timezone used for the source _time_.
+    pub zone: String,
```
b-n, line 31: Yes, an owned string, for a couple of reasons:

- `Offset::Fixed` generates a string for the `time_zone_designation` during runtime. I'm not sure if there's some fancy way to make this into a `&'static`. We do know that `time_zone_designation` is a fixed width u8 buffer, https://docs.rs/tz-rs/0.6.9/src/tz/timezone/mod.rs.html#83, so maybe there's some magic there?
- If we can't use `&'static`, then it's possible `Time` is dropped after creating a `ToA`, so it would make sense to own it instead of specifying a dependent lifetime

I'm all ears if there's another way to do this one of course

spinoso-time/src/time/tzrs/to_a.rs
```diff
@@ -0,0 +1,78 @@
...
+    ///
+    /// [sec, min, hour, day, month, year, wday, yday, isdst, zone]
+    pub const ELEMENTS: usize = 10;
+
+    /// A ten-element array of values for time:
+    ///
+    /// [sec, min, hour, day, month, year, wday, yday, isdst, zone]
+    #[inline]
+    #[must_use]
+    pub fn to_tuple(&self) -> (u8, u8, u8, u8, u8, i32, u8, u16, bool, String) {
```
b-n: This function isn't actually used anywhere from what I can see, I just copied from the chrono impl

lopopolo:
Hi @b-n thanks for the PR. I know it's been up for a bit and I'm sorry for not giving you a review yet. Posting here to let you know that I know this is pending my action and I will review it, but I'm not sure when I'll get to it. Thanks for your patience :)

Cargo.lock
```diff
@@ -511,15 +558,37 @@ version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4d684eb692d561551750a408e516759f550d51033179e7d2f676efc4495b3d4e"
 
+[[package]]
+name = "regex"
+version = "0.2.11"
```
lopopolo: 👀 why did we pull in an ancient version of `regex`? 1.5.5 is already in the lockfile.

lopopolo: Did `cargo deny` not catch this duplicate crate version?

Cargo.lock
```diff
@@ -164,6 +179,28 @@ dependencies = [
  "winapi",
 ]
 
+[[package]]
+name = "const_fn"
```
lopopolo, line 174: I believe that I intentionally disabled default features in `tz-rs` to avoid pulling in this proc macro. Did we need to undo that to maintain const-ness in some `spinoso-time` APIs?

Cargo.lock
```diff
@@ -235,6 +272,16 @@ version = "0.11.2"
...
 checksum = "ab5ef0d4909ef3724cc8cce6ccc8572c5c817592e9285f5464f8e86f8bd3726e"
 
+[[package]]
+name = "iana-time-zone"
+version = "0.1.33"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "df1b4d6691512a74af43bb6de0b7f546fbe70bfa3bac84da8454a00121e63201"
+dependencies = [
+ "core-foundation",
+ "winapi",
```
lopopolo, line 273: oh fun! this is a TODO for later, but `Time` has always been disabled in the Artichoke Playground due to platform specific things like this.

spinoso-time/Cargo.toml
```diff
@@ -13,9 +13,17 @@ license = "MIT"
...
+chrono = ["dep:chrono", "dep:chrono-tz"]
+tzrs = ["dep:lazy_static", "dep:regex", "dep:tz-rs", "dep:tzdb"]
+
 [dependencies]
-chrono = { version = "0.4.19", default-features = false, features = ["clock"] }
-chrono-tz = { version = "0.6.0", default-features = false }
+chrono = { version = "0.4.19", default-features = false, features = ["clock"], optional = true }
+chrono-tz = { version = "0.6.0", default-features = false, optional = true }
+lazy_static = { version = "1.4.0", optional = true }
+regex =  { version = "0.2", optional = true }
```
lopopolo: Can we update this to depend on `regex@1.5.5` like the rest of the workspace, maybe with a minimal set of features like this dep in `spinoso-regexp`?

https://github.com/artichoke/artichoke/blob/06e9e0d88dc7d7451dc5b2269e9d05c69bef5da6/spinoso-regexp/Cargo.toml#L25

spinoso-time/Cargo.toml
```diff
@@ -13,9 +13,17 @@ license = "MIT"
...
+[features]
+chrono = ["dep:chrono", "dep:chrono-tz"]
+tzrs = ["dep:lazy_static", "dep:regex", "dep:tz-rs", "dep:tzdb"]
+
 [dependencies]
-chrono = { version = "0.4.19", default-features = false, features = ["clock"] }
-chrono-tz = { version = "0.6.0", default-features = false }
+chrono = { version = "0.4.19", default-features = false, features = ["clock"], optional = true }
+chrono-tz = { version = "0.6.0", default-features = false, optional = true }
+lazy_static = { version = "1.4.0", optional = true }
```
lopopolo: Artichoke already has `once_cell` as a direct dependency. I'd prefer to use that if we can.

It looks like `lazy_static` is also in the lockfile but only as a transitive dependency of the optional `regexp-oniguruma` feature.

spinoso-time/Cargo.toml
```diff
@@ -13,9 +13,17 @@ license = "MIT"
 keywords = ["artichoke", "date", "datetime", "spinoso", "time"]
 categories = ["date-and-time"]
 
+[features]
```
lopopolo, line 16: Can we enable both backends by default like the rest of the spinoso crates do?

```suggestion
[features]
default-features = ["chrono", "tzrs"]
```

e.g. `spinoso-array`:

https://github.com/artichoke/artichoke/blob/06e9e0d88dc7d7451dc5b2269e9d05c69bef5da6/spinoso-array/Cargo.toml#L26

typically, spinoso crates have a "default" backend that is always available even with `default-features = false`, but that seems fine to not follow here since this crate is still in flux.

spinoso-time/Cargo.toml
```diff
@@ -13,9 +13,17 @@ license = "MIT"
...
+chrono = ["dep:chrono", "dep:chrono-tz"]
+tzrs = ["dep:lazy_static", "dep:regex", "dep:tz-rs", "dep:tzdb"]
```
lopopolo: I love these `dep:` Cargo features!

spinoso-time/Cargo.toml
```diff
@@ -13,9 +13,17 @@ license = "MIT"
...
+
 [dependencies]
-chrono = { version = "0.4.19", default-features = false, features = ["clock"] }
-chrono-tz = { version = "0.6.0", default-features = false }
+chrono = { version = "0.4.19", default-features = false, features = ["clock"], optional = true }
+chrono-tz = { version = "0.6.0", default-features = false, optional = true }
+lazy_static = { version = "1.4.0", optional = true }
+regex =  { version = "0.2", optional = true }
+tz-rs = { version = "0.6.9", optional = true }
```
lopopolo: per my comment in the lockfile, do we need the `const_fn` feature from `tz-rs`?

spinoso-time/src/lib.rs
```diff
@@ -33,28 +33,35 @@
 //! Time.now
 //! ```
 //!
-//! This implementation of `Time` supports the system clock via the
-//! [`chrono`] crate.
+//! This implementation of `Time` is dependant on the selected feature. The `chrono` feature uses the [`chrono`] crate, and the `tzrs` feature uses the [`tzdb`] for getting the local timezone information, and combines with the [`tz-rs`] crate to generate the time.
```
lopopolo, line 36: can you hard wrap all of these doc comments at 80 characters?

spinoso-time/src/lib.rs
```diff
@@ -33,28 +33,35 @@
...
 //!
 //! [`Time`]: https://ruby-doc.org/core-2.6.3/Time.html
 //! [`chrono`]: https://crates.io/crates/chrono
+//! [`tz-rs`]: https://crates.io/crates/tz-rs
+//! [`tzdb`]: https://crates.io/crates/tzdb
 
-// Ensure code blocks in `README.md` compile
-#[cfg(doctest)]
-#[doc = include_str!("../README.md")]
-mod readme {}
```
lopopolo: we can add this back with a bit of `cfg` magic:

```rust
#[cfg(all(doctest, feature = "chrono"))]
#[doc = include_str!("../README.md")]
mod readme {}
```

spinoso-time/src/lib.rs
```diff
@@ -63,69 +70,3 @@ pub const NANOS_IN_SECOND: u32 = Duration::from_secs(1).as_nanos() as u32;
...
-///     wday: 0,
-///     yday: 96,
-///     isdst: true,
-///     zone: Offset::Local,
-/// };
-/// let time = Time::try_from(to_a);
-/// assert_eq!(time, Err(ComponentOutOfRangeError::Time));
-/// ```
-#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
-pub enum ComponentOutOfRangeError {
```
lopopolo, line 115: qqq: where did this enum move to?

spinoso-time/src/time/chrono/error.rs
```diff
@@ -0,0 +1,72 @@
...
+/// };
+/// let time = Time::try_from(to_a);
+/// assert_eq!(time, Err(ComponentOutOfRangeError::Time));
+/// ```
+///
+/// [`Time`]: [`time::chrono::Time`]
+/// [`ToA`]: [`time::chrono::ToA`]
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
+#[allow(clippy::module_name_repetitions)]
+pub enum ComponentOutOfRangeError {
```
lopopolo, line 56: ahh I see it moved to the `time::chrono` module. Why are we not sharing this among backends? I don't think it is tied to the `chrono` implementation.

In practice, `artichoke-backend` implements `RubyException` for "foreign" error types from the spinoso crates, so having unified error types which are shared among all backends makes those downstream trait impls a bit easier to write.

spinoso-time/src/time/mod.rs
```diff
@@ -1,19 +1,29 @@
 //! Implementations of Ruby [`Time`], a timezone-aware datetime.
 //!
-//! This module contains implementations of a timestamp storage struct and
```
lopopolo: "this module" refers to `spinoso_time::time`, which contains several `time` implementations.

Is there a clearer way for us to say that? Does it make sense to revert this change?

spinoso-time/src/time/tzrs/build.rs
```diff
@@ -0,0 +1,35 @@
+use super::Offset;
+use super::Time;
+
+impl Time {
```
lopopolo, line 4: :chef-kiss:! The magic begins.

spinoso-time/src/time/tzrs/convert.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// use spinoso_time::tzrs::Time;
+    /// let now = Time::utc(2022, 05, 26, 13, 16, 22, 0);
+    /// assert_eq!(now.to_string(), "2022-05-26 13:16:22 UTC");
+    /// ```
+    ///
+    /// [`Time#asctime`]: https://ruby-doc.org/core-2.6.3/Time.html#method-i-asctime
+    /// [`Time#ctime`]: https://ruby-doc.org/core-2.6.3/Time.html#method-i-ctime
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        // TODO: future
+        //self.strftime("%Y-%m-%d %H:%M:%S %z")
```
lopopolo, line 23: let's turn this TODO into a ticket.

lopopolo: is this `strftime` format string the default `Time#to_s` output? If so that's amazing and I'd love to see it documented with some `irb` repl examples like:

https://github.com/artichoke/artichoke/blob/06e9e0d88dc7d7451dc5b2269e9d05c69bef5da6/artichoke-backend/src/extn/core/string/trampoline.rs#L125-L142

spinoso-time/src/time/tzrs/convert.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Formats _time_ according to the directives in the given format string.
+    ///
+    /// Can be used to implement [`Time#strftime`]
+    ///
+    /// [`Time#stftime`]: https://ruby-doc.org/core-2.6.3/Time.html#method-i-strftime
+    //#[inline]
+    //#[must_use]
+    //pub fn strftime(&self, _format: &str) -> String {
+    //todo!()
+    //}
```
lopopolo:
```suggestion
    pub fn strftime(&self, _format: &str) -> String {
        todo!("Not implemented GH-XXX");
    }
```

spinoso-time/src/time/tzrs/math.rs
```diff
@@ -0,0 +1,491 @@
...
+    fn add(self, duration: Duration) -> Self::Output {
+        let unix_time = self.inner.unix_time();
+        let nanoseconds = self.inner.nanoseconds();
+        let offset = self.offset;
+
+        let duration_seconds = i64::try_from(duration.as_secs()).expect("Duration too large");
+        let duration_subsecs = duration.subsec_nanos();
+
+        let mut seconds = unix_time.checked_add(duration_seconds).expect("Duration too large");
+        let mut nanoseconds = nanoseconds.checked_add(duration_subsecs).expect("Duration too large");
```
lopopolo, line 69: I think these `expect`s are similar to the ones in the `chrono` `ops::Add` impl. I kinda regret adding these since:

- they make fallibility hard (which we want so we can turn failures into at least a `RuntimeError`)
- I'm not super compelled to optimize for the ergonomics of `+` and `-` over explicit method calls since we control the only consumer of this code.

These are fine to leave in here for now, but I think I noticed you removed use of the `ops` overloads in the trampoline. In the long term, I'd like to remove the `math` impls from the `chrono` and `tzrs` impls.

spinoso-time/src/time/tzrs/math.rs
```diff
@@ -0,0 +1,491 @@
...
+
+        let mut seconds = unix_time.checked_sub(duration_seconds).expect("Duration too large");
+        let nanoseconds = if let Some(nanos) = nanoseconds.checked_sub(duration_subsecs) {
+            nanos
+        } else {
+            seconds -= 1;
+            nanoseconds + NANOS_IN_SECOND - duration_subsecs
+        };
+
+        Self::Output::with_timespec_and_offset(seconds, nanoseconds, offset)
```
lopopolo, line 194: Now that I think about it, maybe we keep these `ops` impls but have them delegate to a fallible method on `Time` and only unwrap here?

spinoso-time/src/time/tzrs/math.rs
```diff
@@ -0,0 +1,491 @@
...
+impl Sub<f64> for Time {
+    type Output = Self;
+
+    fn sub(self, seconds: f64) -> Self::Output {
+        self + -seconds
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
lopopolo, line 280: I love all these tests 😍

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+mod to_a;
+
+pub use offset::Offset;
+pub use to_a::ToA;
+
+use crate::NANOS_IN_SECOND;
+
+/// A wrapper around [`tz::datetime::DateTime`] which contains everything needed for date creation and
+/// conversion to match the ruby spec. Seconds and Subseconds are stored independently as i64 and
+/// u32 respectively, which gives enough granularity to meet the ruby [`Time`] spec.
```
lopopolo: can we hard wrap this at 80 chars too?

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+///
+/// [`Time`]: https://ruby-doc.org/core-2.6.3/Time.html
+#[must_use]
+#[derive(Debug, Copy, Clone)]
+pub struct Time {
+    /// A wrapper around [`tz::datetime::DateTime`] to provide date and time formatting
```
lopopolo:
```suggestion
    /// A wrapper around [`tz::datetime::DateTime`] to provide date and time formatting.
```

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+    inner: DateTime,
+    /// The offset to used for the provided _time_
```
lopopolo: sorry for these nits

```suggestion
    /// The offset to used for the provided _time_.
```

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+    offset: Offset,
+}
+
+impl Hash for Time {
+    #[inline]
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        state.write_i128(self.inner.total_nanoseconds());
```
lopopolo, line 38: this is amazing!

spinoso-time/src/time/chrono/error.rs
lopopolo: Ok I took a peek at https://github.com/b-n/artichoke/pull/171 and this makes much more sense. I retract my concern 😅

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+        day: u8,
+        hour: u8,
+        minute: u8,
+        second: u8,
+        nanoseconds: u32,
+        offset: Offset,
+    ) -> Self {
+        let tz = offset.time_zone_ref();
+        let found_date_times = DateTime::find(year, month, day, hour, minute, second, nanoseconds, tz)
+            .expect("Could not find a matching DateTime for this timezone");
```
lopopolo, line 95: I took a peek at https://github.com/b-n/artichoke/pull/171 and it looks like you have a plan for making these functions fallible and removing the `expect`s. Thank you!

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+mod tests {
+    use super::*;
+
+    fn time_with_fixed_offset(offset: i32) -> Time {
+        let offset = Offset::fixed(offset);
+        Time::with_timespec_and_offset(0, 0, offset)
+    }
+
+    #[test]
+    fn time_zone_fixed_offset() {
```
lopopolo, line 260: thank you so much for getting offsets working!

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+        Self::Tz(local_time_zone())
+    }
+
+    /// Generate an offset with a number of seconds from UTC.
+    #[inline]
+    #[must_use]
+    pub fn fixed(offset: i32) -> Self {
+        let offset_name = offset_hhmm_from_seconds(offset);
+        let local_time_type =
+            LocalTimeType::new(offset, false, Some(offset_name.as_bytes())).expect("Couldn't create fixed offset");
```
lopopolo, line 95: will this function become fallible once the error PR gets put up?

lopopolo: I'm thinking we'll need to handle scenarios like:

```
[3.1.2] > Time.new(2022, 3, 12, 15, 23, 59, -1_000_000)
<internal:timev>:310:in `initialize': utc_offset out of range (ArgumentError)
        from (irb):9:in `new'
        from (irb):9:in `<main>'
[3.1.2] > Time.new(2022, 3, 12, 15, 23, 59, 1_000_000)
<internal:timev>:310:in `initialize': utc_offset out of range (ArgumentError)
        from (irb):10:in `new'
        from (irb):10:in `<main>'
```

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// use spinoso_time::tzrs::Offset;
+    /// use tzdb::time_zone::pacific::AUCKLAND;
+    /// let offset = Offset::tz(AUCKLAND);
+    /// ```
+    ///
+    /// [`tz:timezone::TimeZoneRef`]: https://docs.rs/tz-rs/0.6.9/tz/timezone/struct.TimeZoneRef.html
+    /// [`tzdb`]: https://docs.rs/tzdb/latest/tzdb/index.html
+    #[inline]
+    #[must_use]
+    pub fn tz(tz: TimeZoneRef<'static>) -> Self {
```
lopopolo, line 114: if `TimeZoneRef` is part of `spinoso-time`'s public API, let's make sure we export the type from `spinoso_time::tzrs` with something like:

```rust
#[doc(inline)]
pub use tz::timezone::TimeZoneRef;
```

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+        Self::Tz(tz)
+    }
+
+    /// Returns a `TimeZoneRef` which can be used to generate and project _time_.
+    #[inline]
+    #[must_use]
+    pub fn time_zone_ref(&'a self) -> TimeZoneRef<'a> {
```
lopopolo: this signature can elide the lifetime label:

```suggestion
    pub fn time_zone_ref(&self) -> TimeZoneRef<'_> {
```

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// - `[+/-]HH[:]MM`
+    /// - A-I representing +01:00 to +09:00
+    /// - K-M representing +10:00 to +12:00
+    /// - N-Y representing -01:00 to -12:00
+    /// - Z representing 0 offset
+    ///
+    /// [accepted MRI values]: https://ruby-doc.org/core-2.6.3/Time.html#method-c-new
+    #[inline]
+    #[must_use]
+    fn from(input: &str) -> Self {
```
lopopolo, line 149: woohoo!

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+            "T" => Self::fixed(-7),
+            "U" => Self::fixed(-8),
+            "V" => Self::fixed(-9),
+            "W" => Self::fixed(-10),
+            "X" => Self::fixed(-11),
+            "Y" => Self::fixed(-12),
+            "Z" => Self::utc(),
+            _ => {
+                lazy_static! {
+                    static ref HH_MM_MATCHER: Regex = Regex::new(r"^([\-\+]{1})(\d{2})(\d{2})$").unwrap();
```
lopopolo: looking at the regex it looks like we only need to enable the `unicode-perl` and `std` features on `regex` in `spinoso-time`'s `Cargo.toml`.

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// Construct a Offset with the offset in seconds from UTC
+    #[inline]
+    #[must_use]
+    fn from(seconds: i32) -> Self {
+        Self::fixed(seconds)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
lopopolo, line 228: can we add sad path tests for when the offset string does not match the parsing regex?

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+                    let caps = HH_MM_MATCHER.captures(input).unwrap();
+
+                    let sign = if caps.get(1).unwrap().as_str() == "+" { 1 } else { -1 };
+                    let hours = caps.get(2).unwrap().as_str().parse::<i32>().unwrap();
+                    let minutes = caps.get(3).unwrap().as_str().parse::<i32>().unwrap();
+
+                    let offset_seconds: i32 = sign * ((hours * SECONDS_IN_HOUR) + (minutes * SECONDS_IN_MINUTE));
+                    Self::fixed(offset_seconds)
+                } else {
+                    // TODO: ArgumentError
```
lopopolo, line 189: that means this should probably be a `TryFrom` impl.

lopopolo: this is a place where some REPL snippets would be handy inline comments:

```console
[3.1.2] > Time.new(2022, 3, 12, 15, 23, 59, "dead")
<internal:timev>:310:in `initialize': "+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: dead (ArgumentError)
        from (irb):11:in `new'
        from (irb):11:in `<main>'
```

spinoso-time/src/time/tzrs/offset.rs
lopopolo: It looks like we're missing an optional `:` separator between `HH` and `MM`:

```console
[3.1.2] > Time.new(2022, 3, 12, 15, 23, 59, "+11:00")
=> 2022-03-12 15:23:59 +1100
[3.1.2] > Time.new(2022, 3, 12, 15, 23, 59, "+1100")
=> 2022-03-12 15:23:59 +1100
```

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+
+impl From<TimeZoneRef<'static>> for Offset {
+    #[inline]
+    #[must_use]
+    fn from(tz: TimeZoneRef<'static>) -> Self {
+        Self::tz(tz)
+    }
+}
+
+impl From<i32> for Offset {
```
lopopolo, line 218: Are all `i32` valid offsets or should this be a `TryFrom` impl?

Do the invariants of the `Self::Fixed` variant assume the `i32` is bounded?

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    #[test]
+    fn z_is_utc() {
+        let offset = Offset::from("Z");
+        assert!(matches!(offset, Offset::Utc));
+    }
+
+    #[test]
+    fn from_str_hh_mm() {
```
lopopolo, line 265: let's add a new test method that's basically a duplicate of this but with the colon separators.

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -0,0 +1,273 @@
...
+        assert_eq!(0, offset_seconds_from_fixed_offset("+0000"));
+        assert_eq!(0, offset_seconds_from_fixed_offset("-0000"));
+        assert_eq!(60, offset_seconds_from_fixed_offset("+0001"));
+        assert_eq!(-60, offset_seconds_from_fixed_offset("-0001"));
+        assert_eq!(3600, offset_seconds_from_fixed_offset("+0100"));
+        assert_eq!(-3600, offset_seconds_from_fixed_offset("-0100"));
+        assert_eq!(7320, offset_seconds_from_fixed_offset("+0202"));
+        assert_eq!(-7320, offset_seconds_from_fixed_offset("-0202"));
+        assert_eq!(362_340, offset_seconds_from_fixed_offset("+9999"));
+        assert_eq!(-362_340, offset_seconds_from_fixed_offset("-9999"));
```
lopopolo, line 275: It looks like MRI treats cases where `HH` or `MM` are out of range as errors.

```
[3.1.2] > Time.new(2022, 3, 12, 15, 23, 59, "+9900")
<internal:timev>:310:in `initialize': utc_offset out of range (ArgumentError)
        from (irb):14:in `new'
        from (irb):14:in `<main>'
[3.1.2] > Time.new(2022, 3, 12, 15, 23, 59, "+0199")
<internal:timev>:310:in `initialize': "+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: +0199 (ArgumentError)
        from (irb):15:in `new'
        from (irb):15:in `<main>'
```

spinoso-time/src/time/tzrs/parts.rs
```diff
@@ -0,0 +1,319 @@
+use super::Offset;
+use super::Time;
+use crate::MICROS_IN_NANO;
+
+// Parts
+impl Time {
+    /// Returns the number of nanoseconds for _time_
```
lopopolo: nit
```suggestion
    /// Returns the number of nanoseconds for _time_.
```

spinoso-time/src/time/tzrs/parts.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// let now = Time::utc(2022, 7, 8, 12, 34, 56, 0);
+    /// assert_eq!(now.day_of_week(), 5);
+    /// ```
+    ///
+    /// [`Time#wday`]: https://ruby-doc.org/core-2.6.3/Time.html#method-i-wday
+    #[inline]
+    #[must_use]
+    pub fn day_of_week(&self) -> u8 {
+        self.inner.week_day()
+    }
```
lopopolo, line 283: I implemented all of the `is_monday`, etc, helpers because I can never remember whether 0 is Sunday or Monday-indexed (and this actually varies by locale).

If we could add those back, I'd appreciate it.

spinoso-time/src/time/tzrs/timezone.rs
```diff
@@ -0,0 +1,155 @@
+use tz::datetime::DateTime;
+
+use super::{Offset, Time};
+
+// Timezone conversions (returns new Time)
+impl Time {
+    /// Returns a new Time object representing _time_ based on the provided offset
```
lopopolo: nit
```suggestion
    /// Returns a new Time object representing _time_ based on the provided offset.
```

spinoso-time/src/time/tzrs/timezone.rs
```diff
@@ -0,0 +1,155 @@
...
+    ///
+    /// #Examples
+    ///
+    /// ```
+    /// use spinoso_time::tzrs::Time;
+    /// let local_offset = Time::now().utc_offset();
+    /// let now_utc = Time::utc(2022, 7, 8, 12, 34, 56, 0);
+    /// let now_local = now_utc.to_local();
+    /// assert_eq!(now_local.utc_offset(), local_offset);
+    /// ```
```
lopopolo, line 62: this is a nice example that avoids test time indeterminism

spinoso-time/src/time/tzrs/timezone.rs
```diff
@@ -0,0 +1,155 @@
+use tz::datetime::DateTime;
+
+use super::{Offset, Time};
+
+// Timezone conversions (returns new Time)
```
lopopolo, line 5: thank you thank you a million times for figuring out offsets and timezones!

spinoso-time/src/time/tzrs/to_a.rs
lopopolo: We can use a `Cow<'static, str>` but I'm not sure it's worth the code complexity to save an allocation. What do you think @b-n?

spinoso-time/src/time/tzrs/to_a.rs
lopopolo: It should be used here, but not currently implemented:

https://github.com/artichoke/artichoke/blob/06e9e0d88dc7d7451dc5b2269e9d05c69bef5da6/artichoke-backend/src/extn/core/time/trampoline.rs#L174-L179

lopopolo (changes requested):
This is PR is an amazing bit of work. I left a bunch of comments (some of which are stream of consciousness as I went over the patch), but I think this PR is 90% of the way there.

Thanks so much and sorry for the delay in getting this reviewed. Looking forward to getting this merged.

lopopolo:
Oh interesting we got this warning from `cargo-deny`:

```
warning[A007]: advisory was not encountered
  ┌─ /home/runner/work/artichoke/artichoke/deny.toml:8:3
  │
8 │   "RUSTSEC-2020-0159",
  │   ^^^^^^^^^^^^^^^^^^^ no crate matched advisory criteria
 advisories ok: 0 errors, 1 warnings, 0 notes
       bans ok: 0 errors, 0 warnings, 0 notes
   licenses ok: 0 errors, 0 warnings, 102 notes
    sources ok: 0 errors, 0 warnings, 0 notes
```

probably from flipping off the `chrono` feature?

spinoso-time/src/time/tzrs/math.rs
b-n: I did borrow half of them from your chrono crate :sweat_smile: (The rounding ones etc are all new though :smile:)

spinoso-time/src/time/tzrs/mod.rs
b-n: :bow: 

Although, I would say that the timezone should also be in here somewhere. I think I've added it in the future commit actually, but I'll port it back here.

spinoso-time/src/time/tzrs/mod.rs
b-n: :100: correct. writing these `.expect`s hurt a little at the time, but the future glory will prevail!

spinoso-time/src/time/tzrs/mod.rs
b-n: You would not believe how much of a headache it was. This was iteration like 4 or 5 :grimacing:. I'm mostly certain this will be enough for a full `Time#at` impl.

spinoso-time/src/time/tzrs/offset.rs
b-n: Just wait until you realise that `Self::fixed(1)` gives a 1 second offset, not 1 hour offset. Just discovered this when attempting the `artichoke-backend` integration :grimacing:. I'll fix it here and fix the merge conflicts up the chain though :smile: 

Fixed now though :smile:

spinoso-time/src/time/tzrs/offset.rs
b-n: Just a quick thought - is including the whole regex package actually worth it for this one regex? I'm sure I could smash some rather ugly string manipulation code together for this same area, and it would be one less dependency :thinking:. I know artichoke-backend includes the `regex` crate, but it still feels a bit much for this piece of functionality

spinoso-time/src/time/tzrs/timezone.rs
b-n: I'm confident that my hair will eventually be grey later in life - and I think offset/timezones definitely got me a little closer to that haha.

spinoso-time/Cargo.toml
b-n: I have no idea what my brain was thinking that day, and I didn't look at it again after that. Good catch!

Cargo.lock
b-n: I think `cargo deny` caught it later when trying to integrate the `tz-rs` feature :grimacing:. But this is an easy fix!

Cargo.lock
b-n: Ah, I was meaning to come back and check which features to enable in `tz-rs` instead of just default. I don't think we need `const_fn` at all - but i'll double check

I checked - The functions we use in tz-rs are all `const_fn::const_fn` functions.

Cargo.lock
b-n: I'm 99% certain this crate is required because of `local()`. e.g. if we don't use `local()`, I think you get `Time` in wasm, but just without local timezone discovery.

I'll add a feature (`tzrs-local`?) around the function for getting the local timezone. Opinion: Should `local()` in an environment without the feature panic, or just return utc? :thinking:

spinoso-time/Cargo.toml
b-n: Ah yes - I'll replace with `once_cell`. I was surprised it wasn't somewhere in artichoke already, so I had a feeling you'd have something else similar elsewhere :smile: 

Done!

spinoso-time/Cargo.toml
b-n: Done! Assuming you did mean `default-features` and not just `default` right?

spinoso-time/src/lib.rs
b-n: Resolving because of later comment :smile:. I think it might still be possible to make and/or unify a bunch of errors from this crate. It feels like it'll be easier later - I'll create a follow up issue for it.

Might actually be solvable in the Error PR. https://github.com/b-n/artichoke/pull/171

spinoso-time/src/time/chrono/error.rs
b-n: As comment from `lib.rs`, I'm sure there are some commonalities in the errors. Creating an issue :smile:

spinoso-time/src/time/mod.rs
b-n: Ahhh I see. This was just my bad understanding of "this" - but reading it again actually makes sense with the prior text.

spinoso-time/src/time/tzrs/convert.rs
b-n: > is this strftime format string the default Time#to_s output?
Yep! (Or at least in the case of ruby yep!)

- https://ruby-doc.org/core-2.6.3/Time.html#method-i-inspect
- https://ruby-doc.org/core-2.6.3/Time.html#method-i-to_s

I'll throw it in the PR for integrating this into `artichoke-backend`

Note: I saw something somewhere about the to_s being locale specific - but I'm not sure where I saw that :thinking: 

Side note, apart form chrono, there doesn't appear to be any pure rust crate for strftime. I might take some inspiration from tz-rs + tzdb (for localisation). Would the artichoke org want to take on the package?

spinoso-time/src/time/tzrs/offset.rs
lopopolo: yea there are some hand-rolled parsers in Artichoke already, like the one for `Kernel#Integer`:

https://github.com/artichoke/artichoke/blob/06e9e0d88dc7d7451dc5b2269e9d05c69bef5da6/artichoke-backend/src/extn/core/kernel/integer.rs#L328

For now this seems fine, but if you'd like to revisit this, let's file a ticket.

spinoso-time/src/time/tzrs/math.rs
b-n: Yeah, I was also struggling with the overflow aspect of these ops and how to integrate them. I'll throw it into the errors PR (I've updated it as a TODO)

spinoso-time/src/time/tzrs/math.rs
b-n: I removed the`ops` overloads because they weren't working :grimacing: haha (I was actually going to ask you about that). I've added them as a TODO in the Errors PR though since the encapsulation will likely be a `TimeError` etc.

spinoso-time/src/time/tzrs/offset.rs
b-n: Good catch! and yes! Added to errors PR as a TODO.

It looks like MRI takes up to +/- 24 * 60 * 60 as a valid integer/float value. I'm mildly curious what it does with a float (e.g. nanos?), but let's see

spinoso-time/src/time/tzrs/offset.rs
b-n: I was thinking about this one. How do you feel about only exposing a `&[u8]` interface instead?

I should mention that MRI doesn't come with any timezone data (I believe it only knows about the local timezone via sys calls). Adding a timezone db functionality + Timezone by name is funnily enough more than what MRI supports. The reason for a &[u8] interface is to support a "lookup by name" approach.

Note: MRI also supports a ruby object with the relevant timezone timezone fields (too hard to support in spinoso-time, could be ruby functions etc)

spinoso-time/src/time/tzrs/offset.rs
b-n: I'm not super passionate about a non regex'd solution - but maybe one day for the fun if I remember instead of creating a ticket :smile:

spinoso-time/src/time/tzrs/offset.rs
b-n: Ah yes. I've got a local commit on my integration branch with this change. It's also passing a `&[u8]` instead since that's what's boxed. I'm leaving spinoso-time to convert that to a `&str` since Regex only likes `&str`s from my brief play

spinoso-time/src/time/tzrs/offset.rs
b-n: Hopefully answered in another comment above (e.g. range is +/- 24 * 60 * 60), but will address in errors PR since that has the Errors we want to throw.

spinoso-time/src/time/tzrs/offset.rs
b-n: I've put this into the errors PR too. Right now, if the regex fails it just throws back a UTC timezone which is incorrect at best. I've added some quick conditions that I've thought of, but please add some more if you've got some).

Note: Ruby 3 is much more flexible with these offsets. That'll be a fun migration hah.

spinoso-time/src/time/tzrs/offset.rs
b-n: Thrown into the Errors PR since we wrap the errors in that one.

spinoso-time/src/time/tzrs/timezone.rs
b-n: Thanks! Write once, run everywhere right? :sweat_smile:

spinoso-time/src/time/tzrs/to_a.rs
b-n: That would mean the functions that use this would have to check if it's a `Cow::Borrowed` or a `Cow::Owned` right? If so, I think that's unneeded complexity. Assuming 99% of the use cases are Ruby, then

```
irb(main):001:0> t = Time.now
=> 2022-06-25 20:25:09 +0200
irb(main):002:0> a = t.to_a
=> [9, 25, 20, 25, 6, 2022, 6, 176, true, "CEST"]
irb(main):003:0> a[9] = "HELLO"
=> "HELLO"
irb(main):004:0> t.zone
=> "CEST"
```

^-- e.g. we're note expecting `a[9]` to be it's own alloc'd value and not a ref.

spinoso-time/src/time/tzrs/to_a.rs
b-n: In which case, is there an easy tuple => array somewhere? I'd image this trampoline method would create an array from `crate::core::extn::array::Array`, but I'm not aware of any tuple => array magic that would make that easier than just assigning the values in order via the `pub` accessors

spinoso-time/src/time/tzrs/convert.rs
b-n: Created relevant issue: https://github.com/artichoke/artichoke/issues/1914

b-n:
@lopopolo I've parsed most of the feedback (still want to do the `once_cell`), but have left some comments unresolved where I think your input is needed.

Just checking, you you like to me to work on merging the errors PR into this work before merging into artichoke trunk? Doesn't bother me either way, but I'll likely get you to review the PRs in isolation to try and limit some of the cognitive load

lopopolo:
> Just checking, you you like to me to work on merging the errors PR into this work before merging into artichoke trunk? Doesn't bother me either way, but I'll likely get you to review the PRs in isolation to try and limit some of the cognitive load

No let's merge this PR first and then push the errors PR as a new patch. I agree it'll be easier to review these separately.

spinoso-time/src/time/mod.rs
```diff
@@ -5,15 +5,25 @@
 //! a timezone offset. These timestamps can be used to implement the Ruby `Time`
 //! core class.
 //!
-//! [`Time`](self::chrono::Time) is based on the [`chrono`] crate.
+//! There are two independent backends which can be selected by specifying the
+//! `chrono` or `tzrs` feature.
```
lopopolo:
```suggestion
//! `chrono` or `tzrs` feature:
```

spinoso-time/src/time/mod.rs
```diff
@@ -5,15 +5,25 @@
...
 //!
-//! The chrono backend stores datetimes as a `i64` [Unix timestamp], subsecond
-//! nanoseconds as a `u32`, and a [timezone offset] which can be one of several
+//! `chrono` is based on the [`chrono`] crate.
+//! `tzrs` is based on the [`tz-rs`] crate.
```
lopopolo:
```suggestion
//! - `chrono` is based on the [`chrono`] crate.
//! - `tzrs` is based on the [`tz-rs`] crate.
```

spinoso-time/src/time/tzrs/math.rs
lopopolo: sounds good. thanks

spinoso-time/src/time/tzrs/mod.rs
lopopolo: I just checked and it looks like the time zone is not part of the hash code:

```
[3.1.2] > t = Time.now
=> 2022-06-25 11:56:16.8661 -0700
[3.1.2] > t.zone
=> "PDT"
[3.1.2] > t.hash
=> 1699036252414506474
[3.1.2] > u = t.utc
=> 2022-06-25 18:56:16.8661 UTC
[3.1.2] > u.zone
=> "UTC"
[3.1.2] > u.hash
=> 1699036252414506474
```

spinoso-time/src/time/tzrs/offset.rs
lopopolo: Exposing a `&[u8]` interface works for me. If that's the case we'll want to rework the `Offset` enum to hide its guts:

```rust
/// Represents the number of seconds offset from UTC
pub struct Offset(OffsetType);

/// Represents the number of seconds offset from UTC
#[allow(variant_size_differences)]
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
enum OffsetType {
    /// UTC offset, zero offset, Zulu time
    Utc,
    /// Fixed offset from UTC
    ///
    /// Note: A fixed offset of 0 is different from UTC time
    Fixed([LocalTimeType; 1]),
    /// A time zone based offset
    Tz(TimeZoneRef<'static>),
}
```

This is kinda the same pattern:

https://github.com/artichoke/artichoke/blob/a9a3bb61351f6c81286e3122b0ab53f65907f6e1/spinoso-string/src/enc/inspect.rs#L7

https://github.com/artichoke/artichoke/blob/a9a3bb61351f6c81286e3122b0ab53f65907f6e1/spinoso-string/src/enc/inspect.rs#L30-L57

spinoso-time/src/time/tzrs/offset.rs
lopopolo: sounds good to defer this to the errors PR

spinoso-time/src/time/tzrs/to_a.rs
lopopolo: > That would mean the functions that use this would have to check if it's a Cow::Borrowed or a Cow::Owned right?

Not quite: `Cow<'static, str>` derefs to `&str` and can be turned into a `String` with `Cow::into_owned`.

Agreed though that this doesn't seem worth it.

spinoso-time/src/time/tzrs/to_a.rs
lopopolo: yes that makes sense.

we can nuke this crazy tuple method.

lopopolo (approved):

Cargo.lock
```diff
@@ -391,6 +391,12 @@ dependencies = [
  "autocfg",
 ]
 
+[[package]]
+name = "once_cell"
```
lopopolo, line 395: 😅 omg I misremembered.

Artichoke _did_ have `once_cell` back when it had a lazily-allocated `Regex` backend but that got nuked a while ago. Either way, thanks for making this change. I prefer `once_cell` to `lazy_static`.

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -164,9 +165,9 @@ impl From<&str> for Offset {
             "Y" => Self::fixed(-12 * SECONDS_IN_HOUR),
             "Z" | "UTC" => Self::utc(),
             _ => {
-                lazy_static! {
-                    static ref HH_MM_MATCHER: Regex = Regex::new(r"^([\-\+]{1})(\d{2}):?(\d{2})$").unwrap();
-                }
+                static HH_MM_MATCHER: Lazy<Regex> = Lazy::new(|| {
+                    Regex::new(r"^([\-\+]{1})(\d{2}):?(\d{2})$").unwrap()
+                });
```
lopopolo: :chef-kiss: no more macro magic.

spinoso-time/Cargo.toml
```diff
@@ -13,9 +13,18 @@ license = "MIT"
...
+
 [dependencies]
-chrono = { version = "0.4.19", default-features = false, features = ["clock"] }
-chrono-tz = { version = "0.6.0", default-features = false }
+chrono = { version = "0.4.19", default-features = false, features = ["clock"], optional = true }
+chrono-tz = { version = "0.6.0", default-features = false, optional = true }
+once_cell = { version = "1.12.0", optional = true }
+regex =  { version = "1.5.5", default-features = false, features = ["std", "unicode-perl"], optional = true }
+tz-rs = { version = "0.6.9", default-features = false, optional = true }
+tzdb = { version = "0.2.2", default-features = false, features = ["local"], optional = true }
```
lopopolo: will bumping this to 0.2.4 fix the minimum versions CI check?

To check this yourself you can do:

```console
$ cargo +nightly generate-lockfile -Zminimal-versions
$ cargo check
```

Cargo.lock
b-n: once_cell does look a bit nicer I must say. It's a little more explicit which is better IMO

spinoso-time/src/time/tzrs/to_a.rs
b-n: Nuked!

spinoso-time/Cargo.toml
b-n: ~Well, apparently we do actually need `const_fn` after all: https://docs.rs/tz-rs/0.6.9/src/tz/timezone/mod.rs.html#256-269~

~^-- this is used to generate the fixed timezones. After quickly reading up on `const_fn` (I was ignoring this dep until now), I see it's to help support older versions of rust, so I'm not sure if there is any getting around this apart from forking tz-rs and forcing rust 1.57+ support~

~I've dropped the commits relevant to removing the `const_fn` dep.~

Okay, after looking just a little bit more, I'm really not sure what's going on. I thought it was working for a bit, but then i realised i was running `cargo check` in the root, instead of in `spinoso-time`. Additionally, I checked the documentation for `const_fn` and according to what it says there, everything should still work even with the feature disabled.

Minimal versions downgrades `const_fn` to 0.4.2 (vs. 0.4.9), so my only explanation is that there was a change somewhere in there.

After digging into the errors a bit (rather than the change and :pray: approach), most errors are because `const_fn` is failing to wrap the functions. running with `-Z macro-backtrace` get's this (as an example):

```
error: expected one of: `nightly`, `cfg`, `feature`, string literal
   --> /home/ben/.cargo/registry/src/github.com-1ecc6299db9ec823/tz-rs-0.6.9/src/timezone/mod.rs:445:35
    |
445 |     #[cfg_attr(feature = "const", const_fn::const_fn)]
    |                                   ^^^^^^^^^^^^^^^^^^ in this procedural macro expansion
    |
   ::: /home/ben/.cargo/registry/src/github.com-1ecc6299db9ec823/const_fn-0.4.2/src/lib.rs:78:1
    |
78  | pub fn const_fn(args: TokenStream, input: TokenStream) -> TokenStream {
    | --------------------------------------------------------------------- in this expansion of `#[const_fn::const_fn]`
```

I'll have a play in tz-rs with different const_fn versions and see what's going on - but any ideas here would be A++

spinoso-time/src/time/tzrs/mod.rs
b-n: That is both crazy, but also makes our lives easier!

spinoso-time/src/time/tzrs/offset.rs
b-n: wait it's all enums?

but in all seriousness, I think this makes a lot of sense. Something bugged me about exposing the internals of tz-rs to the outside world, and this would make things a lot more encapsulated.

spinoso-time/Cargo.toml
b-n: Okay, found something. `tz-rs` needs `const_fn` >= 0.4.4.

e.g. this fails in tz-rs, but doesn't if `const_fn` is >= 0.4.4.
```
cargo update -Z minimal-versions
cargo check
```

PR to fix in tz-rs: https://github.com/x-hgg-x/tz-rs/pull/28

lopopolo: awesome! thanks for digging.

I made the PR to `tz-rs` that used `const_fn` in this way to make it optional. Whoops.

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -185,6 +185,13 @@ impl From<&str> for Offset {
     }
 }
 
+impl From<&[u8]> for Offset {
+    fn from(input: &[u8]) -> Self {
+        let input = str::from_utf8(input).expect("Invalid UTF8");
```
lopopolo, line 199: `From` impls should never panic. This one should probably be a `TryFrom` impl, but we can defer fixing that until the errors PR.

lopopolo (approved):

spinoso-time/src/time/tzrs/offset.rs
b-n: Ah yes. It's consistently incorrect with this implementation - rather than inconsistently incorrect :sweat_smile:

Cargo.lock
b-n: I've added `tzrs-local` as a feature flag. It returns GMT if disabled. Documented the behavior and added to feature flags docs.

More good news: disabling this feature = no [iana-time-zone](https://github.com/Kijewski/tzdb/blob/main/tzdb/Cargo.toml#L38) = no winapi/core-foundation when using tzrs time feature. :boom:

b-n:
@lopopolo I'm 99% sure I've captured all the feedback and made all the changes for this PR. The only thing I'm missing is wrapping the OffsetType. I've sent you a message on discord about it. I'll start on work rebasing the time-errors branch in the interim.

spinoso-time/Cargo.toml
lopopolo: It looks like your PR was merged upstream and released as 0.6.10. I'll follow up in a new PR and remove the load bearing dep on `const_fn`.

lopopolo (approved):
woohoo! ⏳

PR #1919, opened 2022-06-26
Fast follow from #1902

lopopolo:
Followup for small things left over from #1902.

cc @b-n


spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -172,10 +179,18 @@ impl From<&str> for Offset {
             "W" => Self::fixed(-10 * SECONDS_IN_HOUR),
             "X" => Self::fixed(-11 * SECONDS_IN_HOUR),
             "Y" => Self::fixed(-12 * SECONDS_IN_HOUR),
+            // ```console
```
b-n, line 182: Ah Ruby REPL here is what you meant - sorry, I thought you meant in artichoke-backend, but this is clear and makes way more sense!

b-n:
Solid changes all round!

PR #1924, opened 2022-06-28
Private offsettypes

b-n:
A quick change/fix for `spinoso-time::tzrs`.

With this PR, we hide all the tz-rs dependencies from anyone that implements this library.

Future: We will add a feature from tzdb called `find_from_name` as a feature, which will allow people to do things like `Offset::try_from("Europe/Amsterdam")`.

This is a follow on from #1902

### DIFF
--- a/spinoso-time/src/time/tzrs/offset.rs
+++ b/spinoso-time/src/time/tzrs/offset.rs
@@ -63,13 +63,13 @@ fn offset_hhmm_from_seconds(seconds: i32) -> String {
 /// Represents the number of seconds offset from UTC.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct Offset {
-    pub(crate) inner: OffsetType,
+    inner: OffsetType,
 }
 
 /// Represents the type of offset from UTC.
 #[allow(variant_size_differences)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
-pub(crate) enum OffsetType {
+enum OffsetType {
     /// UTC offset, zero offset, Zulu time
     Utc,
     /// Fixed offset from UTC.
@@ -120,10 +120,17 @@ impl<'a> Offset {
     /// predefined IANA time zones.
     #[inline]
     #[must_use]
-    pub(crate) fn tz(tz: TimeZoneRef<'static>) -> Self {
+    fn tz(tz: TimeZoneRef<'static>) -> Self {
         Self {
             inner: OffsetType::Tz(tz),
         }
+    }
+
+    /// Returns whether this offset is UTC.
+    #[inline]
+    #[must_use]
+    pub fn is_utc(&self) -> bool {
+        matches!(self.inner, OffsetType::Utc)
     }
 
     /// Returns a `TimeZoneRef` which can be used to generate and project
@@ -269,20 +276,26 @@ mod tests {
     #[test]
     fn fixed_zero_is_not_utc() {
         let offset = Offset::from(0);
-        assert!(matches!(offset.inner, OffsetType::Fixed(_)));
+        assert!(!offset.is_utc());
+    }
+
+    #[test]
+    fn utc_is_utc() {
+        let offset = Offset::utc();
+        assert!(offset.is_utc());
     }
 
     #[test]
     fn z_is_utc() {
         let offset = Offset::from("Z");
-        assert!(matches!(offset.inner, OffsetType::Utc));
+        assert!(offset.is_utc());
     }
 
     #[test]
     fn from_binary_string() {
         let tz: &[u8] = b"Z";
         let offset = Offset::from(tz);
-        assert!(matches!(offset.inner, OffsetType::Utc));
+        assert!(offset.is_utc());
     }
 
     #[test]
--- a/spinoso-time/src/time/tzrs/offset.rs
+++ b/spinoso-time/src/time/tzrs/offset.rs
@@ -59,16 +59,22 @@ fn offset_hhmm_from_seconds(seconds: i32) -> String {
     format!("{}{:0>2}{:0>2}", flag, offset_hours, offset_minutes)
 }
 
-/// Represents the number of seconds offset from UTC.
+/// Represents the number of seconds offset from UTC
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct Offset {
+    pub(crate) inner: OffsetType,
+}
+
+/// Represents the type of offset from UTC
 #[allow(variant_size_differences)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
-pub enum Offset {
-    /// UTC offset, zero offset, Zulu time.
+pub(crate) enum OffsetType {
+    /// UTC offset, zero offset, Zulu time
     Utc,
     /// Fixed offset from UTC.
     ///
     /// **Note**: A fixed offset of 0 is different from UTC time.
-    Fixed([LocalTimeType; 1]),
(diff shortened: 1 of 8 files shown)


lopopolo:
> Future: We will add a feature from tzdb called `find_from_name` as a feature, which will allow people to do things like `Offset::from("Europe/Amsterdam")`.

you mean `Offset::try_from` right? if the conversion is fallible (and not all strings are valid timezone offsets), let's make the conversion a `TryFrom` impl and push the decision to panic to callers.

b-n:
Ah yes! I have updated the description for posterity!

Also, `b-n/time-errors` has already fixed these non idiomatic impls :-D

b-n:
Note: In theory you could consider this as a breaking change since we are hiding one of the previously available methods from pub (e.g. `Offset::tz` now longer is accessible outside the crate). I'm not sure how strict you are on verisoning numbers in this case, because it could be seen to be 0.5 instead of 0.4.1

spinoso-time/src/time/tzrs/parts.rs
```diff
@@ -216,7 +218,7 @@ impl Time {
     #[inline]
     #[must_use]
     pub fn is_utc(&self) -> bool {
-        Offset::Utc == self.offset
+        OffsetType::Utc == self.offset.inner
```
lopopolo: nit, but for matching a single variant like this to implement a boolean accessor, this codebase tends to prefer `matches!` because it can be `const`:

```suggestion
        matches!(self.offset.inner, OffsetType::Utc)
```

like:

https://github.com/artichoke/artichoke/blob/c45f0abcb8799f328db8e7a603f5da490443dc9a/spinoso-string/src/codepoints.rs#L102-L112

spinoso-time/src/time/tzrs/parts.rs
```diff
@@ -194,8 +194,10 @@ impl Time {
     #[inline]
     #[must_use]
     pub fn time_zone(&self) -> &str {
-        match self.offset {
-            Offset::Utc => "UTC",
+        // We can usually get the name from wrapped DateTime, however UTC is a
+        // special case which is an empty string, thus the OffsetType is safer.
```
lopopolo, line 197: thanks for adding this comment

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -250,30 +261,30 @@ mod tests {
...
-        assert!(matches!(offset, Offset::Utc));
+        assert!(matches!(offset.inner, OffsetType::Utc));
     }
 
     #[test]
     fn from_binary_string() {
         let tz: &[u8] = b"Z";
         let offset = Offset::from(tz);
-        assert!(matches!(offset, Offset::Utc));
+        assert!(matches!(offset.inner, OffsetType::Utc));
```
lopopolo: can we add `assert!(offset.is_utc());` to get some coverage on that accessor too?

ditto for all of these.

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -100,38 +108,41 @@ impl<'a> Offset {
...
-                Ok(tz) => tz,
-                Err(_) => GMT,
-            },
-
-            Self::Tz(zone) => *zone,
+    pub(crate) fn time_zone_ref(&self) -> TimeZoneRef<'_> {
+        match self.inner {
+            OffsetType::Utc => TimeZoneRef::utc(),
+            OffsetType::Fixed(ref local_time_type) => {
+                match TimeZoneRef::new(&[], std::slice::from_ref(local_time_type), &[], &None) {
```
lopopolo: can you `use std::slice;` at the top of this file and change this to:

```suggestion
                match TimeZoneRef::new(&[], slice::from_ref(local_time_type), &[], &None) {
```

this codebase tends not to use fully qualified paths into `std` and `core`.

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -100,38 +108,41 @@ impl<'a> Offset {
...
     /// This can be combined with [`tzdb`] to generate offsets based on
     /// predefined IANA time zones.
     ///
-    /// ```
-    /// use spinoso_time::tzrs::Offset;
-    /// use tzdb::time_zone::pacific::AUCKLAND;
-    /// let offset = Offset::tz(AUCKLAND);
-    /// ```
+    /// [`tz:timezone::TimeZoneRef`]: https://docs.rs/tz-rs/0.6.9/tz/timezone/struct.TimeZoneRef.html
+    /// [`tzdb`]: https://docs.rs/tzdb/latest/tzdb/index.html
```
lopopolo: these comments are not necessary. rustdoc will resolve these links to the actual docs built.

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -59,16 +59,22 @@ fn offset_hhmm_from_seconds(seconds: i32) -> String {
     format!("{}{:0>2}{:0>2}", flag, offset_hours, offset_minutes)
 }
 
-/// Represents the number of seconds offset from UTC.
+/// Represents the number of seconds offset from UTC
```
lopopolo:
```suggestion
/// Represents the number of seconds offset from UTC.
```

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -59,16 +59,22 @@ fn offset_hhmm_from_seconds(seconds: i32) -> String {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct Offset {
+    pub(crate) inner: OffsetType,
+}
+
+/// Represents the type of offset from UTC
```
lopopolo:
```suggestion
/// Represents the type of offset from UTC.
```

lopopolo (approved):
left some feedback but this mostly looks good to me.

lopopolo:
> I'm not sure how strict you are on verisoning numbers in this case, because it could be seen to be 0.5 instead of 0.4.1

All of the version bumps have been mostly judgement calls and almost exclusively for breaking changes. Since we aren't publishing these, whatever you feel is appropriate works for me.

spinoso-time/src/time/tzrs/offset.rs
b-n, line 122:
```suggestion
```

b-n: Must have come in with a merge resolution and I thought they were meant to be there. whoops, removing

spinoso-time/src/time/tzrs/offset.rs
b-n: Are you talking about adding a `is_utc()` option to `Offset` ?

Come to think of it, that would mean I can get rid of `pub(crate) OffsetType` :thinking: 

Edit: Ignore this, I'm going to add this (it just makes more sense):

```
impl Offset {
  pub fn is_utc() -> bool {}
}
```

Ninja edit: I had the idea of adding `time_zone_name` to `Offset`, but that doesn't work since we want the timezone name of the Datetime, which varies depending on whether it was constructed in DST or not. I'm going to add a comment, because I keep thinking it's possible, and it's now)

b-n: And it should now look a little better in 3aa2e02f45c3664f85a43840ce732109d63899b0. But please let me know if you were thinking of something else

lopopolo: looks great

lopopolo (approved):
Approved. Trying out our new process and delegating the merge to you.

PR #1927, opened 2022-06-29
Adding Result/Error constructs to `tzrs` feature in spinoso-time

b-n:
#1902 Added a new backend - however some of the idiomatic principles weren't followed. Specifically there are some `From` impls which can panic - These sorts of things should be `TryFrom` as an example.

This PR fixes a number of these issues. Specifically it aims to achieve these:

- [x] `ops` would ideally be `checked_add` and `checked_sub` due to overflows
- [x] `Offset::fixed()` should be fallible (range +/- 24*60*60)
- [x] `Offset::from(&str)` should be fallible
- [x] `Offset::from(&str)` tests should also test out of bound Regexs (e.g. `+HH:MMM`, `HHMM`, `/HH-MM`, `+HHH:MM`, `+HH::MM`)
- [x] `Offset::from(&str)` should not accept `+0060`, `+2400` timezones. (they are out of range).
~- [ ] Attempt to unify the errors across chrono and tz-rs since things like `ComponentOutOfRangeError` might be sharable now~ (This makes more sense to try after we align some errors)
- [x] Use the more idiomatic `TimeError` instead of `TimeErr`


b-n:
@lopopolo I've done some digging, and it doesn't look like there are any "good" options for doing checked add on the datetimes. I was somewhat surprise there wasn't something like `std::ops::CheckedAdd` or anything similar. We can just add the `Add` trait, but something doesn't feel super right about this:

```
impl Add<T> for Time {
  type Output = Result<Self, Error>;
  fn try_from(value: T) -> Self::Output { }
}
```

This would end us in situations of
```
let now = Time::now();
let next = now + 1; // <-- next is a Result<Time, Error>, not a Time;
```

And I didn't try it until just now, but I see rust doesn't like function overloading (thank goodness), so this means just impl functions won't cut it.

My thoughts is implementing a trait:
```
pub trait CheckedAdd<T>: Sized {
    type Error;
    fn checked_add(value: T) -> Result<Self, Self::Error>;
}
```

Chrono currently uses `num-traits` for some of their additions of two times, however unfortuantely [the implementation can only add the same type to the same type](https://docs.rs/num-traits/0.2.15/num_traits/ops/checked/trait.CheckedAdd.html) where as we are wanting to add a different type to another type.

Do you have any other bright[er] ideas than implementing this trait?

lopopolo:
> My thoughts is implementing a trait:

For these cases, I think we can just implement `Time::checked_add` and `Time::checked_sub` that take `Time` as regular methods. (and maybe similar methods that take `f64`, `i64`, and `Duration`?)

I don't think the complexity of a trait is worth it here, especially because in the trampolines for `Time#add` and similar, we do the unboxing to specific values, so if we are in the branch that unboxes a `Float`, we know we need to call `Time::checked_add_f64`, or something like that.

Then if we still want to implement `std::ops::Add`, the body of the trait impl can be `time.checked_add(other).unwrap()`.

spinoso-time/src/time/tzrs/convert.rs
```diff
@@ -12,7 +12,7 @@ impl Display for Time {
     ///
     /// ```
     /// use spinoso_time::tzrs::Time;
-    /// let now = Time::utc(2022, 05, 26, 13, 16, 22, 0);
+    /// let now = Time::utc(2022, 05, 26, 13, 16, 22, 0).unwrap();
```
lopopolo: let's use the trick to wrap all of these doc tests in a function so the examples can use idiomatic `?`.

If we're gonna give example code, we might as well encourage folks to propagate errors instead of unwrapping.

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,82 @@
+use core::fmt;
+use std::error;
+
+use tz::error::{DateTimeError, ProjectDateTimeError, TzError};
+
+use super::offset::TzStringError;
+
+/// A wrapper around some of the errors provided by `tz-rs`.
+#[derive(Debug)]
+pub enum TimeErr {
```
lopopolo: Can we rename this to `TimeError`.

In `std`, error types tend to have `Error` suffixes in their names, like `ParseIntError`: https://doc.rust-lang.org/stable/std/num/struct.ParseIntError.html.

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl error::Error for TimeErr {}
+
+impl fmt::Display for TimeErr {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            TimeErr::ProjectionError(error) => error.fmt(f),
+            TimeErr::ComponentOutOfRangeError(error) => error.fmt(f),
+            TimeErr::TzStringError(error) => error.fmt(f),
```
lopopolo: this is pretty nifty

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,82 @@
...
+    ///
+    /// Note2: This is different from how MRI ruby is implemented. e.g. Second 60 is valid in MRI, and
+    /// will just add an additional second instead of erroring.
+    ComponentOutOfRangeError(DateTimeError),
+
+    /// The provided time zone string cannot be used
+    TzStringError(TzStringError),
+}
+
+impl error::Error for TimeErr {}
```
lopopolo: can we impl `Error::source` to give folks access to the inner error?

https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.source

Like this example from an error enum in `spinoso-random`:

https://github.com/artichoke/artichoke/blob/0bc3e01aee2886b91fd0036e984a3fb98869fa2d/spinoso-random/src/lib.rs#L183-L192

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,82 @@
...
+            TzError::TzStringError(error) => panic!("{}", error), //Occurs during parsing of TZif files (POSIX string parsing)
+            TzError::OutOfRangeError(error) => panic!("{}", error), //Occurs during int conversion (e.g. i64 => i32)
+            TzError::LocalTimeTypeError(error) => panic!("{}", error), //Occurs during creation of TimeZoneRef
+            TzError::TransitionRuleError(error) => panic!("{}", error), //Occurs during creation of TimeZoneRef
+            TzError::TimeZoneError(error) => panic!("{}", error), //Occurs during creation of TimeZoneRef
+            TzError::FindLocalTimeTypeError(error) => panic!("{}", error), //Wrapped by ProjectDateTimeError
+            TzError::IoError(error) => panic!("{}", error),       //Never explicitly returned
+            TzError::Utf8Error(error) => panic!("{}", error),     //Never explicity returned
+            TzError::TryFromSliceError(error) => panic!("{}", error), //Never explicitly returned
+            _ => panic!("Unhandled error"),
```
lopopolo: is this because `TzError` is a `#[non_exhaustive]` enum? I think a comment to that effect and squashing this case to a generic `TimeErr::Unknown` variant would be preferable to panicking.

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,82 @@
...
+            TzError::TzFileError(error) => panic!("{}", error),     //Occurs during parsing of TZif files
+            TzError::TzStringError(error) => panic!("{}", error), //Occurs during parsing of TZif files (POSIX string parsing)
+            TzError::OutOfRangeError(error) => panic!("{}", error), //Occurs during int conversion (e.g. i64 => i32)
+            TzError::LocalTimeTypeError(error) => panic!("{}", error), //Occurs during creation of TimeZoneRef
+            TzError::TransitionRuleError(error) => panic!("{}", error), //Occurs during creation of TimeZoneRef
+            TzError::TimeZoneError(error) => panic!("{}", error), //Occurs during creation of TimeZoneRef
+            TzError::FindLocalTimeTypeError(error) => panic!("{}", error), //Wrapped by ProjectDateTimeError
+            TzError::IoError(error) => panic!("{}", error),       //Never explicitly returned
+            TzError::Utf8Error(error) => panic!("{}", error),     //Never explicity returned
+            TzError::TryFromSliceError(error) => panic!("{}", error), //Never explicitly returned
```
lopopolo: do these errors never occur for the APIs we call from `tzdb`?

If that's the case, I think `unreachable!` communicates that intent more than `panic!`.

I think I'd prefer each of these branches to be something like:

```rust
TzError::TzFileError(error) => {
    if cfg!(debug_assertions) {
        panic!("Unexpected error from tzdb: {}", error);
    }
    TimeErr::Unknown
}
```

This way we'd preserve the panic in debug mode but have hope to turn this into a `RuntimeError` in the Ruby VM in production.

b-n:
> we can just implement Time::checked_add and Time::checked_sub that take Time

Funnily enough, it's not supported (in ruby 2.6.3, and also in 3.1.2):

```
irb(main):004:0> Time.at(1000) + Time.at(1000)
        6: from /home/ben/.rbenv/versions/2.6.3/bin/irb:23:in `<main>'
        5: from /home/ben/.rbenv/versions/2.6.3/bin/irb:23:in `load'
        4: from /home/ben/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        3: from (irb):4
        2: from (irb):4:in `rescue in irb_binding'
        1: from (irb):4:in `+'
TypeError (time + time?)
```

A couple of options:

1. `checked_add(value: Duration)` - this is the type we can use for addition because it supports both seconds and nanos without float problems
2. `checked_add_duration(value: Duration)` - slighly more explicit version of the above

I'm in favour of `checked_add(value: Duration)` - e.g. we force rust people to think of adding durations by default. Do you have a pref?

Either way, documenting here for posterity, but i'll add `checked_[add|sub]_[i8|i16|i32|i64|u8|u16|u32|u64]` like we currently are.

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -197,23 +215,25 @@ impl From<ToA> for Time {
...
+    /// # Errors
+    ///
+    /// Can produce a [`TimeErr`], generally when provided values are out of range
     #[inline]
-    fn from(to_a: ToA) -> Self {
+    fn try_from(to_a: ToA) -> Result<Self> {
+        let offset = match Offset::try_from(to_a.zone) {
+            Ok(offset) => offset,
+            Err(_) => Offset::utc(),
+        };
```
lopopolo: style nit:

```suggestion
        let offset = Offset::try_from(to_a.zone).unwrap_or_else(Offset::utc);
```

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -8,6 +10,17 @@ use tz::timezone::{LocalTimeType, TimeZoneRef};
 use tzdb::local_tz;
 use tzdb::time_zone::etc::GMT;
 
+#[derive(Debug, Clone)]
+pub struct TzStringError(String);
```
lopopolo: given the error messages raised by `Time.new`, I think this inner field should be a `Vec<u8>` of the given timezone string:

```
[3.1.2] > Time.new(2022, 6, 7, 12, 34, 12, '\xFF')
<internal:timev>:310:in `initialize': "+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: \xFF (ArgumentError)
        from (irb):2:in `new'
        from (irb):2:in `<main>'
```

Then we can add a `TzStringError::message` function that formats the user input into the `ArgumentError` message.

For the `fmt::Display` impl, we can use `String::from_utf8_lossy`.

spinoso-time/src/time/tzrs/timezone.rs
```diff
@@ -73,17 +89,31 @@ impl Time {
...
-        if let Ok(time) = self.inner.project(time_zone_ref) {
-            self.inner = time;
+
+        match self.inner.project(time_zone_ref) {
+            Ok(time) => {
+                self.inner = time;
+                self.offset = offset;
+                Ok(())
+            }
+            Err(error) => Err(TimeErr::from(error)),
```
lopopolo: this branch is equivalent to doing

```rust
let time = self.inner.project(time_zone_ref)?;
```

which I prefer since it removes a few levels of rightward drift with the brackets.

spinoso-time/src/time/tzrs/offset/error.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+impl error::Error for TzStringError {}
+
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct OutOfRangeError;
+
+impl fmt::Display for OutOfRangeError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(f, "utc_offset out of range")
```
lopopolo: Ahh I see that this is that error message now.

```
[3.1.2] > Time.new(2022, 6, 7, 12, 34, 12, '+45:00')
<internal:timev>:310:in `initialize': utc_offset out of range (ArgumentError)
        from (irb):3:in `new'
        from (irb):3:in `<main>'
```

spinoso-time/src/time/tzrs/offset/error.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn from(error: OutOfRangeError) -> Self {
+        Self::OutOfRangeError(error)
+    }
+}
+
+impl From<Utf8Error> for OffsetError {
+    fn from(_: Utf8Error) -> Self {
+        Self::TzStringError(TzStringError("timezone contains invalid UTF8".to_string()))
+    }
+}
```
lopopolo: neato! let the `?` do all the work.

spinoso-time/src/time/tzrs/offset/error.rs
```diff
@@ -21,12 +21,12 @@ impl fmt::Display for OffsetError {
...
 #[allow(clippy::module_name_repetitions)]
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct TzStringError(pub(crate) String);
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct TzStringError;
 
 impl fmt::Display for TzStringError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        f.write_str(&self.0)
+        write!(f, "expected [+/-]HH[:]MM, A..K, J..Z for utc_offset")
```
lopopolo: is this the error message from 2.6.3? It looks different from 3.1.2 (which I have locally):

```
[3.1.2] > Time.new(2022, 6, 7, 12, 34, 12, '\xFF')
<internal:timev>:310:in `initialize': "+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: \xFF (ArgumentError)
```

if this message comes from 2.6.3, can we attach a REPL snippet in a comment?

spinoso-time/src/time/tzrs/offset/mod.rs
```diff
@@ -382,6 +382,20 @@ mod tests {
         );
     }
 
+    #[test]
+    fn from_str_invalid_fixed_strings() {
+        let invalid_fixed_strings = [
+            "+01:010", "+010:10", "+010:010", "0110", "01:10", "01-10", "+01-10", "+01::10",
+        ];
+
+        for invalid_string in invalid_fixed_strings {
```
lopopolo: yay table-driven tests.

spinoso-time/src/time/tzrs/offset/mod.rs
```diff
@@ -382,6 +382,20 @@ mod tests {
...
+            assert_eq!(
+                Err(OffsetError::TzStringError(TzStringError)),
+                Offset::try_from(invalid_string)
+            );
```
lopopolo: since this a table driven test, `assert_eq!` accepts a format string as its third argument which we can use to inform folks which case failed.

```rust
            assert_eq!(
                Err(OffsetError::TzStringError(TzStringError)),
                Offset::try_from(invalid_string),
                "Expected Offset::TzStringError for {}",
                invalid_string,
            );
```

spinoso-time/src/time/tzrs/build.rs
```diff
@@ -7,7 +7,7 @@ impl Time {
     ///
     /// # Errors
     ///
-    /// Can produce a [`TimeErr`], generally when provided values are out of range
+    /// Can produce a [`TimeError`], generally when provided values are out of range
```
lopopolo: whoops disregard my previous comment (I'm going commit-by-commit).

lopopolo:
> irb(main):004:0> Time.at(1000) + Time.at(1000)

Add but Time subtraction is supported!

```
[3.1.2] > Time.now - Time.now
=> -2.0e-06
```

> A couple of options:
> 
> 1. `checked_add(value: Duration)` - this is the type we can use for addition because it supports both seconds and nanos without float problems
> 2. `checked_add_duration(value: Duration)` - slighly more explicit version of the above

I'm in favor of this, but with the caveat that `spinoso-time` should probably provide some free functions which convert `i64` and `f64` to `core::time::Duration`. It would be not fun to implement all of that conversion code in the trampolines.

lopopolo:
> Either way, documenting here for posterity, but i'll add `checked_[add|sub]_[i8|i16|i32|i64|u8|u16|u32|u64]` like we currently are.

Let's impl these APIs in terms of `Duration` like you suggested, but make it easy for glue code in the trampolines to get a `Duration` from the primitive types. I guess that means these converter APIs would return `Option<Duration>` in case a negative value is passed?

So something like:

```rust
let float = ...;
let duration = spinoso_time::duration_from_f64(float)
    .or_else(|| {
        float.checked_neg().and_then(spinoso_time::duration_from_f64)
    })
    .ok_or_else(|| FloatDomainError::with_message("out of range")?;
```

Or maybe `spinoso_time::duration_from_f64` returns an enum so callers can disambiguate when they need to call `checked_add` or `checked_sub`?

spinoso-time/src/time/tzrs/error.rs
b-n: I ~stole~ borrowed it from the tz-rs crate :sweat_smile:.

spinoso-time/src/time/tzrs/error.rs
b-n: Ooo yes, good call!

spinoso-time/src/time/tzrs/offset/error.rs
b-n: Ahhh yes, this is where it gets fun. When originally developing this, I was running irb in artichoke - so it was picking up 3.1.2 and I didn't notice until later. I did some double checking, and 2.6.3 only supports the +HH:MM and -HH:MM syntax. It doesn't support the A-IK-Z or +HHMM/-HHMM syntax :facepalm:.

Is it best in this case just to support only the 2.6.3 version?

spinoso-time/src/time/tzrs/offset/mod.rs
b-n: hah yes! I was going to ask whether there is a style preference here. The same thing could be done with other tests and cheap hashmap tables (or a tuple lists i guess).

b-n:
> Let's impl these APIs in terms of Duration like you suggested, but make it easy for glue code in the trampolines to get a Duration from the primitive types. I guess that means these converter APIs would return Option<Duration> in case a negative value is passed?

I've got one better I think (psuedo codeish):

```rust
fn checked_add(&self, value: Duration) -> Self {
  //magic
}

fn checked_add_f64(&self, value: f64) -> Self {
  let nanos = nanos_from_float(value);
  let duration = Duration::new(value.abs().floor(), nanos); 
  if value > 0 {
    self.checked_add(duration)
  } else {
    self.checked_sub(duration)
  }
}
```

spinoso-time/src/time/tzrs/offset/error.rs
lopopolo: the 2.6.3 target isn't something I take super seriously. The 2.6 line will be out of support soon. I tend to run the latest stable Ruby locally and use that to base MRI behavior off of.

Since we've already got the 3.1.2 behavior implemented, let's keep it.

We should probably rethink what Artichoke's MRI target is and how to message that.

spinoso-time/src/time/tzrs/error.rs
b-n: I'm mostly sure I've found a way to report the underlying error to the Ruby VM in production by wrapping these in their own special Unknown. This gives the ability to still `fmt::Display` them and `error::Error::source` them. How do you feel about these issues surfacing in the Ruby VM?

Note: all of these errors shouldn't happen ever because of how this library is written #famouslastwords

spinoso-time/src/time/tzrs/mod.rs
b-n: The great thing about working with other people, these little nuggets!

spinoso-time/src/time/tzrs/timezone.rs
b-n: yes! much tidier!

spinoso-time/src/time/tzrs/offset/error.rs
b-n: I'm in two minds about providing the string, or keeping it defaulted. The limitation of keeping it defaulted like this is that it's hard to inject what value was incorrect (e.g. in TzStringError it would be nice to error with the value that was provided).

The reason I put the defaulted values in like this was specifically because matching the error in the test was giving me some hassle, and I just wanted to get the test cases working first.

Do you think it's worth changing `TzStringError` tto provide the provided value?

spinoso-time/src/time/tzrs/offset/error.rs
b-n: Super!

I'll update the message to match Ruby 3 then. Maybe worth while thinking about [this](https://github.com/artichoke/artichoke/pull/1927#discussion_r910298551) as to whether we really want the provided value in the error etc.

spinoso-time/src/time/tzrs/offset/error.rs
lopopolo: I think for now this is fine and we can defer this decision to later.

lopopolo:
this looks mostly good so far and I think you've got a solution to the `ops` problem in progress. I'm gonna step back on this PR. Once it's no longer draft, ping me and I'll take another look.

spinoso-time/src/time/tzrs/offset/error.rs
lopopolo: oo @b-n I found this tweet which basically says the same thing: https://twitter.com/artichokeruby/status/1514027254715584513

b-n:
@lopopolo How do you feel about [58d30dc](https://github.com/artichoke/artichoke/pull/1927/commits/58d30dcedfed23b8e02fd930ed1b2bafdccc9626) ?

(IMO, it needs a bit of work on the wording, but my brain is not turning up anything else noteworthy today)

spinoso-time/src/time/tzrs/offset.rs
b-n: Resolving because of https://github.com/artichoke/artichoke/pull/1927#discussion_r910142415 <-- defer decision later, but let me know if we want to tackle this one now.

spinoso-time/src/time/tzrs/offset/error.rs
b-n: I've updated to the MRI 3.x value

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,205 @@
...
+    TzStringError(TzStringError),
+
+    /// The provided tz offset seconds offset is outside of the allowed range.
+    TzOutOfRangeError(TzOutOfRangeError),
+
+    /// A rescuable Integer overflow error. Caused when trying to exceed the
+    /// bounds of an int.
+    IntOverflowError(IntOverflowError),
+
+    /// An rescuable unknown error (instead of panicing)
```
lopopolo:
```suggestion
    /// An rescuable unknown error (instead of panicking)
```

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,205 @@
+use core::fmt;
+use core::num::TryFromIntError;
+use std::error;
+use std::str::Utf8Error;
+
+use tz::error::{DateTimeError, ProjectDateTimeError, TzError};
+
+/// A wrapper around some of the errors provided by `tz-rs`.
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug)]
```
lopopolo, line 10: can this enum derive `Clone`?

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,205 @@
...
+
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct TzStringError;
+
+impl fmt::Display for TzStringError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(
+            f,
+            "+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset"
```
lopopolo: nit, this might be clearer with a "raw string":

```suggestion
            r#""+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset"#
```

and fixes the missing opening quote on `"+HH:MM"`

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,205 @@
...
+impl error::Error for TzStringError {}
+
+impl From<Utf8Error> for TzStringError {
+    fn from(_: Utf8Error) -> Self {
+        TzStringError
+    }
+}
+
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
lopopolo:
```suggestion
#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
```

https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,205 @@
...
+pub struct TzOutOfRangeError;
+
+impl fmt::Display for TzOutOfRangeError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(f, "utc_offset out of range")
```
lopopolo: nit, when implementing `Display` with a fixed string like this, you can use: `fmt::Formatter::write_str` and avoid the macro (and a minuscule compile time hit).

https://github.com/artichoke/artichoke/blob/b06bb37b64dd9c2415fbac011b2029ba18b36c83/spinoso-securerandom/src/lib.rs#L162-L167

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+}
+
+impl error::Error for TzOutOfRangeError {}
+
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
lopopolo:
```suggestion
#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
```

ditto

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -0,0 +1,205 @@
...
+pub struct IntOverflowError;
```
lopopolo: minor nit, for zero sized error structs like this, I typically define them with a private member so they can't be constructed outside of this module.

https://github.com/artichoke/artichoke/blob/b06bb37b64dd9c2415fbac011b2029ba18b36c83/spinoso-securerandom/src/lib.rs#L278-L281

with a corresponding `new` method if it is required:

https://github.com/artichoke/artichoke/blob/b06bb37b64dd9c2415fbac011b2029ba18b36c83/spinoso-securerandom/src/lib.rs#L292-L306

spinoso-time/src/time/tzrs/math.rs
```diff
@@ -81,225 +86,309 @@ impl Time {
...
-    fn add(self, seconds: i16) -> Self::Output {
-        self + i64::from(seconds)
+    /// Addition — Adds some i16 to _time_ and returns that value as a new Time
+    /// object.
+    ///
+    /// # Errors
+    ///
+    /// If this function attempts to overflow the the number of seconds as an
+    /// i64 then a [`TimeError`] will be returned.
+    pub fn checked_add_i16(&self, seconds: i16) -> Result<Self, TimeError> {
```
lopopolo: I'm not convinced of the value of the `[iu]8`, `[iu]16`, and `[iu]32` variants of these methods. Can we drop these and push the conversion to `i64` and `u64` to callers?

I think all of the `Add` and `Sub` impls were there for inference reasons, but in the absence of operator overloading, I don't think they're necessary anymore.

spinoso-time/src/time/tzrs/math.rs
```diff
@@ -81,225 +86,309 @@ impl Time {
...
 
-    fn add(self, seconds: i64) -> Self::Output {
+    /// Addition — Adds some i64 to _time_ and returns that value as a new Time
+    /// object.
+    ///
+    /// # Errors
+    ///
+    /// If this function attempts to overflow the the number of seconds as an
+    /// i64 then a [`TimeError`] will be returned.
+    pub fn checked_add_i64(&self, seconds: i64) -> Result<Self, TimeError> {
```
lopopolo, line 126: this is so nice!

spinoso-time/src/time/tzrs/math.rs
```diff
@@ -81,225 +86,309 @@ impl Time {
...
-    fn add(self, seconds: f32) -> Self::Output {
-        self + f64::from(seconds)
+    /// Addition — Adds some f32 fraction seconds to _time_ and returns that
+    /// value as a new Time object.
+    ///
+    /// # Errors
+    ///
+    /// If this function attempts to overflow the the number of seconds as an
+    /// i64 then a [`TimeError`] will be returned.
+    pub fn checked_add_f32(&self, seconds: f32) -> Result<Self, TimeError> {
```
lopopolo: ditto I think we can drop this in favor of pushing the conversion to `f64` to callers

spinoso-time/src/time/tzrs/math.rs
```diff
@@ -81,225 +86,309 @@ impl Time {
...
+    /// value as a new Time object.
+    ///
+    /// # Errors
+    ///
+    /// If this function attempts to overflow the the number of seconds as an
+    /// i64 then a [`TimeError`] will be returned.
+    pub fn checked_add_f64(&self, seconds: f64) -> Result<Self, TimeError> {
         if seconds.is_sign_positive() {
-            self + Duration::from_secs_f64(seconds)
+            self.checked_add(Duration::from_secs_f64(seconds))
```
lopopolo, line 165: `Duration::from_secs_f64` panics in the following scenarios:

> This constructor will panic if secs is negative, overflows Duration or not finite.

https://doc.rust-lang.org/1.61.0/std/time/struct.Duration.html#panics-1

Should we be returning `TimeError` (out of range error?) if `seconds` is `NaN` or infinite?

spinoso-time/src/time/tzrs/math.rs
```diff
@@ -81,225 +86,309 @@ impl Time {
...
     }
 }
 
-impl Sub<Duration> for Time {
-    type Output = Self;
-
-    fn sub(self, duration: Duration) -> Self::Output {
+// Subtraction
+impl Time {
+    /// Subtration — Subtracts the given duration from _time_ and returns that
```
lopopolo:
```suggestion
    /// Subtraction — Subtracts the given duration from _time_ and returns that
```

spinoso-time/src/time/tzrs/math.rs
```diff
@@ -81,225 +86,309 @@ impl Time {
...
-    fn sub(self, seconds: i8) -> Self::Output {
-        self - i64::from(seconds)
+    /// Subtration — Subtracts the given i8 from _time_ and returns that value
+    /// as a new Time object.
+    ///
+    /// # Errors
+    ///
+    /// If this function attempts to overflow the the number of seconds as an
+    /// i64 then a [`TimeError`] will be returned.
+    pub fn checked_sub_i8(self, seconds: i8) -> Result<Self, TimeError> {
```
lopopolo: ditto add, let's nuke the 8, 16, and 32 bit variants.

spinoso-time/src/time/tzrs/math.rs
```diff
@@ -81,225 +86,309 @@ impl Time {
...
-    fn sub(self, seconds: f64) -> Self::Output {
-        self + -seconds
+    /// Subtration — Subtracts the given f64 as fraction seconds from _time_
+    /// and returns that value as a new Time object.
+    ///
+    /// # Errors
+    ///
+    /// If this function attempts to overflow the the number of seconds as an
+    /// i64 then a [`TimeError`] will be returned.
+    pub fn checked_sub_f64(self, seconds: f64) -> Result<Self, TimeError> {
```
lopopolo, line 239: like `checked_add_f64` above, I think we need to check for `NaN` and inifinities and return `TimeError` to avoid panics.

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -111,16 +124,24 @@ impl Time {
...
+    /// let offset = Offset::try_from("+1200")?;
     /// let t = Time::new(2022, 9, 25, 1, 30, 0, 0, offset);
+    /// # Ok(())
+    /// # }
+    /// # example().unwrap();
     /// ```
     ///
+    /// # Errors
+    ///
+    /// Can produce a [`TimeError`], generally when provided values are out of range
```
lopopolo:
```suggestion
    /// Can produce a [`TimeError`], generally when provided values are out of range.
```

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -134,14 +155,13 @@ impl Time {
...
-        let found_date_times = DateTime::find(year, month, day, hour, minute, second, nanoseconds, tz)
-            .expect("Could not find a matching DateTime for this timezone");
-        let dt = found_date_times
-            .unique()
-            .expect("Could not find a matching DateTime for this timezone");
-        Self { inner: dt, offset }
+        let found_date_times = DateTime::find(year, month, day, hour, minute, second, nanoseconds, tz)?;
+
+        // .latest() will always return Some(DateTime)
+        let dt = found_date_times.latest().expect("No datetime found with this offset");
```
lopopolo, line 163: is this expect guaranteed to not panic?

The docs for `DateTime::find` and its return type `FoundDateTimeList` indicate that the returned result may be empty:

https://docs.rs/tz-rs/latest/tz/datetime/struct.FoundDateTimeList.html

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -151,17 +171,25 @@ impl Time {
...
+    /// # fn example() -> Result<(), TimeError> {
+    /// let now = Time::now()?;
+    /// # Ok(())
+    /// # }
+    /// # example().unwrap();
     /// ```
     ///
+    /// # Errors
+    ///
+    /// Can produce a [`TimeError`], however these should never been seen in regular usage
```
lopopolo:
```suggestion
    /// Can produce a [`TimeError`], however these should never been seen in regular usage.
```

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -172,22 +200,32 @@ impl Time {
...
+    /// let t = Time::with_timespec_and_offset(0, 0, offset)?;
     /// assert_eq!(t.to_int(), 0);
+    /// # Ok(())
+    /// # }
+    /// # example().unwrap();
     /// ```
     ///
+    /// # Errors
+    ///
+    /// Can produce a [`TimeError`], however these should not be seen during regular usage
```
lopopolo:
```suggestion
    /// Can produce a [`TimeError`], however these should not be seen during regular usage.
```

spinoso-time/src/time/tzrs/mod.rs
```diff
@@ -196,24 +234,27 @@ impl From<ToA> for Time {
...
     /// assert_eq!(now.second(), from_to_a.second());
     /// assert_ne!(now.nanoseconds(), from_to_a.nanoseconds());
+    /// # Ok(())
+    /// # }
+    /// # example().unwrap();
     /// ```
+    ///
+    /// # Errors
+    ///
+    /// Can produce a [`TimeError`], generally when provided values are out of range
```
lopopolo:
```suggestion
    /// Can produce a [`TimeError`], generally when provided values are out of range.
```

spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -103,15 +111,26 @@ impl Offset {
     }
 
     /// Generate an offset with a number of seconds from UTC.
+    ///
+    /// # Errors
+    ///
+    /// Return a [`TimeError::TzOutOfRangeError`] when outside of range of
+    /// acceptable offset of [`MIN_OFFSET_SECONDS`] to [`MAX_OFFSET_SECONDS`]
```
lopopolo:
```suggestion
    /// acceptable offset of [`MIN_OFFSET_SECONDS`] to [`MAX_OFFSET_SECONDS`].
```

lopopolo (changes requested):

spinoso-time/src/time/tzrs/error.rs
b-n: Not unless we make some PRs into tz-rs. the wrapped tz-rs errors don't implement it at present unfortunately.

spinoso-time/src/time/tzrs/error.rs
b-n: TIL raw string - nice!

spinoso-time/src/time/tzrs/error.rs
b-n: And it look nicer too!

spinoso-time/src/time/tzrs/math.rs
b-n: Ah yes, that makes total sense. I'm with you on `a + b` needs to support more than an explicit `a.checked_add_<type>(b)` :+1:

spinoso-time/src/time/tzrs/math.rs
b-n: Good catch! TIL floats can be NAN or Infinite.

spinoso-time/src/time/tzrs/mod.rs
b-n: Yes, I think that's true only when using `.unique` after having a look at the underlying implementation. `.latest` will always return [a result that we're able to work with](https://docs.rs/tz-rs/latest/src/tz/datetime/find.rs.html#50-56) luckily enough!

spinoso-time/src/time/tzrs/error.rs
b-n: ah yes. I assume this is because you don't want someone else constructing a `spinoso-securerandom::RandomBytesError` from some completely unrelated crate. Makes sense.

I ran across thiswhen creating the IntOverflowError because the std crate uses this: https://doc.rust-lang.org/stable/src/core/num/error.rs.html#6-9 instead - I probably should have clued into it then too hah.

I'll make the change to keep it consistent with `spinoso` style :+1:

b-n: Okie dokie, all done in 821218b9.

From my understanding, these errors were still hidden inside the crate internals, and coudn't be constructed outside of the crate. Specifically because the lack of `pub use` on the internal types (only `Timerror` has `pub` from within `mod`). I would claim this was by design, but it was a total fluke. I've added the docs to keep it consistent though.

Does this feel in line with the rest of `spinoso` etc?

Note, on the `message` fn - I see this would be super helpful in unwrapping `TimeError` into `ArgumentError`. However, that would also make sense to be added to `TimeError` too then - but I'm at a loss for the "best" way to get the error message from non `spinoso-time` errors, specifically the wrapped `tz-rs` errors. Any great thoughts here?

spinoso-time/src/time/tzrs/mod.rs
lopopolo:
```rust
    /// Returns the latest found date time if existing
    pub fn latest(&self) -> Option<DateTime> {
        // Found date times are computed in ascending order of Unix times
        match *self.0.last()? {
            FoundDateTimeKind::Normal(date_time) => Some(date_time),
            FoundDateTimeKind::Skipped { after_transition, .. } => Some(after_transition),
        }
    }
```

that `self.0.last()?` means that if the vec in `self.0` is empty, `None` is returned. We can leave this for now,  but I'd love to maybe find a test case for this empty vec scenario.

spinoso-time/src/time/tzrs/error.rs
lopopolo: this LGTM. With `TimeError` we can delegate to `fmt::Display` impl to get the underlying message. The `fn message` is less useful if we need to allocate a `String`.

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -154,9 +154,15 @@ impl From<IntOverflowError> for TimeError {
     }
 }
 
+/// Error that indicates that the provided string cannot be used in the creation
+/// of a timezone.
+///
+/// This error is returned by [`Offset::try_from`]
```
lopopolo:
```suggestion
/// This error is returned by [`Offset::try_from`].
```

spinoso-time/src/time/tzrs/error.rs
```diff
@@ -165,15 +171,29 @@ impl fmt::Display for TzStringError {
...
-        TzStringError
+        TzStringError::new()
     }
 }
 
+/// Error that indicates that a provided value is outside the allowed range of
+/// values. Generally seen when constructing an offset that is greater than the
+/// allowed range.
+///
+/// This error is returned by [`Offset::try_from`]
```
lopopolo:
```suggestion
/// This error is returned by [`Offset::try_from`].
```

lopopolo (approved):
Two missing `.` in some doc comments but this is good to go! Thanks for pushing this through. This is a final approval, merge whenever you feel comfortable (you don't need to kick it back to me if you fix the typos).

🎉

spinoso-time/src/time/tzrs/mod.rs
b-n: Hah, yes, good point!

I'll do some more digging through the generation and the tests there, there must be a use case where it does indeed fail. If it exists, I will find it 😄

PR #1955, opened 2022-07-12
Fix `Regexp::escape` for meta characters

### LINKED ISSUE #1738: `Regexp::escape` differs between MRI and Artichoke for meta characters
## MRI

```console
$ irb
[3.1.1] > s = '\.+*?()|[]{}^$#&-~'
=> "\\.+*?()|[]{}^$#&-~"
[3.1.1] > puts Regexp.escape(s)
\\\.\+\*\?\(\)\|\[\]\{\}\^\$\#&\-~
=> nil
[3.1.1] >
```

## Artichoke

```console
$ cargo run --bin airb -q
artichoke 0.1.0-pre.0 (2022-03-11 revision 5352) [x86_64-apple-darwin]
[rustc 1.58.1 (db9d1b20b 2022-01-20) on x86_64-apple-darwin]
>>> s = '\.+*?()|[]{}^$#&-~'
=> "\\.+*?()|[]{}^$#&-~"
>>> puts Regexp.escape(s)
\\\.\+\*\?\(\)\|\[\]\{\}\^\$\#\&\-\~
=> nil
>>>
```

## Diff

```
M: \\\.\+\*\?\(\)\|\[\]\{\}\^\$\#&\-~
A: \\\.\+\*\?\(\)\|\[\]\{\}\^\$\#\&\-\~
```

It looks like Artichoke is adding an escape slash for `&` and `~` where MRI is not.

## Implicated Code

https://github.com/artichoke/artichoke/blob/15726cd327e519cee2dab9cc7fab717be265d8f6/artichoke-backend/src/extn/core/regexp/syntax.rs#L48-L63

choznerol:
Close #1738 (Thanks for the detailed description!)

P.s. I tested the same pattern with MRI 3.1.2, 2.6.5, and 2.7.5 and the results are all the same.

### DIFF
--- a/artichoke-backend/src/extn/core/regexp/syntax.rs
+++ b/artichoke-backend/src/extn/core/regexp/syntax.rs
@@ -49,9 +49,7 @@ pub fn escape_into(text: &str, buf: &mut String) {
 #[must_use]
 pub fn is_meta_character(c: char) -> bool {
     match c {
-        '\\' | '.' | '+' | '*' | '?' | '(' | ')' | '|' | '[' | ']' | '{' | '}' | '^' | '$' | '#' | '&' | '-' | '~' => {
-            true
-        }
+        '\\' | '.' | '+' | '*' | '?' | '(' | ')' | '|' | '[' | ']' | '{' | '}' | '^' | '$' | '#' | '-' => true,
         // This match arm differs from `regex-syntax` by including '/'.
         // Ruby uses '/' to mark `Regexp` literals in source code.
         '/' => true,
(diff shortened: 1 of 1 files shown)


artichoke-backend/src/extn/core/regexp/syntax.rs
```diff
@@ -49,9 +49,7 @@ pub fn escape_into(text: &str, buf: &mut String) {
 #[must_use]
 pub fn is_meta_character(c: char) -> bool {
```
lopopolo, line 52: This change introduced a rustfmt change which triggered a clippy lint due to a prior false negative.

I filed these bugs upstream:

- https://github.com/rust-lang/rust-clippy/issues/9163
- https://github.com/rust-lang/rust-clippy/issues/9164

This set of `allow` attributes should get clippy happy:

```suggestion
#[allow(clippy::match_like_matches_macro)]
#[allow(clippy::match_same_arms)]
pub fn is_meta_character(c: char) -> bool {
```

lopopolo (approved):
Thanks for the PR and test @choznerol!

I left a suggestion for some `allow` attributes which will make clippy happy. I'll apply the patch and once CI is green, this is good to merge.

artichoke-backend/src/extn/core/regexp/syntax.rs
choznerol: Thanks for the detailed explanation and report upstream! The CI didn't run, so I was not aware of this issue when opening PR.

PR #1956, opened 2022-07-16
Switch artichoke-backend to use spinoso-time tzrs feature

b-n:
#1927 Added enough to hopefully make tzrs feature safe for replacement of the chrono feature in arthicoke-backend. This PR achieves to do exactly that

## Compatibility

The previous version of `Time#at` supported only the syntax that `mruby` did. This PR introduces all options available as at MRI 3.1.2.

`Time#succ` now uses the same logic as `Time#+` which wasn't implemented before.

`Time#+` is now implemented, and follows MRI

`Time#-` changes:

- The previous version use to just return rhs when subtracting one time from another (this now returns a float as per MRI)
- Arthimetic with intergers/floats now works

### DIFF
--- a/artichoke-backend/src/extn/core/time/mod.rs
+++ b/artichoke-backend/src/extn/core/time/mod.rs
@@ -21,13 +21,18 @@
 //! [`chrono-tz`]: https://crates.io/crates/chrono-tz
 
 use crate::convert::HeapAllocatedData;
+use crate::extn::prelude::*;
 
 pub mod mruby;
 pub mod trampoline;
 
 #[doc(inline)]
-pub use spinoso_time::chrono::Time;
+pub use spinoso_time::tzrs::*;
 
 impl HeapAllocatedData for Time {
     const RUBY_TYPE: &'static str = "Time";
+}
+
+pub(crate) fn convert_time_error_to_argument_error(error: TimeError) -> ArgumentError {
+    ArgumentError::from(format!("{}", error))
 }
--- a/artichoke-backend/src/extn/core/time/mod.rs
+++ b/artichoke-backend/src/extn/core/time/mod.rs
@@ -25,6 +25,7 @@ use crate::extn::prelude::*;
 
 pub mod mruby;
 pub mod trampoline;
+pub mod subsec;
 
 #[doc(inline)]
 pub use spinoso_time::tzrs::*;
--- a/artichoke-backend/src/extn/core/time/mod.rs
+++ b/artichoke-backend/src/extn/core/time/mod.rs
@@ -24,9 +24,9 @@ use crate::convert::HeapAllocatedData;
 use crate::extn::prelude::*;
 
 pub mod mruby;
-pub mod trampoline;
-pub mod subsec;
 pub mod offset;
+pub mod subsec;
+pub mod trampoline;
 
 #[doc(inline)]
 pub use spinoso_time::tzrs::*;
--- a/artichoke-backend/src/extn/core/time/mod.rs
+++ b/artichoke-backend/src/extn/core/time/mod.rs
@@ -26,6 +26,7 @@ use crate::extn::prelude::*;
 pub mod mruby;
 pub mod trampoline;
 pub mod subsec;
+pub mod offset;
 
 #[doc(inline)]
 pub use spinoso_time::tzrs::*;
--- a/artichoke-backend/src/extn/core/time/mod.rs
+++ b/artichoke-backend/src/extn/core/time/mod.rs
@@ -33,6 +33,8 @@ impl HeapAllocatedData for Time {
     const RUBY_TYPE: &'static str = "Time";
 }
 
-pub(crate) fn convert_time_error_to_argument_error(error: TimeError) -> ArgumentError {
-    ArgumentError::from(format!("{}", error))
+impl From<TimeError> for Error {
+    fn from(error: TimeError) -> Error {
+        ArgumentError::from(format!("{}", error)).into()
+    }
 }
--- a/artichoke-backend/src/extn/core/time/mruby.rs
+++ b/artichoke-backend/src/extn/core/time/mruby.rs
@@ -19,7 +19,7 @@ pub fn init(interp: &mut Artichoke) -> InitializeResult<()> {
         .value_is_rust_object()
         // Constructor
         .add_self_method("now", time_self_now, sys::mrb_args_none())?
-        .add_self_method("at", time_self_at, sys::mrb_args_req(1) | sys::mrb_args_opt(3))?
+        .add_self_method("at", time_self_at, sys::mrb_args_req_and_opt(1, 3))?
         .add_self_method("utc", time_self_mkutc, sys::mrb_args_any())?
         .add_self_method("gm", time_self_mkutc, sys::mrb_args_any())?
         .add_self_method("local", time_self_mktime, sys::mrb_args_any())?
--- a/artichoke-backend/src/extn/core/time/mruby.rs
+++ b/artichoke-backend/src/extn/core/time/mruby.rs
@@ -19,7 +19,7 @@ pub fn init(interp: &mut Artichoke) -> InitializeResult<()> {
         .value_is_rust_object()
         // Constructor
         .add_self_method("now", time_self_now, sys::mrb_args_none())?
-        .add_self_method("at", time_self_at, sys::mrb_args_req(1))?
+        .add_self_method("at", time_self_at, sys::mrb_args_req(1) | sys::mrb_args_opt(3))?
         .add_self_method("utc", time_self_mkutc, sys::mrb_args_any())?
         .add_self_method("gm", time_self_mkutc, sys::mrb_args_any())?
         .add_self_method("local", time_self_mktime, sys::mrb_args_any())?
@@ -112,11 +112,15 @@ unsafe extern "C" fn time_self_now(mrb: *mut sys::mrb_state, _slf: sys::mrb_valu
 }
 
 unsafe extern "C" fn time_self_at(mrb: *mut sys::mrb_state, _slf: sys::mrb_value) -> sys::mrb_value {
-    let (seconds, microseconds) = mrb_get_args!(mrb, required = 1, optional = 1);
+    let (seconds, opt1, opt2, opt3) = mrb_get_args!(mrb, required = 1, optional = 3);
     unwrap_interpreter!(mrb, to => guard);
     let seconds = Value::from(seconds);
-    let microseconds = microseconds.map(Value::from);
-    let result = trampoline::at(&mut guard, seconds, microseconds);
+
+    let opt1 = opt1.map(Value::from);
+    let opt2 = opt2.map(Value::from);
+    let opt3 = opt3.map(Value::from);
+
+    let result = trampoline::at(&mut guard, seconds, opt1, opt2, opt3);
     match result {
         Ok(value) => value.inner(),
         Err(exception) => error::raise(guard, exception),
--- a/artichoke-backend/src/extn/core/time/offset.rs
+++ b/artichoke-backend/src/extn/core/time/offset.rs
@@ -95,6 +95,8 @@ impl TryConvertMut<Value, Option<Offset>> for Artichoke {
 
 #[cfg(test)]
 mod tests {
+    use bstr::ByteSlice;
+
     use crate::extn::core::time::Offset;
     use crate::test::prelude::*;
 
@@ -118,7 +120,7 @@ mod tests {
         let error = result.unwrap_err();
 
         assert_eq!(error.name(), "ArgumentError");
-        assert_eq!(error.message(), b"unknown keyword: foo".as_slice());
+        assert_eq!(error.message().as_bstr(), b"unknown keyword: foo".as_slice().as_bstr());
     }
 
     #[test]
@@ -132,8 +134,10 @@ mod tests {
 
         assert_eq!(error.name(), "ArgumentError");
         assert_eq!(
-            error.message(),
-            br#"+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: J"#.as_slice()
+            error.message().as_bstr(),
+            br#"+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: J"#
+                .as_slice()
+                .as_bstr()
         );
     }
 
@@ -186,7 +190,10 @@ mod tests {
         let result: Result<Option<Offset>, Error> = interp.try_convert_mut(options);
         let error = result.unwrap_err();
 
-        assert_eq!(error.message(), b"utc_offset out of range".as_slice());
+        assert_eq!(
+            error.message().as_bstr(),
+            b"utc_offset out of range".as_slice().as_bstr()
+        );
         assert_eq!(error.name(), "ArgumentError");
     }
 
@@ -200,7 +207,10 @@ mod tests {
         let result: Result<Option<Offset>, Error> = interp.try_convert_mut(options);
         let error = result.unwrap_err();
 
-        assert_eq!(error.message(), b"utc_offset out of range".as_slice());
+        assert_eq!(
+            error.message().as_bstr(),
+            b"utc_offset out of range".as_slice().as_bstr()
+        );
         assert_eq!(error.name(), "ArgumentError");
 
         // this value is i32::MAX + 1.
@@ -209,7 +219,10 @@ mod tests {
         let result: Result<Option<Offset>, Error> = interp.try_convert_mut(options);
         let error = result.unwrap_err();
 
-        assert_eq!(error.message(), b"utc_offset out of range".as_slice());
+        assert_eq!(
+            error.message().as_bstr(),
+            b"utc_offset out of range".as_slice().as_bstr()
+        );
         assert_eq!(error.name(), "ArgumentError");
     }
 }
--- a/artichoke-backend/src/extn/core/time/offset.rs
+++ b/artichoke-backend/src/extn/core/time/offset.rs
@@ -1,0 +1,78 @@
+use crate::extn::core::time::Offset;
+use crate::extn::prelude::*;
+use crate::convert::{implicitly_convert_to_int, implicitly_convert_to_string};
+use crate::extn::core::symbol::Symbol;
+
+const MAX_FLOAT_OFFSET: f64 = i32::MAX as f64;
+const MIN_FLOAT_OFFSET: f64 = i32::MIN as f64;
+
+impl TryConvertMut<Value, Option<Offset>> for Artichoke {
+    type Error = Error;
+
+    fn try_convert_mut(&mut self, options: Value) -> Result<Option<Offset>, Self::Error> {
+        let hash: Vec<(Value, Value)> = self.try_convert_mut(options)?;
+
+        // Short circuit the offset parameter fetching.
+        if hash.len() == 0 {
+            return Ok(None)
+        }
+
+        // Extract the value from the options hash with the key `:in`,
+        // rejecting any other values as ArgumentError.
+        let mut in_param: Option<Value> = None;
+
+        for (mut key, value) in hash.iter() {
+            let k = unsafe { Symbol::unbox_from_value(&mut key, self)? }.bytes(self);
+            if k == b"in" {
+                in_param = Some(*value)
+            } else {
+                let mut message = b"unknown keyword: ".to_vec();
+                message.extend(k.to_vec());
+                Err(ArgumentError::from(message))?
+            }
+        }
+
+        if let Some(mut in_param) = in_param {
+            match in_param.ruby_type() {
+                Ruby::String => {
+                    let offset_str = unsafe { implicitly_convert_to_string(self, &mut in_param) }?;
+
+                    let offset = Offset::try_from(offset_str)
+                        .map_err(|_| {
+                            let mut message = b"\"+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset: ".to_vec();
+                            message.extend(offset_str.to_vec());
+                            ArgumentError::from(message)
+                        })?;
+
+                    Ok(Some(offset))
+                },
+                Ruby::Float => {
+                    // This impl differs from MRI. MRI supports any float
+                    // value and will set an offset with subsec fractions
+                    // however this is not supported in spinoso_time with
+                    // the `tzrs` feature.
+                    let offset_seconds: f64 = self.try_convert(in_param)?;
+
+                    if !(MIN_FLOAT_OFFSET..=MAX_FLOAT_OFFSET).contains(&offset_seconds) {
+                        Err(ArgumentError::with_message("utc_offset out of range").into())
+                    } else {
+                        Ok(Some(Offset::try_from(offset_seconds as i32)?))
+                    }
+                },
+                _ => {
+                    let offset_seconds = implicitly_convert_to_int(self, in_param)
+                        .and_then(|seconds| {
+                            i32::try_from(seconds)
+                                .map_err(|_| ArgumentError::with_message("utc_offset out of range").into())
+                        })?;
+
+                    Ok(Some(Offset::try_from(offset_seconds)?))
+                }
+            }
+        } else {
+            // The parameter parsing loop will rejected all params except `in`,
+            // so the side affect is that this branch is never reachable.
+            unreachable!("should not have attempted to parse an empty `in` option");
+        }
+    }
+}
--- a/artichoke-backend/src/extn/core/time/offset.rs
+++ b/artichoke-backend/src/extn/core/time/offset.rs
@@ -54,7 +54,7 @@ impl TryConvertMut<Value, Option<Offset>> for Artichoke {
 
                 let offset = Offset::try_from(offset_str).map_err(|_| {
                     let mut message =
-                        b"\"+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset: "
+                        br#"+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: "#
                             .to_vec();
                     message.extend_from_slice(offset_str);
                     ArgumentError::from(message)
@@ -126,7 +126,7 @@ mod tests {
         assert_eq!(error.name(), "ArgumentError");
         assert_eq!(
             error.message(),
-            b"\"+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset: J".as_slice()
+            br#"+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: J"#.as_slice()
         );
     }
 
@@ -134,7 +134,6 @@ mod tests {
     fn provides_an_int_based_offset() {
         let mut interp = interpreter();
 
-        // this value is i32::MIN - 1.
         let options = interp.eval(b"{ in: 3600 }").unwrap();
 
         let result: Option<Offset> = interp.try_convert_mut(options).unwrap();
@@ -145,7 +144,6 @@ mod tests {
     fn provides_a_float_based_offset() {
         let mut interp = interpreter();
 
-        // this value is i32::MIN - 1.
         let options = interp.eval(b"{ in: 3600.0 }").unwrap();
 
         let result: Option<Offset> = interp.try_convert_mut(options).unwrap();
@@ -156,7 +154,6 @@ mod tests {
     fn provides_a_string_based_offset() {
         let mut interp = interpreter();
 
-        // this value is i32::MIN - 1.
         let options = interp.eval(b"{ in: 'A' }").unwrap();
 
         let result: Option<Offset> = interp.try_convert_mut(options).unwrap();
--- a/artichoke-backend/src/extn/core/time/offset.rs
+++ b/artichoke-backend/src/extn/core/time/offset.rs
@@ -1,3 +1,5 @@
+//! Parser for Ruby Time offset parameter to help generate `Time`.
+
 use crate::convert::{implicitly_convert_to_int, implicitly_convert_to_string};
 use crate::extn::core::symbol::Symbol;
 use crate::extn::core::time::Offset;
--- a/artichoke-backend/src/extn/core/time/offset.rs
+++ b/artichoke-backend/src/extn/core/time/offset.rs
@@ -1,7 +1,7 @@
-use crate::extn::core::time::Offset;
-use crate::extn::prelude::*;
 use crate::convert::{implicitly_convert_to_int, implicitly_convert_to_string};
 use crate::extn::core::symbol::Symbol;
+use crate::extn::core::time::Offset;
+use crate::extn::prelude::*;
 
 const MAX_FLOAT_OFFSET: f64 = i32::MAX as f64;
 const MIN_FLOAT_OFFSET: f64 = i32::MIN as f64;
@@ -12,67 +12,71 @@ impl TryConvertMut<Value, Option<Offset>> for Artichoke {
     fn try_convert_mut(&mut self, options: Value) -> Result<Option<Offset>, Self::Error> {
         let hash: Vec<(Value, Value)> = self.try_convert_mut(options)?;
 
-        // Short circuit the offset parameter fetching.
+        // Short circuit. An empty options hash does not error.
+        //
+        // Example:
+        //
+        // ```console
+        // [2.6.3]> Time.at(0, {})
+        // => 1970-01-01 01:00:00 +0100
+        // ```
         if hash.len() == 0 {
-            return Ok(None)
+            return Ok(None);
         }
 
-        // Extract the value from the options hash with the key `:in`,
-        // rejecting any other values as ArgumentError.
-        let mut in_param: Option<Value> = None;
-
-        for (mut key, value) in hash.iter() {
+        // All other keys are rejected.
+        //
+        // Example:
+        // ```console
+        // [2.6.3]> Time.at(0, i: 0)
+        // ArgumentError (unknown keyword: i)
+        // ```
+        for (mut key, _) in hash.iter() {
             let k = unsafe { Symbol::unbox_from_value(&mut key, self)? }.bytes(self);
-            if k == b"in" {
-                in_param = Some(*value)
-            } else {
+            if k != b"in" {
                 let mut message = b"unknown keyword: ".to_vec();
                 message.extend_from_slice(k);
                 Err(ArgumentError::from(message))?
             }
         }
 
-        if let Some(mut in_param) = in_param {
-            match in_param.ruby_type() {
-                Ruby::String => {
-                    let offset_str = unsafe { implicitly_convert_to_string(self, &mut in_param) }?;
+        // Based on the above logic, the only option in the hash is `in`.
+        // >0 keys, and all other keys are rejected).
+        let mut in_value = hash.get(0).expect("Only the `in` parameter should be available").1;
 
-                    let offset = Offset::try_from(offset_str)
-                        .map_err(|_| {
-                            let mut message = b"\"+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset: ".to_vec();
-                            message.extend_from_slice(offset_str);
-                            ArgumentError::from(message)
-                        })?;
+        match in_value.ruby_type() {
+            Ruby::String => {
+                let offset_str = unsafe { implicitly_convert_to_string(self, &mut in_value) }?;
 
-                    Ok(Some(offset))
-                },
-                Ruby::Float => {
-                    // This impl differs from MRI. MRI supports any float
-                    // value and will set an offset with subsec fractions
-                    // however this is not supported in spinoso_time with
-                    // the `tzrs` feature.
-                    let offset_seconds: f64 = self.try_convert(in_param)?;
+                let offset = Offset::try_from(offset_str).map_err(|_| {
+                    let mut message =
+                        b"\"+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset: "
+                            .to_vec();
+                    message.extend_from_slice(offset_str);
+                    ArgumentError::from(message)
+                })?;
 
-                    if !(MIN_FLOAT_OFFSET..=MAX_FLOAT_OFFSET).contains(&offset_seconds) {
-                        Err(ArgumentError::with_message("utc_offset out of range").into())
-                    } else {
-                        Ok(Some(Offset::try_from(offset_seconds as i32)?))
-                    }
-                },
-                _ => {
-                    let offset_seconds = implicitly_convert_to_int(self, in_param)
-                        .and_then(|seconds| {
-                            i32::try_from(seconds)
-                                .map_err(|_| ArgumentError::with_message("utc_offset out of range").into())
-                        })?;
+                Ok(Some(offset))
+            }
+            Ruby::Float => {
+                // This impl differs from MRI. MRI supports any float value and
+                // will set an offset with subsec fractions however this is not
+                // supported in spinoso_time with the `tzrs` feature.
+                let offset_seconds: f64 = self.try_convert(in_value)?;
 
-                    Ok(Some(Offset::try_from(offset_seconds)?))
+                if !(MIN_FLOAT_OFFSET..=MAX_FLOAT_OFFSET).contains(&offset_seconds) {
+                    Err(ArgumentError::with_message("utc_offset out of range").into())
+                } else {
+                    Ok(Some(Offset::try_from(offset_seconds as i32)?))
                 }
             }
-        } else {
-            // The parameter parsing loop will rejected all params except `in`,
-            // so the side affect is that this branch is never reachable.
-            unreachable!("should not have attempted to parse an empty `in` option");
+            _ => {
+                let offset_seconds = implicitly_convert_to_int(self, in_value).and_then(|seconds| {
+                    i32::try_from(seconds).map_err(|_| ArgumentError::with_message("utc_offset out of range").into())
+                })?;
+
+                Ok(Some(Offset::try_from(offset_seconds)?))
+            }
         }
     }
 }
--- a/artichoke-backend/src/extn/core/time/offset.rs
+++ b/artichoke-backend/src/extn/core/time/offset.rs
@@ -35,6 +35,14 @@ impl TryConvertMut<Value, Option<Offset>> for Artichoke {
         // [2.6.3]> Time.at(0, i: 0)
         // ArgumentError (unknown keyword: i)
         // ```
+        //
+        // FIXME: In Ruby 3.1.2, this exception message formats the symbol with
+        // `Symbol#inspect`:
+        //
+        // ```console
+        // [3.1.2] > Time.at(0, i: 0)
+        // <internal:timev>:270:in `at': unknown keyword: :i (ArgumentError)
+        // ```
         for (mut key, _) in &hash {
             let k = unsafe { Symbol::unbox_from_value(&mut key, self)? }.bytes(self);
             if k != b"in" {
@@ -54,8 +62,7 @@ impl TryConvertMut<Value, Option<Offset>> for Artichoke {
 
                 let offset = Offset::try_from(offset_str).map_err(|_| {
                     let mut message =
-                        br#"+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: "#
-                            .to_vec();
+                        br#"+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: "#.to_vec();
                     message.extend_from_slice(offset_str);
                     ArgumentError::from(message)
                 })?;
--- a/artichoke-backend/src/extn/core/time/offset.rs
+++ b/artichoke-backend/src/extn/core/time/offset.rs
@@ -80,3 +80,128 @@ impl TryConvertMut<Value, Option<Offset>> for Artichoke {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::extn::core::time::Offset;
+    use crate::test::prelude::*;
+
+    #[test]
+    fn no_options_does_not_raise() {
+        let mut interp = interpreter();
+
+        let options = interp.eval(b"{}").unwrap();
+
+        let offset: Option<Offset> = interp.try_convert_mut(options).unwrap();
+        assert_eq!(offset, None);
+    }
+
+    #[test]
+    fn raises_on_keys_except_in() {
+        let mut interp = interpreter();
+
+        let options = interp.eval(b"{ foo: 'bar' }").unwrap();
+
+        let result: Result<Option<Offset>, Error> = interp.try_convert_mut(options);
+        let error = result.unwrap_err();
+
+        assert_eq!(error.name(), "ArgumentError");
+        assert_eq!(error.message(), b"unknown keyword: foo".as_slice());
+    }
+
+    #[test]
+    fn raises_on_invalid_timezone_string() {
+        let mut interp = interpreter();
+
+        let options = interp.eval(b"{ in: 'J' }").unwrap();
+
+        let result: Result<Option<Offset>, Error> = interp.try_convert_mut(options);
+        let error = result.unwrap_err();
+
+        assert_eq!(error.name(), "ArgumentError");
+        assert_eq!(
+            error.message(),
+            b"\"+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset: J".as_slice()
+        );
+    }
+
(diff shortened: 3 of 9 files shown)


artichoke-backend/src/extn/core/time/mod.rs
```diff
@@ -21,13 +21,18 @@
...
 #[doc(inline)]
-pub use spinoso_time::chrono::Time;
+pub use spinoso_time::tzrs::*;
 
 impl HeapAllocatedData for Time {
     const RUBY_TYPE: &'static str = "Time";
 }
+
+pub(crate) fn convert_time_error_to_argument_error(error: TimeError) -> ArgumentError {
+    ArgumentError::from(format!("{:?}", error))
```
b-n, line 37: This works but it also feels wrong. Is there a more idiomatic way of achieving the conversion from `TimeError` => `ArgumentError` ?  (`impl from<TimeError> for ArgumentError` doesn't work)

artichoke-backend/src/extn/core/time/mod.rs
```diff
@@ -21,13 +21,18 @@
...
 #[doc(inline)]
-pub use spinoso_time::chrono::Time;
+pub use spinoso_time::tzrs::*;
 
 impl HeapAllocatedData for Time {
     const RUBY_TYPE: &'static str = "Time";
 }
+
+pub(crate) fn convert_time_error_to_argument_error(error: TimeError) -> ArgumentError {
+    ArgumentError::from(format!("{}", error))
```
b-n: Is there any better way of doing this?

artichoke-backend/src/extn/core/time/mruby.rs
```diff
@@ -112,11 +112,15 @@ unsafe extern "C" fn time_self_now(mrb: *mut sys::mrb_state, _slf: sys::mrb_valu
 }
 
 unsafe extern "C" fn time_self_at(mrb: *mut sys::mrb_state, _slf: sys::mrb_value) -> sys::mrb_value {
-    let (seconds, microseconds) = mrb_get_args!(mrb, required = 1, optional = 1);
+    let (seconds, opt1, opt2, opt3) = mrb_get_args!(mrb, required = 1, optional = 3);
```
b-n, line 115: Could quite possibly be done with `args_rest`, but this will get a more strict error (and more in line with MRI) if I'm not mistaken.

Ref: https://ruby-doc.org/core-2.6.3/Time.html#method-c-at

artichoke-backend/src/extn/core/time/mruby.rs
```diff
@@ -19,7 +19,7 @@ pub fn init(interp: &mut Artichoke) -> InitializeResult<()> {
         .value_is_rust_object()
         // Constructor
         .add_self_method("now", time_self_now, sys::mrb_args_none())?
-        .add_self_method("at", time_self_at, sys::mrb_args_req(1))?
+        .add_self_method("at", time_self_at, sys::mrb_args_req(1) | sys::mrb_args_opt(3))?
```
b-n: My understanding is this should work. However it funnily enough seems to not do anything regarding the validation. Is it expected to, or is this just for documentation to some degree?

artichoke-backend/src/extn/core/time/mod.rs
lopopolo: `impl From<tzrs::TimeError> for Error`

Then question mark will work. (Sorry for being brief, on mobile)

artichoke-backend/src/extn/core/time/mruby.rs
lopopolo: mruby only raises an ArgumentError from the VM when there is an arg mismatch for methods that take no args.

It is the call to the `mrb_get_args!` that triggers arity checking.

See:

- https://github.com/mruby/mruby/issues/4688

lopopolo:
Hi @b-n I've seen this but haven't had time to sit down and give a proper review yet. I took a glance and things look like they came together nicely. I hope to get to this sometime this week.

lopopolo:
Heads up, despite the crate being named `artichoke-backend` this change should be tagged `B-mruby`.

The `B-` tags refer to the VM implementation. `B-Artichoke` is a hypothetical pure Rust VM that we'll write in house.

artichoke-backend/src/extn/core/time/mod.rs
b-n: Ahhh, I tried this, got some failure, but it's because I needed an `.into()` after constructing the ArgumentError - whoops.

artichoke-backend/src/extn/core/time/mruby.rs
b-n: Hrrrmmm, I must have been thinking about something else when I wrote this comment, my bad! I realized just now that it's working as expected :+1:

lopopolo: there's another sys args helper that does this operation: https://artichoke.github.io/artichoke/artichoke_backend/sys/fn.mrb_args_req_and_opt.html

```suggestion
        .add_self_method("at", time_self_at, sys:: mrb_args_req_and_opt(1, 3))?
```

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -1,35 +1,151 @@
...
+    subsec_type.map_or(Ok(MICROS_IN_NANO), |mut value| {
+        let subsec_symbol = unsafe { Symbol::unbox_from_value(&mut value, interp)? }.bytes(interp);
+        match subsec_symbol {
+            b"milliseconds" => Ok(MILLIS_IN_NANO),
+            b"usec" => Ok(MICROS_IN_NANO),
+            b"nsec" => Ok(NANOS_IN_NANO),
+            _ => Err(ArgumentError::with_message("unexpected unit. expects :milliseconds, :usec, :nsec").into()),
+        }
+    })
+}
```
lopopolo: To make this more consistent with the rest of the codebase, can we convert these `i64` constants into an enum and change this free function to a `TryConvertMut` impl like this one in `SecureRandom`?

https://github.com/artichoke/artichoke/blob/fb65bedcd76245121e0a1a3f2b10192d068de24c/artichoke-backend/src/extn/stdlib/securerandom/mod.rs#L71-L98

I'm thinking something like this:

```rust
enum SubsecMultiplier {
    Millis,
    Micros,
    Nanos,
}

impl SubsecMultiplier {
    const fn as_nanos(self) -> i64 {
        match self {
            Self::Millis => MILLIS_IN_NANO,
            // etc...
        }
    }
}

impl TryConvertMut<Option<Value>, SubsecMultiplier> for Artichoke {
    type Error = Error;

    fn try_convert_mut(&mut self, max: Option<Value>) -> Result<SubsecMultiplier, Self::Error> {
        todo!();
    }
}
```

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -1,35 +1,151 @@
...
+        Err(ArgumentError::with_message(
+            "+HH:MM, -HH:MM, UTC, A..I,K..Z, or a signed number of seconds expected for utc_offset",
+        )
+        .into())
+    }
+}
 
+// Check a Ruby Value Hash for the appropriate values to create and Offset
+// for a _time_.
+fn offset_from_options(interp: &mut Artichoke, options: Value) -> Result<Offset, Error> {
```
lopopolo: I'd love to see this free function similarly converted but I'm not sure if the orphan rule / trait coherence will allow us to do that without a newtype around the foreign `Offset` type.

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -1,35 +1,151 @@
...
+    let hash: Vec<(Value, Value)> = interp.try_convert_mut(options)?;
```
lopopolo: nifty! I see you found some of the more obscure `TryConvertMut` impls!

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -1,35 +1,151 @@
...
+        .into_iter()
+        .map(|(k, v)| {
+            let key = unsafe { Symbol::unbox_from_value(&mut k.clone(), interp)? }.bytes(interp);
+            if key == b"in" {
+                Ok(v)
+            } else {
+                Err(ArgumentError::with_message("unknown keyword"))?
+            }
+        })
+        .collect::<Result<Vec<Value>, Error>>()?;
```
lopopolo: I think this might be more appropriately implemented with `Iterator::find_map` -- https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map

It will help avoid an allocation with the collect and collapse the error handling code and the below match into an:

```rust
if let Some(tz) = x {
    offset_from_value(tz)
} else {
    Err(...)
}
```

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -1,35 +1,151 @@
...
+    // ```
+    let (subsec, subsec_type, options) = match (
+        first.map(|opt| opt.ruby_type()),
+        second.map(|opt| opt.ruby_type()),
+        third.map(|opt| opt.ruby_type()),
+    ) {
+        (None, None, None) => (None, None, None),
+        (Some(Ruby::Hash), None, None) => (None, None, first),
+        (Some(Ruby::Fixnum), None, None) => (first, None, None),
+        (Some(_), None, None) => Err(ArgumentError::with_message("expected a number"))?,
```
lopopolo: I think in this case we need to do the "implicitly convert to int" dance:

```
[3.1.2] > Time.at ""
<internal:timev>:272:in `at': can't convert String into an exact number (TypeError)
        from (irb):1:in `<main>'
[3.1.2] > class X; def to_int; 5; end; end
=> :to_int
[3.1.2] > Time.at(X.new)
=> 1969-12-31 16:00:05 -0800
[3.1.2] > class Y; def to_int; ""; end; end
=> :to_int
[3.1.2] > Time.at(Y.new)
<internal:timev>:272:in `at': can't convert Y into an exact number (TypeError)
        from (irb):5:in `<main>'
[3.1.2] > class Z; def to_int; 5.0; end; end
=> :to_int
[3.1.2] > Time.at(Z.new)
<internal:timev>:272:in `at': can't convert Z into an exact number (TypeError)
        from (irb):7:in `<main>'
```

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -1,35 +1,151 @@
...
+        (Some(Ruby::Fixnum), Some(Ruby::Hash), None) => (first, None, second),
+        (Some(Ruby::Fixnum), Some(Ruby::Symbol), None) => (first, second, None),
+        (Some(Ruby::Fixnum), Some(_), None) => Err(ArgumentError::with_message(
+            "expected one of [:milliseconds, :usec, :nsec]",
```
lopopolo: we can add this as a todo but it looks like the error message is `"unexpected unit: #{x.inspect}"`:

```
[3.1.2] > Time.at(1, 1, :xxx)
<internal:timev>:274:in `at': unexpected unit: xxx (ArgumentError)
        from (irb):12:in `<main>'
[3.1.2] > Time.at(1, 1, Object.new)
<internal:timev>:274:in `at': unexpected unit: #<Object:0x0000000111579b78> (ArgumentError)
        from (irb):13:in `<main>'
[3.1.2] > Time.at(1, 1, "nsec")
<internal:timev>:274:in `at': unexpected unit: nsec (ArgumentError)
        from (irb):14:in `<main>'
```

artichoke-backend/src/extn/core/time/trampoline.rs
lopopolo: it looks like the implicit convert to int step is done below, should we collapse this match arm with the one above?

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -180,37 +296,46 @@ pub fn to_array(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
...
 
-pub fn plus(interp: &mut Artichoke, time: Value, other: Value) -> Result<Value, Error> {
-    let _ = interp;
-    let _ = time;
-    let _ = other;
-    Err(NotImplementedError::new().into())
+pub fn plus(interp: &mut Artichoke, mut time: Value, mut other: Value) -> Result<Value, Error> {
+    let time = unsafe { Time::unbox_from_value(&mut time, interp)? };
+    if unsafe { Time::unbox_from_value(&mut other, interp) }.is_ok() {
+        Err(TypeError::with_message("time + time?").into())
```
lopopolo, line 269:
```suggestion
        // ```console
        // [3.1.2] > Time.now + Time.now
        // (irb):15:in `+': time + time? (TypeError)
        //         from (irb):15:in `<main>'
        //         from /usr/local/var/rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
        //         from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `load'
        //         from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `<main>'
        // ```
        Err(TypeError::with_message("time + time?").into())
```

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -180,37 +296,46 @@ pub fn to_array(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
...
+    } else if let Ok(other) = implicitly_convert_to_int(interp, other) {
```
lopopolo: this makes me realize that our `implicitly_convert_to_xx` helpers don't distinguish between "conversion failed" and "conversion raised an Exception".

No action for now, but I'll see if I can put up a PR.

lopopolo (changes requested):

artichoke-backend/src/extn/core/time/trampoline.rs
lopopolo: Actually after taking a look, I think in practice, the implicit conversion to integer should be the last branch in the if else chain and errors from it should be propagated with `?`.

b-n:
@lopopolo I'll pick this one up again shortly, sorry for taking a while to get back around to it!

artichoke-backend/src/extn/core/time/trampoline.rs
b-n: I just discovered `Time.at(1, 0,75)` is a thing:

```
irb(main):001:0> Time.at(1, 0.75)
=> 1970-01-01 01:00:01.00000075 +0100
```

Which is 0.75 micros (the default if nil), or 750 nanos. Of course I had the naive "this must just be an integer" kind of thoughts, but of course it's the `to_int` dance too! :dancers: 

```
irb(main):001:0> class A; def to_int; 1000; end; end
=> :to_int
irb(main):002:0> Time.at(0, A.new).tv_nsec
=> 1000000
```

How do you feel about:

```rust
struct Subsec(i64);

impl TryConvertMut<(Option<Value>, Option<Value>, Subsec> for Artichoke {
    type Error = Error;

    fn try_convert_mut(&mut self, subsec_and_type: (Option<Value>, Option<Value>)) -> Result<Subsec, Self::Error> {
        todo!();
    }
}
```

:point_up: This would take both the subsecond param, and the optional subsecond type.

Could also use the `SubsecMultiplier` in conjunction with this too of course, but it feels like this logic needs to be pulled out a little.

lopopolo: ooof fun times with implicit conversions. This looks like a good approach to me. ooc, did you find this when playing with the specs?

b-n: Yep! The specs lead me to it, so they are actually quite helpful! I apparently missed the 5th line of implementation in the ruby docs, but I must admit even they are light on this (I probably should have interpreted `with_frac`, but completely overlooked it): https://ruby-doc.org/core-2.6.3/Time.html#method-c-at

b-n:
@lopopolo Converting this one back to draft since there's still some reasonably sized changes that I need to check. https://github.com/artichoke/artichoke/pull/1956/commits/3dd1b6b6db84c04850cc4b4a71823f9c2bed8846 might be pique your interest :sweat_smile:

b-n:
Also, do you have a preference on tabled vs. non tabled tests? (below are the same tests)

Table tests:

```rust
    #[test]
    fn subsec_millis_tabled() {
        let mut interp = interpreter();

        let table: HashMap<&[u8], (i64, u32)> = HashMap::from([
            (b"-1001".as_slice(), (-2, 999_000_000)),
            (b"-1000".as_slice(), (-2, 0)),
            (b"-999".as_slice(), (-1, 1_000_000)),
            (b"-1".as_slice(), (-1, 999_000_000)),
            (b"0".as_slice(), (0, 0)),
            (b"999".as_slice(), (0, 999_000_000)),
            (b"1000".as_slice(), (1, 0)),
            (b"1001".as_slice(), (1, 1_000_000))
        ]);

        let unit = b":milliseconds";

        for (input, expectation) in table {
            let result = subsec(&mut interp, (Some(input), Some(unit))).unwrap();
            assert_eq!(
                result.to_tuple(),
                expectation,
                "Expected TryConvertMut<(Some({}), Some({})), Result<Subsec>>, to return {} secs, {} nanos",
                input.as_bstr(),
                unit.as_bstr(),
                expectation.0,
                expectation.1
            );
        }
    }
```

Not so tabled:

```rust
    #[test]
    fn subsec_millis() {
        let mut interp = interpreter();

        let result = subsec(&mut interp, (Some(b"-1001"), Some(b":milliseconds"))).unwrap();
        assert_eq!(result.to_tuple(), (-2, -999_000_000));

        let result = subsec(&mut interp, (Some(b"-1000"), Some(b":milliseconds"))).unwrap();
        assert_eq!(result.to_tuple(), (-2, 0));

        let result = subsec(&mut interp, (Some(b"-999"), Some(b":milliseconds"))).unwrap();
        assert_eq!(result.to_tuple(), (-1, 1_000_000));

        let result = subsec(&mut interp, (Some(b"-1"), Some(b":milliseconds"))).unwrap();
        assert_eq!(result.to_tuple(), (-1, 999_000_000));

        let result = subsec(&mut interp, (Some(b"0"), Some(b":milliseconds"))).unwrap();
        assert_eq!(result.to_tuple(), (0, 0));

        let result = subsec(&mut interp, (Some(b"999"), Some(b":milliseconds"))).unwrap();
        assert_eq!(result.to_tuple(), (0, 999_000_000));

        let result = subsec(&mut interp, (Some(b"1000"), Some(b":milliseconds"))).unwrap();
        assert_eq!(result.to_tuple(), (1, 0));

        let result = subsec(&mut interp, (Some(b"1001"), Some(b":milliseconds"))).unwrap();
        assert_eq!(result.to_tuple(), (1, 1_000_000));
    }
```

lopopolo:
> Also, do you have a preference on tabled vs. non tabled tests? (below are the same tests)

I've been leaning toward table driven tests lately, especially for things like parsers, and have been liking how clean things look. I haven't been using `HashMap`s for this though, instead preferring an array of tuples, like:

https://github.com/artichoke/artichoke/blob/995b8caf09dfbfba84b1ad001e024fe427d432a6/scolapasta-int-parse/src/error.rs#L348-L362

b-n:
Ah yes! I apparently reverted back to an old habit. That Hashmap is redundant here :-D

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,78 @@
...
+        // rejecting any other values as ArgumentError.
+        let mut in_param: Option<Value> = None;
+
+        for (mut key, value) in hash.iter() {
+            let k = unsafe { Symbol::unbox_from_value(&mut key, self)? }.bytes(self);
+            if k == b"in" {
+                in_param = Some(*value)
+            } else {
+                let mut message = b"unknown keyword: ".to_vec();
+                message.extend(k.to_vec());
```
lopopolo: We can avoid an allocation here.

```suggestion
                message.extend_from_slice(k);
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        if let Some(mut in_param) = in_param {
+            match in_param.ruby_type() {
+                Ruby::String => {
+                    let offset_str = unsafe { implicitly_convert_to_string(self, &mut in_param) }?;
+
+                    let offset = Offset::try_from(offset_str)
+                        .map_err(|_| {
+                            let mut message = b"\"+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset: ".to_vec();
+                            message.extend(offset_str.to_vec());
```
lopopolo: Ditto

```suggestion
                            message.extend_from_slice(offset_str);
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,78 @@
...
+        let mut in_param: Option<Value> = None;
+
+        for (mut key, value) in hash.iter() {
+            let k = unsafe { Symbol::unbox_from_value(&mut key, self)? }.bytes(self);
+            if k == b"in" {
+                in_param = Some(*value)
+            } else {
+                let mut message = b"unknown keyword: ".to_vec();
+                message.extend(k.to_vec());
+                Err(ArgumentError::from(message))?
```
lopopolo: I think clippy won't like this:

```suggestion
                return Err(ArgumentError::from(message).into());
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+
+        if let Some(mut in_param) = in_param {
```
lopopolo: to avoid some rightward drift here we can rewrite this as:

```suggestion
        let mut in_param = in_param.unwrap_or_else(|| unreachable!("..."));
```

lopopolo: or `in_param.expect("...")`

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -135,9 +94,11 @@ pub fn at(
         .checked_add(subsec_secs)
         .ok_or(ArgumentError::with_message("Time too large"))?;
 
-    let offset = match options {
-        Some(options) => offset_from_options(interp, options)?,
-        _ => Offset::local(),
+    let offset: Offset = if let Some(options) = options {
+        let offset: Option<Offset> = interp.try_convert_mut(options)?;
+        offset.unwrap_or(Offset::local())
```
lopopolo: since `Offset::local` is potentially an expensive call, can we use `offset.unwrap_or_else(Offset::local)` instead?

lopopolo (commented):
some drive by comments, I know you're still working on this

artichoke-backend/src/extn/core/time/offset.rs
b-n: Good catch!

artichoke-backend/src/extn/core/time/trampoline.rs
b-n: I've done some fairly big changes to this part. Going to resolve this because of those changes

artichoke-backend/src/extn/core/time/subsec.rs
```diff
@@ -0,0 +1,557 @@
...
+///! and nanos).
+use crate::extn::prelude::*;
+
+const NANOS_IN_SECOND: i64 = 1_000_000_000;
+
+const MILLIS_IN_NANO: i64 = 1_000_000;
+const MICROS_IN_NANO: i64 = 1_000;
+const NANOS_IN_NANO: i64 = 1;
+
+#[allow(clippy::cast_precision_loss)]
```
b-n, line 18: There is a not insignificant amount of casting going on in this class - and I'm not sure if it's avoidable. I'm throwing a comment here, because I'm sure there are better solutions than what I've done here, so I'm all :ear:s

artichoke-backend/src/extn/core/time/subsec.rs
```diff
@@ -0,0 +1,557 @@
...
+const MIN_FLOAT_SECONDS: f64 = i64::MIN as f64;
+#[allow(clippy::cast_precision_loss)]
+const MAX_FLOAT_SECONDS: f64 = i64::MAX as f64;
+const MIN_FLOAT_NANOS: f64 = 0.0;
+#[allow(clippy::cast_precision_loss)]
+const MAX_FLOAT_NANOS: f64 = NANOS_IN_SECOND as f64;
+
+enum SubsecMultiplier {
```
b-n, line 26: This enum is luckily not exposed anywhere (yet). Is it worthwhile throwing docs on it?

artichoke-backend/src/extn/core/time/subsec.rs
```diff
@@ -0,0 +1,562 @@
...
+        if let Some(subsec) = subsec {
+            let multiplier: SubsecMultiplier = self.try_convert_mut(subsec_unit)?;
+            let multiplier_nanos = multiplier.as_nanos();
+            // `subsec` represents the user provided value in `subsec_unit`
+            // resolution. The base used to derive the number of seconds is
+            // based on the `subsec_unit`. e.g. `1_001` milliseconds is 1
+            // second, and `1_000_000` nanoseconds.
+            let seconds_base = NANOS_IN_SECOND / multiplier_nanos;
+
+            if subsec.ruby_type() == Ruby::Float {
```
b-n, line 105: The number of casts below is somewhat worrisome, but I'm not sure of a better way to achieve them.

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -1,35 +1,110 @@
...
+    // ```
+    //
+    // The below logic:
+    // - ensures the third parameter is a Ruby::Hash if provided.
+    // - if third param is not options, check the second paramter, if it is a
+    //   Ruby::Hash then assume this is the options hash, and clear out the
+    //   second parameter.
+    // - if second param is not options, check the first param, if it is a
+    //   Ruby::Hash then assume this is the options hash, and clear out the
+    //   first parameter.
```
b-n, line 66: It appears that MRI supports this kind of syntax is slightly easier to parse way than mruby. I'm not sure where I spotted it, but it MRI had something around an `options hash`, where as ruby parses the named parameters as the last provided parameter instead. :shrug:.

I have a feeling the below might look a lot nicer with a `match`, but i'm open to ideas!

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -1,35 +1,110 @@
...
+            }
+        } else {
+            None
+        }
     }
+
+    let subsec: Subsec = interp.try_convert_mut((subsec, subsec_unit))?;
+    let (subsec_secs, subsec_nanos) = subsec.to_tuple();
+
+    let seconds = implicitly_convert_to_int(interp, seconds)?
```
b-n, line 94: This PR is already getting a bit large, but I know that this line isn't 100% correct either - and that we need to do the ol `float/to_i` dance.

```console
[3.1.2]> Time.at(1.5, 1500000, :usec).to_f
=> 3.0
```

:point_up: Yep, 1.5seconds + 1.5seconds = 3 seconds!

Better to do in this PR or in the next?

b-n:
@lopopolo Your input on this stuff would be much appreciated (I've thrown some comments in places). I'm happy for you to push commits if you have some fancy ideas on how to fix some of these things.

lopopolo:
pushed a rebase to fix a merge conflict

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+            if k != b"in" {
+                let mut message = b"unknown keyword: ".to_vec();
+                message.extend_from_slice(k);
+                Err(ArgumentError::from(message))?;
+            }
+        }
+
+        // Based on the above logic, the only option in the hash is `in`.
+        // >0 keys, and all other keys are rejected).
+        let mut in_value = hash.get(0).expect("Only the `in` parameter should be available").1;
```
lopopolo, line 57: I think clippy will whine about this.

```suggestion
        let mut in_value = hash.first().expect("Only the `in` parameter should be available").1;
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        // All other keys are rejected.
+        //
+        // Example:
+        // ```console
+        // [2.6.3]> Time.at(0, i: 0)
+        // ArgumentError (unknown keyword: i)
+        // ```
+        for (mut key, _) in &hash {
+            let k = unsafe { Symbol::unbox_from_value(&mut key, self)? }.bytes(self);
```
lopopolo, line 47: it looks like we need to handle non-`Symbol` args in the same way:

```
[3.1.2] > Time.at(12, 'in' => 'Z')
<internal:timev>:270:in `at': unknown keyword: "in" (ArgumentError)
        from (irb):1:in `<main>'
[3.1.2] > Time.at(12, Object.new => 'Z')
<internal:timev>:270:in `at': unknown keyword: #<Object:0x00000001091c7d70> (ArgumentError)
        from (irb):2:in `<main>'
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        // All other keys are rejected.
+        //
+        // Example:
+        // ```console
+        // [2.6.3]> Time.at(0, i: 0)
+        // ArgumentError (unknown keyword: i)
+        // ```
+        for (mut key, _) in &hash {
```
lopopolo, line 46: Agreed per the conversation we had on Discord. This looks like we should extract this to the `convert` module.

`String::new` wants the same functionality with its `capacity:` kwarg:

```
[3.1.2] > String.new(capacity: 10)
=> ""
[3.1.2] > String.new(capa: 10)
(irb):4:in `initialize': unknown keyword: :capa (ArgumentError)
        from (irb):4:in `new'
        from (irb):4:in `<main>'
[3.1.2] > String.new(capa: 10)
^C
[3.1.2] > String.new(capacity: 10, nine: 10)
(irb):5:in `initialize': unknown keyword: :nine (ArgumentError)
        from (irb):5:in `new'
        from (irb):5:in `<main>'
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+            let k = unsafe { Symbol::unbox_from_value(&mut key, self)? }.bytes(self);
+            if k != b"in" {
+                let mut message = b"unknown keyword: ".to_vec();
+                message.extend_from_slice(k);
```
lopopolo, line 50: this should be the result of `key.inspect`.

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+
+                if (MIN_FLOAT_OFFSET..=MAX_FLOAT_OFFSET).contains(&offset_seconds) {
+                    #[allow(clippy::cast_possible_truncation)]
+                    Ok(Some(Offset::try_from(offset_seconds as i32)?))
+                } else {
+                    Err(ArgumentError::with_message("utc_offset out of range").into())
+                }
+            }
+            _ => {
+                let offset_seconds = implicitly_convert_to_int(self, in_value).and_then(|seconds| {
```
lopopolo, line 86: we need to short circuit return from `implicitly_convert_to_int` here to handle this case:

```
[3.1.2] > class A; def to_int; raise 'bam'; end; end
=> :to_int
[3.1.2] > Time.at(12, in: A.new)
(irb):6:in `to_int': bam (RuntimeError)
        from <internal:timev>:274:in `at'
        from (irb):7:in `<main>'
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+                Err(ArgumentError::from(message))?;
+            }
+        }
+
+        // Based on the above logic, the only option in the hash is `in`.
+        // >0 keys, and all other keys are rejected).
+        let mut in_value = hash.get(0).expect("Only the `in` parameter should be available").1;
+
+        match in_value.ruby_type() {
+            Ruby::String => {
```
lopopolo, line 60: This doesn't look quite right.

```
[3.1.2] > class B; def to_str; 'B'; end; end
=> :to_str
[3.1.2] > Time.at(12, in: B.new)
=> 1970-01-01 02:00:12 +0200
[3.1.2] > class C; def to_str; raise 'crash'; end; end
=> :to_str
[3.1.2] > Time.at(12, in: C.new)
(irb):10:in `to_str': crash (RuntimeError)
        from <internal:timev>:274:in `at'
        from (irb):11:in `<main>'
[3.1.2] > class D; def to_int; raise 'dint'; end; def to_str; raise 'dstr'; end; end
=> :to_str
[3.1.2] > Time.at(12, in: D.new)
(irb):12:in `to_str': dstr (RuntimeError)
        from <internal:timev>:274:in `at'
        from (irb):13:in `<main>'
```

It looks like MRI prefers to implicitly convert through `#to_str` if it exists and otherwise will convert through `#to_int`. We can check for this by doing a `Value::respond_to` check.

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+                let offset_str = unsafe { implicitly_convert_to_string(self, &mut in_value) }?;
+
+                let offset = Offset::try_from(offset_str).map_err(|_| {
+                    let mut message =
+                        b"\"+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset: "
```
lopopolo: we can use a raw string here to make it a bit more readable.

```suggestion
                        br#"+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: "#
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        let options = interp.eval(b"{ in: 'J' }").unwrap();
+
+        let result: Result<Option<Offset>, Error> = interp.try_convert_mut(options);
+        let error = result.unwrap_err();
+
+        assert_eq!(error.name(), "ArgumentError");
+        assert_eq!(
+            error.message(),
+            b"\"+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset: J".as_slice()
```
lopopolo:
```suggestion
            br#"+HH:MM", "-HH:MM", "UTC" or "A".."I","K".."Z" expected for utc_offset: J"#.as_slice()
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+        let mut interp = interpreter();
+
+        let options = interp.eval(b"{ in: 'J' }").unwrap();
+
+        let result: Result<Option<Offset>, Error> = interp.try_convert_mut(options);
+        let error = result.unwrap_err();
+
+        assert_eq!(error.name(), "ArgumentError");
+        assert_eq!(
+            error.message(),
```
lopopolo: turning byte strings into `BStr`s with `.as_bstr()` makes the `assert_eq!` error messages easier to read by printing out a UTF-8-ish form instead of a sequence of `u8` bytes.

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+                })?;
+
+                Ok(Some(Offset::try_from(offset_seconds)?))
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
lopopolo, line 97: These tests are great. Not many custom implementations of `TryConvertMut` have tests and it is good that this is changing.

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        let result: Option<Offset> = interp.try_convert_mut(options).unwrap();
+        assert_eq!(result.unwrap(), Offset::fixed(3600).unwrap());
+    }
+
+    #[test]
+    fn provides_a_string_based_offset() {
+        let mut interp = interpreter();
+
+        // this value is i32::MIN - 1.
```
lopopolo: It looks like this comment is a copy-paste mixup from the `raises_on_float_out_of_range` test.

```suggestion
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        let result: Option<Offset> = interp.try_convert_mut(options).unwrap();
+        assert_eq!(result.unwrap(), Offset::fixed(3600).unwrap());
+    }
+
+    #[test]
+    fn provides_a_float_based_offset() {
+        let mut interp = interpreter();
+
+        // this value is i32::MIN - 1.
```
lopopolo: It looks like this comment is a copy-paste mixup from the `raises_on_float_out_of_range` test.

```suggestion
```

artichoke-backend/src/extn/core/time/offset.rs
```diff
@@ -0,0 +1,211 @@
...
+            error.message(),
+            b"\"+HH:MM\", \"-HH:MM\", \"UTC\" or \"A\"..\"I\",\"K\"..\"Z\" expected for utc_offset: J".as_slice()
+        );
+    }
+
+    #[test]
+    fn provides_an_int_based_offset() {
+        let mut interp = interpreter();
+
+        // this value is i32::MIN - 1.
```
lopopolo: It looks like this comment is a copy-paste mixup from the `raises_on_float_out_of_range` test.

```suggestion
```

artichoke-backend/src/extn/core/time/subsec.rs
lopopolo: Unfortunately there's not a much better way to do this on stable today with "checked"-style APIs.

There are _unchecked_ APIs:

- [`f64::to_int_unchecked`](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.to_int_unchecked)

Relevant tracking issues are here:

- https://internals.rust-lang.org/t/pre-rfc-add-explicitly-named-numeric-conversion-apis/11395
- https://github.com/rust-lang/rust/pull/66841
- https://github.com/rust-lang/rust/issues/67057
- https://github.com/rust-lang/rust/issues/67058

artichoke-backend/src/extn/core/time/subsec.rs
lopopolo: I don't think so. Internal types haven't been getting docs historically.

artichoke-backend/src/extn/core/time/subsec.rs
```diff
@@ -0,0 +1,562 @@
...
+        }
+    }
+}
+
+impl TryConvertMut<Option<Value>, SubsecMultiplier> for Artichoke {
+    type Error = Error;
+
+    fn try_convert_mut(&mut self, subsec_type: Option<Value>) -> Result<SubsecMultiplier, Self::Error> {
+        if let Some(mut subsec_type) = subsec_type {
+            let subsec_symbol = unsafe { Symbol::unbox_from_value(&mut subsec_type, self)? }.bytes(self);
```
lopopolo, line 48: if we fail to unpack as a symbol, we need to return an `ArgumentError` instead of the `TypeError` this unboxing will return:

```
[3.1.2] > Time.at(12, 15000, "usec")
<internal:timev>:274:in `at': unexpected unit: usec (ArgumentError)
        from (irb):1:in `<main>'
[3.1.2] > Time.at(12, 15000, Object.new)
<internal:timev>:274:in `at': unexpected unit: #<Object:0x000000010e0e7b80> (ArgumentError)
        from (irb):2:in `<main>'
```

artichoke-backend/src/extn/core/time/subsec.rs
```diff
@@ -0,0 +1,562 @@
...
+                // Time. MRI uses `to_r` for subsec calculation on floats
+                // subsec nanos, and this could result in different values.
+
+                let subsec: f64 = self.try_convert(subsec)?;
+
+                if subsec.is_nan() {
+                    return Err(FloatDomainError::with_message("NaN").into());
+                }
+                if subsec.is_infinite() {
+                    return Err(FloatDomainError::with_message("Infinity").into());
```
lopopolo, line 119: negative infinity gets a different exception message:

```
[3.1.2] > Time.at(12, -Float::INFINITY, :usec)
<internal:timev>:274:in `to_r': -Infinity (FloatDomainError)
```

lopopolo: also fun, it looks like this converts through `Rational` which Artichoke doesn't support yet.

artichoke-backend/src/extn/core/time/subsec.rs
```diff
@@ -0,0 +1,562 @@
...
+                #[allow(clippy::cast_precision_loss)]
+                let seconds_base = seconds_base as f64;
+                #[allow(clippy::cast_precision_loss)]
+                let multiplier_nanos = multiplier_nanos as f64;
+
+                let mut secs = subsec / seconds_base;
+                let mut nanos = (subsec % seconds_base) * multiplier_nanos;
+
+                // is_sign_negative() is not enough here, since this logic
+                // should als be skilled for negative zero.
```
lopopolo:
```suggestion
                // should also be skilled for negative zero.
```

artichoke-backend/src/extn/core/time/subsec.rs
```diff
@@ -0,0 +1,562 @@
...
+                })
+            }
+        } else {
+            Ok(Subsec { secs: 0, nanos: 0 })
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
lopopolo, line 197: this looks pretty awesome and we're in a good place to iterate on this later.

artichoke-backend/src/extn/core/time/subsec.rs
```diff
@@ -0,0 +1,562 @@
...
+    secs: i64,
+    nanos: u32,
+}
+
+impl Subsec {
+    /// Returns a tuple of (seconds, nanoseconds). Subseconds are provided in
+    /// various accuracies, and can overflow. e.g. 1001 milliseconds, is 1
+    /// second, and `1_000_000` nanoseconds.
+    #[must_use]
+    pub fn to_tuple(&self) -> (i64, u32) {
```
lopopolo, line 85: per discussion on Discord, is this better represented as `pub fn to_duration(&self) -> core::time::Duration`?

lopopolo: One caveat here is the `Duration` type in `core` is unsigned.

artichoke-backend/src/extn/core/time/subsec.rs
```diff
@@ -0,0 +1,562 @@
...
+        let mut interp = interpreter();
+
+        let err = subsec(&mut interp, (Some(b"Float::NAN"), None)).unwrap_err();
+
+        assert_eq!(err.name(), "FloatDomainError");
+        assert_eq!(err.message(), b"NaN".as_slice());
+    }
+
+    #[test]
+    fn float_infinite_raises() {
```
lopopolo, line 547: we need a test case for `-Float::INFINITY`.

artichoke-backend/src/extn/core/time/subsec.rs
```diff
@@ -0,0 +1,562 @@
...
+        let mut interp = interpreter();
+
+        let err = subsec(&mut interp, (Some(b"Float::INFINITY"), None)).unwrap_err();
+
+        assert_eq!(err.name(), "FloatDomainError");
+        assert_eq!(err.message(), b"Infinity".as_slice());
+    }
+
+    #[test]
+    fn invalid_subsec_unit() {
```
lopopolo, line 562: needs test for non-`Symbol` unit like `"bad_unit"` and `Object.new`.

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -1,35 +1,110 @@
...
+                    Some(first_param)
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
+        } else {
+            None
+        }
```
lopopolo, line 88: wow this is a gnarly expression! I don't see a better way to do this though.

artichoke-backend/src/extn/core/time/trampoline.rs
lopopolo: let's do it in the next PR.

Filed an issue to track:

- https://github.com/artichoke/artichoke/issues/2090

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -180,37 +255,54 @@ pub fn to_array(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
...
 
 // Coarse math
 
-pub fn succ(interp: &mut Artichoke, mut time: Value) -> Result<Value, Error> {
+pub fn succ(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
     interp.warn(b"warning: Time#succ is obsolete; use time + 1")?;
-    let time = unsafe { Time::unbox_from_value(&mut time, interp)? };
-    let next = time.succ();
-    Time::alloc_value(next, interp)
+    plus(interp, time, interp.convert(1))
```
lopopolo, line 305: nice!

lopopolo (approved):
I'll make a few tweaks and file some tickets and get this merged.

I hope it's ok if I mash the merge button on this 😅

artichoke-backend/src/extn/core/time/offset.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/2091

artichoke-backend/src/extn/core/time/offset.rs
lopopolo: Filed https://github.com/artichoke/artichoke/issues/2092.

This is related to https://github.com/artichoke/artichoke/issues/2091.

artichoke-backend/src/extn/core/time/offset.rs
lopopolo: hmm it looks like clippy is ok with this. discarding suggestion

artichoke-backend/src/extn/core/time/offset.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/2093

artichoke-backend/src/extn/core/time/offset.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/2094

artichoke-backend/src/extn/core/time/subsec.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/2095

artichoke-backend/src/extn/core/time/subsec.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/2096

artichoke-backend/src/extn/core/time/subsec.rs
lopopolo: resolving per above. Not much we can do about this given the APIs that currently exist in `std`.

At least we have the `allow(..)`'s in here so we can easily find these spots.

artichoke-backend/src/extn/core/time/subsec.rs
lopopolo: Filed:

- https://github.com/artichoke/artichoke/issues/2095#issuecomment-1221610741

artichoke-backend/src/extn/core/time/trampoline.rs
lopopolo: yea Ruby 3 changed how kwargs are handled. They are no longer converted to `Hash`.

This is a good description of the changes: https://eregon.me/blog/2019/11/10/the-delegation-challenge-of-ruby27.html.

artichoke-backend/src/extn/core/time/offset.rs
lopopolo: looks like this is a difference between ruby 2.6.3 and Ruby 3.1.2.

Added a FIXME in the comment.

artichoke-backend/src/extn/core/time/subsec.rs
lopopolo: added a test and fixed up the exception message.

artichoke-backend/src/extn/core/time/subsec.rs
lopopolo: added a test and added a branch for `subsec.is_sign_negative()`.

lopopolo (approved):
All comments resolved and fixups filed as tickets.

Will merge this when the build is green.

artichoke-backend/src/extn/core/time/trampoline.rs
b-n: Oh good! I was like "really, is this it? There must be a better way". I'm sure there some array plus iterator magic some code golfer could come up with hah.

lopopolo: hopefully once `if let ... else` lands this can get cleaned up a bunch.

PR #1958, opened 2022-07-22
Add support for a `Range` argument to `String#byteslice`

### LINKED ISSUE #1916: `String#byteslice` is missing support for `Range` argument

lopopolo:
Add support for `Range` arguments in `String#byteslice` of the form:

```ruby
s = "abc"
raise unless s.byteslice(1..2) == "bc"
```

As of this commit all `String#byteslice` specs pass except for two which
require the `begin` and `end` components of the `Range` to go through
the "implicitly convert to int" flow. Range parsing is still implemented
in mruby which does not do implicit conversions.

Getting the specs to pass required significant rework of the
`protect::is_range` function to distinguish between the "range out" case
and the "valid" case. For example:

```ruby
"".byteslice(-1..-1)
```

was previously squashing this range into the `None` case, which led to
`byteslice` blowing up further down the routine when attempting to fall
back to the integer argument path.

This change led to re-evaluating all other instances of range argument
handling and corrected the response when the "out of range" case is
encountered. The most significant change was in `Array#[]=` which raises
an exception in this case.

Fixes #1916.

Among the core specs:

Before:
Passed 3944, skipped 1063, not implemented 724, failed 4494 specs.

After:
~~Passed 4014, skipped 1065, not implemented 724, failed 4561 specs.~~
Passed 4051, skipped 1075, not implemented 746, failed 4469 specs.

The relevant mruby code is here:

https://github.com/artichoke/artichoke/blob/f86225f2256b51939d7a2ea0c73ddb36a99f64ab/artichoke-backend/vendor/mruby/src/range.c#L456-L488

### DIFF
--- a/artichoke-backend/src/extn/core/array/args.rs
+++ b/artichoke-backend/src/extn/core/array/args.rs
@@ -83,27 +83,28 @@ pub fn element_assignment(
             write!(&mut message, "negative length ({})", slice_len).map_err(WriteError::from)?;
             Err(IndexError::from(message).into())
         }
-    } else if let Ok(index) = implicitly_convert_to_int(interp, first) {
-        if let Ok(index) = usize::try_from(index) {
-            Ok((index, None, second))
-        } else {
-            let idx = index
-                .checked_neg()
-                .and_then(|index| usize::try_from(index).ok())
-                .and_then(|index| len.checked_sub(index));
-            if let Some(idx) = idx {
-                Ok((idx, None, second))
-            } else {
-                let mut message = String::new();
-                write!(&mut message, "index {} too small for array; minimum: -{}", index, len)
-                    .map_err(WriteError::from)?;
-                Err(IndexError::from(message).into())
-            }
-        }
     } else {
         let rangelen = i64::try_from(len).map_err(|_| Fatal::from("Range length exceeds Integer max"))?;
         match first.is_range(interp, rangelen)? {
-            None => {}
+            None => {
+                let index = implicitly_convert_to_int(interp, first)?;
+                if let Ok(index) = usize::try_from(index) {
+                    Ok((index, None, second))
+                } else {
+                    let idx = index
+                        .checked_neg()
+                        .and_then(|index| usize::try_from(index).ok())
+                        .and_then(|index| len.checked_sub(index));
+                    if let Some(idx) = idx {
+                        Ok((idx, None, second))
+                    } else {
+                        let mut message = String::new();
+                        write!(&mut message, "index {} too small for array; minimum: -{}", index, len)
+                            .map_err(WriteError::from)?;
+                        Err(IndexError::from(message).into())
+                    }
+                }
+            }
             // ```
             // [3.1.2] > a = []
             // => []
@@ -114,46 +115,12 @@ pub fn element_assignment(
             //         from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `load'
             //         from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `<main>'
             // ```
-            Some(protect::Range::Out) => return Err(RangeError::with_message("out of range").into()),
+            Some(protect::Range::Out) => Err(RangeError::with_message("out of range").into()),
             Some(protect::Range::Valid { start, len }) => {
                 let start = usize::try_from(start)
                     .unwrap_or_else(|_| unimplemented!("should throw RangeError (-11..1 out of range)"));
                 let len = usize::try_from(len).unwrap_or_else(|_| unreachable!("Range can't have negative length"));
-                return Ok((start, Some(len), second));
-            }
-        };
-        let start = first.funcall(interp, "begin", &[], None)?;
-        let start = implicitly_convert_to_int(interp, start)?;
-        let end = first.funcall(interp, "last", &[], None)?;
-        let end = implicitly_convert_to_int(interp, end)?;
-        // TODO: This conditional is probably not doing the right thing
-        if start + (end - start) < 0 {
-            let mut message = String::new();
-            write!(&mut message, "{}..{} out of range", start, end).map_err(WriteError::from)?;
-            return Err(RangeError::from(message).into());
-        }
-        match (usize::try_from(start), usize::try_from(end)) {
-            (Ok(start), Ok(end)) => Ok((start, end.checked_sub(start), second)),
-            (Err(_), Ok(end)) => {
-                let pos = start
-                    .checked_neg()
-                    .and_then(|start| usize::try_from(start).ok())
-                    .and_then(|start| len.checked_sub(start));
-                if let Some(start) = pos {
-                    Ok((start, end.checked_sub(start), second))
-                } else {
-                    let mut message = String::new();
-                    write!(&mut message, "index {} too small for array; minimum: -{}", start, len)
-                        .map_err(WriteError::from)?;
-                    Err(IndexError::from(message).into())
-                }
-            }
-            (Ok(start), Err(_)) => Ok((start, None, second)),
-            (Err(_), Err(_)) => {
-                let mut message = String::new();
-                write!(&mut message, "index {} too small for array; minimum: -{}", start, len)
-                    .map_err(WriteError::from)?;
-                Err(IndexError::from(message).into())
+                Ok((start, Some(len), second))
             }
         }
     }
--- a/artichoke-backend/src/extn/core/array/args.rs
+++ b/artichoke-backend/src/extn/core/array/args.rs
@@ -20,23 +20,31 @@ pub fn element_reference(
     if let Some(len) = len {
         let start = implicitly_convert_to_int(interp, elem)?;
         let len = implicitly_convert_to_int(interp, len)?;
-        if let Ok(len) = usize::try_from(len) {
+        return if let Ok(len) = usize::try_from(len) {
             Ok(ElementReference::StartLen(start, len))
         } else {
             Ok(ElementReference::Empty)
+        };
+    }
+    let rangelen = i64::try_from(ary_len).map_err(|_| Fatal::from("Range length exceeds Integer max"))?;
+    match elem.is_range(interp, rangelen)? {
+        None => {
+            let index = implicitly_convert_to_int(interp, elem)?;
+            Ok(ElementReference::Index(index))
         }
-    } else if let Ok(index) = implicitly_convert_to_int(interp, elem) {
-        Ok(ElementReference::Index(index))
-    } else {
-        let rangelen = i64::try_from(ary_len).map_err(|_| Fatal::from("Range length exceeds Integer max"))?;
-        if let Some(protect::Range { start, len }) = elem.is_range(interp, rangelen)? {
+        // ```
+        // [3.1.2] > a = []
+        // => []
+        // [3.1.2] > a[-1..-1]
+        // => nil
+        // ```
+        Some(protect::Range::Out) => Ok(ElementReference::Empty),
+        Some(protect::Range::Valid { start, len }) => {
             if let Ok(len) = usize::try_from(len) {
                 Ok(ElementReference::StartLen(start, len))
             } else {
                 Ok(ElementReference::Empty)
             }
-        } else {
-            Ok(ElementReference::Empty)
         }
     }
 }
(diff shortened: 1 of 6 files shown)


artichoke-backend/src/extn/core/string/mruby.rs
```diff
@@ -213,7 +213,17 @@ unsafe extern "C" fn string_byteslice(mrb: *mut sys::mrb_state, slf: sys::mrb_va
...
     match result {
-        Ok(value) => value.inner(),
+        Ok(value) if value.is_nil() => value.inner(),
+        Ok(value) => {
+            let rclass = sys::mrb_sys_class_of_value(mrb, slf);
+            let value = value.inner();
+            let target_rbasic = value.value.p.cast::<sys::RBasic>();
+
+            // Copy `RClass` from source class to newly allocated `Array`.
+            (*target_rbasic).c = rclass;
```
lopopolo, line 223: to get the specs to pass, `String#byteslice` needs to return slices as a the same class as the receiver.

All `String` and `String` subclasses are backed by `spinoso_string::String` and have type tag of `MRB_TT_STRING`, but they may vary in their `*mut RClass` pointer.

`String::alloc_value` by default hardcodes the `String` class pointer, so we fix it up here.

artichoke-backend/src/extn/core/string/mruby.rs
```diff
@@ -213,7 +213,17 @@ unsafe extern "C" fn string_byteslice(mrb: *mut sys::mrb_state, slf: sys::mrb_va
     let length = length.map(Value::from);
     let result = trampoline::byteslice(&mut guard, value, index, length);
     match result {
-        Ok(value) => value.inner(),
+        Ok(value) if value.is_nil() => value.inner(),
```
lopopolo, line 216: `nil` values are structured differently and if we don't short circuit here, attempting to set the class pointer on a `nil` will segfault.

lopopolo (commented):
@b-n it'd probably be easiest to review this by suppressing whitespace changes in the diff. A lot of the LOC changed are reindenting due to the new match structure as opposed to `if let ... else`.

artichoke-backend/src/extn/core/array/args.rs
```diff
@@ -20,23 +20,31 @@ pub fn element_reference(
...
+        return if let Ok(len) = usize::try_from(len) {
             Ok(ElementReference::StartLen(start, len))
         } else {
             Ok(ElementReference::Empty)
+        };
+    }
+    let rangelen = i64::try_from(ary_len).map_err(|_| Fatal::from("Range length exceeds Integer max"))?;
+    match elem.is_range(interp, rangelen)? {
+        None => {
+            let index = implicitly_convert_to_int(interp, elem)?;
```
lopopolo, line 32: moving the implicit conversion into the "this is not a range" case allows us to properly propagate the underlying `TypeError`s.

artichoke-backend/src/extn/core/string/string_test.rb
```diff
@@ -47,6 +48,18 @@ def string_byteslice
...
+  raise unless s.byteslice(1..-1) == 'bcdefghijk'
+  raise unless s.byteslice(10..-1) == 'k'
+  raise unless s.byteslice(20..-1).nil?
+  raise unless s.byteslice(20..-20).nil?
+  raise unless s.byteslice(2..-20) == ''
+  raise unless s.byteslice(-1..20) == 'k'
+  raise unless s.byteslice(-20..20).nil?
+  raise unless s.byteslice(-5..-1) == 'ghijk'
+  raise unless s.byteslice(-5..1) == ''
+  raise unless s.byteslice(-5..8) == 'ghi'
```
b-n: I think I've figured out all the permutations of ranges that could be provided (and their `irb` values):

`s` =  start
`e` =  end
`S` = -start
`E` = -end
`x` =  start/end same value
`X` = -start/-end same value
`*` =  same element, different signs, see range

```plain
  "abcdefghijk".byteslice(range)
   |---------|   

# Normal ranges
   x---------|   =   0..0   | "a"
   se--------|   =   0..1   | "ab"
   s---------e   =   0..10  | "abcdefghijk"
   |s-------e|   =   1..9   | "bcdefghij"
   |--------se   =   9..10  | "jk"
   |--------s-e  =   9..11  | "jk"
   |---------x   =  10..10  | "k"
   |---------se  =  10..11  | "k"
   |---------|x  =  11..11  | ""
   |---------|se =  11..12  | ""
  
# Inverse ranges
   es--------|   =   1..0   | "" 
   e---------s   =  10..0   | "" 
   |e-------s|   =   9..1   | ""
   |--------es   =  10..9   | ""
   |--------e-s  =  11..9   | ""
   |---------es  =  11..10  | ""
   |---------|es =  12..11  | nil
  
# S-Negative normal ranges
  Se---------|   = -12..0   | nil 
  S-e--------|   = -12..1   | nil
   *---------|   = -11..0   | "a"
   Se--------|   = -11..1   | "ab"
   S---------e   = -11..10  | "abcdefghijk"
   S----------e  = -11..11  | "abcdefghijk"
   |S-------e|   = -10..9   | "bcdefghij"
   |--------Se   =  -2..10  | "jk"
   |---------*   =  -1..10  | "k"
  
# E-Negative normal ranges
   *---------|   =   0..-11 | "a"
   sE--------|   =   0..-10 | "ab"
   s---------E   =   0..-1  | "abcdefghijk"
   |s-------E|   =   1..-2  | "bcdefghij"
   |--------sE   =   9..-1  | "jk"
   |---------*   =  10..-1  | "k"

# S-Negative and E-Negative inverse ranges
  Es---------|   =   0..-12 | ""
  E-s--------|   =   1..-12 | ""
   Es--------|   =   1..-11 | ""
   |--------Es   =  10..-2  | ""
   |--------E-s  =  11..-2  | ""
   |---------Es  =  11..-1  | ""

# Dbl-Negative ranges
 SE|---------|   = -13..-12 | nil
  X|---------|   = -12..-12 | nil
  SE---------|   = -12..-11 | nil
  S-E--------|   = -12..-10 | nil
   X---------|   = -11..-11 | "a"
   SE--------|   = -11..-10 | "ab"
   S---------E   = -11..-1  | "abcdefghijk"
   |--------SE   =  -2..-1  | "jk"
   |---------X   =  -1..-1  | "k"

# Dbl-Negative inverse ranges
 ES|---------|   = -12..-13 | nil
  ES---------|   = -11..-12 | ""
  E-S--------|   = -10..-12 | ""
   ES--------|   = -10..-11 | ""
   E---------S   =  -1..-11 | ""
   |--------ES   =  -1..-2  | ""
```

I think there might be some cases that can be added if you're really keen :smile:

b-n: This is the same as the above table, just condensed into the ruby tests (passes on MRI 3.1.2)
```suggestion
  raise unless s.byteslice(0..0) == "a"
  raise unless s.byteslice(0..1) == "ab"
  raise unless s.byteslice(0..10) == "abcdefghijk"
  raise unless s.byteslice(1..9) == "bcdefghij"
  raise unless s.byteslice(9..10) == "jk"
  raise unless s.byteslice(9..11) == "jk"
  raise unless s.byteslice(10..10) == "k"
  raise unless s.byteslice(10..11) == "k"
  raise unless s.byteslice(11..11) == ""
  raise unless s.byteslice(11..12) == ""
  raise unless s.byteslice(1..0) == "" 
  raise unless s.byteslice(10..0) == "" 
  raise unless s.byteslice(9..1) == ""
  raise unless s.byteslice(10..9) == ""
  raise unless s.byteslice(11..9) == ""
  raise unless s.byteslice(11..10) == ""
  raise unless s.byteslice(12..11).nil?
  raise unless s.byteslice(-12..0).nil? 
  raise unless s.byteslice(-12..1).nil?
  raise unless s.byteslice(-11..0) == "a"
  raise unless s.byteslice(-11..1) == "ab"
  raise unless s.byteslice(-11..10) == "abcdefghijk"
  raise unless s.byteslice(-11..11) == "abcdefghijk"
  raise unless s.byteslice(-10..9) == "bcdefghij"
  raise unless s.byteslice(-2..10) == "jk"
  raise unless s.byteslice(-1..10) == "k"
  raise unless s.byteslice(0..-11) == "a"
  raise unless s.byteslice(0..-10) == "ab"
  raise unless s.byteslice(0..-1) == "abcdefghijk"
  raise unless s.byteslice(1..-2) == "bcdefghij"
  raise unless s.byteslice(9..-1) == "jk"
  raise unless s.byteslice(10..-1) == "k"
  raise unless s.byteslice(0..-12) == ""
  raise unless s.byteslice(1..-12) == ""
  raise unless s.byteslice(1..-11) == ""
  raise unless s.byteslice(10..-2) == ""
  raise unless s.byteslice(11..-2) == ""
  raise unless s.byteslice(11..-1) == ""
  raise unless s.byteslice(-13..-12).nil?
  raise unless s.byteslice(-12..-12).nil?
  raise unless s.byteslice(-12..-11).nil?
  raise unless s.byteslice(-12..-10).nil?
  raise unless s.byteslice(-11..-11) == "a"
  raise unless s.byteslice(-11..-10) == "ab"
  raise unless s.byteslice(-11..-1) == "abcdefghijk"
  raise unless s.byteslice(-2..-1) == "jk"
  raise unless s.byteslice(-1..-1) == "k"
  raise unless s.byteslice(-12..-13).nil?
  raise unless s.byteslice(-11..-12) == ""
  raise unless s.byteslice(-10..-12) == ""
  raise unless s.byteslice(-10..-11) == ""
  raise unless s.byteslice(-1..-11) == ""
  raise unless s.byteslice(-1..-2) == ""
```

artichoke-backend/src/sys/protect.rs
```diff
@@ -148,10 +148,10 @@ pub unsafe fn is_range(
     }
 }
 
-#[derive(Default, Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
-pub struct Range {
-    pub start: sys::mrb_int,
-    pub len: sys::mrb_int,
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum Range {
+    Valid { start: sys::mrb_int, len: sys::mrb_int },
```
b-n, line 153: I'm assuming that the start value is enforced to be an `sys::mrb_int`?

```
irb(main):001:0> 1.5..3
=> 1.5..3
```

^-- this is MRI 2.6.3, but gives  the same on MRI 3.1.2. I never really thought of ranges having fractional parts, but apparently that's the MRI world.

artichoke-backend/src/sys/protect.rs
```diff
@@ -164,20 +164,28 @@ struct IsRange {
...
-        if check_range == sys::mrb_range_beg_len::MRB_RANGE_OK {
-            let start = start.assume_init();
-            let range_len = range_len.assume_init();
-            let out = Range { start, len: range_len };
-            let out = Box::new(out);
-            let out = Box::into_raw(out);
-            sys::mrb_sys_cptr_value(mrb, out.cast::<c_void>())
-        } else {
-            sys::mrb_sys_nil_value()
+        match check_range {
```
b-n, line 174: Am I interpreting this correctly, does mruby change a Range into a start + len for us? That's pretty nice.

e.g.
```
a = "hello"
a.byteslice(-3..-3)
```

In the above, `sys::mrb_range_beg_len` would return `start: 2, len: 1`? Very nice if it does

artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -410,6 +418,74 @@ pub fn byteslice(
...
+                Some(index)
+            } else {
+                index
+                    .checked_neg()
+                    .and_then(|index| usize::try_from(index).ok())
+                    .and_then(|index| s.len().checked_sub(index))
+            };
+            let index = match index {
+                None => return Ok(Value::nil()),
+                Some(index) if index > s.len() => return Ok(Value::nil()),
```
b-n, line 440: I think my understanding of how mruby unwraps into `Range::Valid` isn't clear. Should mruby take care of these checks and provide a `Range::Out` instead?

b-n (approved):
Pretty nice refactorings. I've got a couple of questions around the `sys::mrb_range_beg_len` and a couple of tests for byteslice that could be added.

I'm wondering whether something like this would ever make sense:

```rust
impl for Range {
    pub fn array_accessor(interp: &mut Artichoke, cap: Option<Value>) -> (usize, usize) {
        match Range {
            Range::Valid { start, len } => {
                  //do all the checking of start/len against the cap
                  (start, len)
            },
            _ => Err()
        }
    }
}
```

:point_up: Something that does all the conversion to usize, and just returns the elements to pluck out of a list. I don't think there's anything particular special about a range, except it gives you many ways to access elements in an array in this case.

I'm approving since I don't see anything in particular that could/should break.

artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -410,6 +418,74 @@ pub fn byteslice(
     length: Option<Value>,
 ) -> Result<Value, Error> {
     let s = unsafe { super::String::unbox_from_value(&mut value, interp)? };
+    let maybe_range = if length.is_none() {
+        index.is_range(interp, s.char_len() as i64)?
```
lopopolo:
```suggestion
        index.is_range(interp, s.len() as i64)?
```

since we're doing byte slices.

artichoke-backend/src/extn/core/string/trampoline.rs
lopopolo: I think you're probably right let me check this case in the repl ...

```
artichoke 0.1.0-pre.0 (2022-07-23 revision 5844) [x86_64-apple-darwin]
[rustc 1.62.1 (e092d0b6b 2022-07-16) on x86_64-apple-darwin]
>>> s = "Abc"
=> "Abc"
>>> s.byteslice(4..6)
[artichoke-backend/src/extn/core/string/trampoline.rs:440] Value::nil() = Value(
    nil,
)
=> nil
>>> s.byteslice(10..20)
[artichoke-backend/src/extn/core/string/trampoline.rs:440] Value::nil() = Value(
    nil,
)
=> nil
```

Somehow this case is getting hit.

artichoke-backend/src/sys/protect.rs
lopopolo: When `Array`s and `String`s accept range arguments, they require them to be ints.

The mruby `mrb_range_beg_len` requires `start` and `end` to be ints:

https://github.com/artichoke/artichoke/blob/7bf3f033e020dd60f4b5fd5974dcb19af2617f6d/artichoke-backend/vendor/mruby/src/range.c#L466-L467

artichoke-backend/src/sys/protect.rs
lopopolo: yes that's correct. The routine looks like this:

https://github.com/artichoke/artichoke/blob/7bf3f033e020dd60f4b5fd5974dcb19af2617f6d/artichoke-backend/vendor/mruby/src/range.c#L456-L488

artichoke-backend/src/extn/core/string/trampoline.rs
lopopolo: After taking a peek at the code, the given length argument is only used for wrapping negative start and end of the `Range`.

As long as these come out postive, regardless of whether they overlap with `0..len`, `MRB_RANGE_OK` is returned.

https://github.com/artichoke/artichoke/blob/7bf3f033e020dd60f4b5fd5974dcb19af2617f6d/artichoke-backend/vendor/mruby/src/range.c#L456-L488

lopopolo:
> 

Great idea! Filed https://github.com/artichoke/artichoke/issues/1971.

lopopolo:
Added some tests with multibyte UTF-8 character sequence to cover the `char_len` -> `bytesize` fix.

PR #1961, opened 2022-07-23
Fix argument unboxing precedence for `MatchData#[]`

lopopolo:
Combining the new `Range` parsing in #1958 with the recent discovery that implicit conversions to integer need to be in the tail of argument unboxing (so `TypeError`s can be propagated), we get this new monstrosity in `matchdata::trampoline::element_reference`.

Added lots more docs with REPL output.

This new argument umarshalling code matches better to Sorbet's type signatures (note that sorbet does not support implicit conversions in its type system):

https://github.com/sorbet/sorbet/blob/37cb4686948095ddb795e2eeed0e6f33f3276e1e/rbi/core/match_data.rbi#L97-L122

### DIFF
--- a/artichoke-backend/src/extn/core/matchdata/trampoline.rs
+++ b/artichoke-backend/src/extn/core/matchdata/trampoline.rs
@@ -1,7 +1,8 @@
-use crate::convert::{implicitly_convert_to_int, implicitly_convert_to_string};
+use crate::convert::implicitly_convert_to_int;
 use crate::extn::core::array::Array;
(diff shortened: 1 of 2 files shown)


lopopolo:
It looks like checks won't run until #1958 is merged and I retarget this branch to `trunk`.

b-n (approved):
I imagine a day where the capture groups would be exposed. This PR makes me think more about a utility function of for range values, and getting a start/end out of them based on a given size, but that's for #1971

Issue #1912, opened 2022-06-21
Enable all `String` ruby/specs


lopopolo:
Modify `enforced-specs.toml` to `include = "all"` for `String` in Ruby Core:

https://github.com/artichoke/artichoke/blob/e545f655dac9a08a67b6e7183858b8dcef72aaf8/spec-runner/enforced-specs.toml#L94-L96

After running the specs with:

```console
$ cd spec-runner
$ cargo run --bin spec-runner -q -- enforced-specs.toml
```

See which ones fail and please file issues.

lopopolo:
See:

- #1915 
- #1916 
- #1917 
- #1926

PR #1963, opened 2022-07-23
Fix spec compliance for `String#<<`

lopopolo:
Fix issues with `String#<<`:

- Raise `FrozenError` if the receiver is frozen.
- Add support for pushing codepoints to a string.
- Make pushing codepoints to a string encoding-aware.
- Add support for in-place encoding mutations when pushing non-ASCII bytes to an ASCII-encoded string.
- Make pushing strings to a string encoding-aware.
- Add support for in-place encoding mutations when pushing to an empty string.

Progress on https://github.com/artichoke/artichoke/issues/1912.

To make these changes, some additional feature work was required:

- Add a `spinoso_string::String::set_encoding` API (@b-n: it's back!).
- Fix a bug in `impl BoxUnboxVmValue for String` where encoding flags were not being zeroed before repacking.
- Add a new implicit conversion routine to string where the underlying `spinoso_string::String` is returned instead of the underlying bytes.

## ruby/spec

Before: Passed 1917, skipped 250, not implemented 15, failed 0 specs.
After: Passed 1932, skipped 260, not implemented 15, failed 2 specs.

I declined to add any additional tests since all of this new feature work is exercised by ruby/spec. After these changes, these are the only failures (due to missing support for `String#encode` and `Encoding::UTF16_LE`):

```
Passed 1932, skipped 260, not implemented 15, failed 2 specs.

String#<< when self is in an ASCII-incompatible encoding incompatible with the argument's encoding raises Encoding::CompatibilityError if neither are empty
SpecExpectationNotMetError: Expected Encoding::CompatibilityError
but no exception was raised ("xy" was returned)

/artichoke/virtual_root/src/lib/core/string/shared/concat.rb:115:in protect
(eval):165:in all?
(eval):590:in each
(eval):590:in each
/artichoke/virtual_root/src/lib/core/string/append_spec.rb:8
(eval):590:in each
/artichoke/virtual_root/src/lib/spec_runner.rb:74:in run_specs
(eval):1

String#<< when the argument is in an ASCII-incompatible encoding incompatible with self's encoding raises Encoding::CompatibilityError if neither are empty
SpecExpectationNotMetError: Expected Encoding::CompatibilityError
but no exception was raised ("xy" was returned)

/artichoke/virtual_root/src/lib/core/string/shared/concat.rb:133:in protect
(eval):165:in all?
(eval):590:in each
(eval):590:in each
/artichoke/virtual_root/src/lib/core/string/append_spec.rb:8
(eval):590:in each
/artichoke/virtual_root/src/lib/spec_runner.rb:74:in run_specs
(eval):1

Passed 1932, skipped 260, not implemented 15, failed 2 specs.
```

### DIFF
--- a/artichoke-backend/src/convert/implicit.rs
+++ b/artichoke-backend/src/convert/implicit.rs
@@ -1,7 +1,10 @@
+use core::mem;
+
 use artichoke_core::debug::Debug as _;
 use artichoke_core::value::Value as _;
 use spinoso_exception::TypeError;
 
+use crate::convert::{BoxUnboxVmValue, UnboxedValueGuard};
 use crate::error::Error;
 use crate::types::Ruby;
 use crate::value::Value;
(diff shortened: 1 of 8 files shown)


spinoso-string/src/enc/mod.rs
```diff
@@ -147,6 +148,16 @@ impl EncodedString {
...
         }
     }
+
+    #[inline]
+    pub fn set_encoding(&mut self, encoding: Encoding) {
+        if self.encoding() == encoding {
+            return;
+        };
+        let s = mem::take(self);
+        let buf = s.into_vec();
```
lopopolo: this will need a rebase once #1976 lands.

lopopolo: done

artichoke-backend/src/convert/implicit.rs
```diff
@@ -483,3 +486,213 @@ pub unsafe fn implicitly_convert_to_nilable_string<'a>(
...
+/// # Safety
+///
+/// Callers must ensure that `value` does not outlive the given interpreter.
+///
+/// If a garbage collection can possibly run between calling this function and
+/// using the returned slice, callers should convert the slice to an owned byte
+/// vec.
+///
+/// [`Symbol`]: crate::extn::core::symbol::Symbol
+pub unsafe fn implicitly_convert_to_spinoso_string<'a>(
```
b-n, line 563: What are your thoughts on making this a `implicitly_convert_to_nilable_spinoso_string` and making it return `Result<Option<UnboxedValueGuard<'a, spinoso_string::String>>, Error>` ?

I'd imagine this might enable to checking of things like this easier:

```
irb(main):001:0> Time.at(1000, in: "")
        5: from /home/ben/.rbenv/versions/2.6.3/bin/irb:23:in `<main>'
        4: from /home/ben/.rbenv/versions/2.6.3/bin/irb:23:in `load'
        3: from /home/ben/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        2: from (irb):1
        1: from (irb):1:in `at'
ArgumentError ("+HH:MM" or "-HH:MM" expected for utc_offset)
irb(main):002:0> Time.at(1000, in: nil)
        6: from /home/ben/.rbenv/versions/2.6.3/bin/irb:23:in `<main>'
        5: from /home/ben/.rbenv/versions/2.6.3/bin/irb:23:in `load'
        4: from /home/ben/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        3: from (irb):2
        2: from (irb):2:in `rescue in irb_binding'
        1: from (irb):2:in `at'
TypeError (can't convert nil into an exact number)
irb(main):003:0> 
```

:point_up: There might be better ways to implement this kind of switching, since in theory `Time.at(1000, in: 1000)` is also valid (e.g. the switch logic needs to handle 3 types in this above case :thinking:)

I can imagine the moot of `nil` is never a `String`, so your current implementation feels more correct, but just thinking out loud.

b-n, line 563: Additionally, another alternative `implicitly_convert_to_ruby_string` ?

artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -42,31 +44,324 @@ pub fn add(interp: &mut Artichoke, mut value: Value, mut other: Value) -> Result
...
+                // The string is reboxed before any intervening operations on the
+                // interpreter.
+                // The string is reboxed without any intervening mruby allocations.
+                unsafe {
+                    let string_mut = s.as_inner_mut();
+                    // XXX: This call doesn't do a check to see if we'll exceed the max allocation
+                    //    size and may panic or abort.
+                    string_mut.push_byte(byte);
+                    if !byte.is_ascii() {
+                        string_mut.set_encoding(Encoding::Binary);
```
b-n, line 95: Does this really need the conversion?

```
irb(main):001:0> a = "hello"
=> "hello"
irb(main):002:0> i = 64
=> 64
irb(main):003:0> a << i
=> "hello@"
irb(main):004:0> a.encoding
=> #<Encoding:UTF-8>
irb(main):005:0> a.force_encoding(Encoding::ASCII)
=> "hello@"
irb(main):006:0> a.encoding
=> #<Encoding:US-ASCII>
irb(main):007:0> a << i
=> "hello@@"
irb(main):008:0> a.encoding
=> #<Encoding:US-ASCII>
```

It looks like it survives as ASCII afterwards

artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -42,31 +44,324 @@ pub fn add(interp: &mut Artichoke, mut value: Value, mut other: Value) -> Result
...
+                //
+                // The string is reboxed before any intervening operations on the
+                // interpreter.
+                // The string is reboxed without any intervening mruby allocations.
+                unsafe {
+                    let string_mut = s.as_inner_mut();
+                    // XXX: This call doesn't do a check to see if we'll exceed the max allocation
+                    //    size and may panic or abort.
+                    string_mut
+                        .try_push_codepoint(int)
```
b-n, line 76: Rust is crazy cool with it's conversions here. I was just playing with `try_push_codepoint` expecting it to not work, but of course it does work. I will play with this one day, since `char.try_from(192)` giving unicode `\u00C0` is very cool and partially black magic to me right now.

artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -42,31 +44,324 @@ pub fn add(interp: &mut Artichoke, mut value: Value, mut other: Value) -> Result
...
+        let message = "can't modify frozen String: "
+            .chars()
+            .chain(s.inspect())
+            .collect::<super::String>();
+        return Err(FrozenError::from(message.into_vec()).into());
+    }
+
     let mut s = unsafe { super::String::unbox_from_value(&mut value, interp)? };
+    if let Ok(int) = other.try_convert_into::<i64>(interp) {
+        return match s.encoding() {
```
b-n, line 64: How do you feel about `push_int` in `spinoso-string` and letting the encoded types to deal with this?

spinoso-string/src/enc/mod.rs
```diff
@@ -147,6 +148,16 @@ impl EncodedString {
...
     }
+
+    #[inline]
+    pub fn set_encoding(&mut self, encoding: Encoding) {
+        if self.encoding() == encoding {
+            return;
+        };
+        let s = mem::take(self);
+        let buf = s.into_buf();
+        *self = Self::new(buf, encoding);
```
b-n, line 159: Nice!

Just checking, `mem::take` is letting this function own self, and then we're just setting that ownership back to self with the `*self = ` assignment right? That's very handy in this niche cache.

artichoke-backend/src/convert/implicit.rs
lopopolo: We have both `implicitly_convert_to_string` and `implicitly_convert_to_nilable_string`. If you find a need for implicitly converting to a nilable spinoso string, it is just a few lines to add that function. Since I don't need it here, I'd prefer to hold off on adding it.

https://github.com/artichoke/artichoke/blob/9a979d9f57ad2afe4b7d312e91a07e15aee93d28/artichoke-backend/src/convert/implicit.rs#L475-L485

artichoke-backend/src/extn/core/string/trampoline.rs
lopopolo: 🦀 🚀

artichoke-backend/src/extn/core/string/trampoline.rs
lopopolo: 😉 try this again with `s << 255`. That's what the `!byte.is_ascii()` checks for.

spinoso-string/src/enc/mod.rs
lopopolo: `mem::take` moves the `EncodedString` struct out of `&mut self` and leaves `EncodedString::default()` in its place.

Then we grab its buf, construct a new `EncodedString` with the buf and given encoding.

Then we move that new `EncodedString` back into the `&mut self`.

throughout this whole dance, `self` is valid and initialized.

artichoke-backend/src/convert/implicit.rs
lopopolo: /shrug I'm ambivalent between both of these options.

artichoke-backend/src/extn/core/string/trampoline.rs
lopopolo: that's a good idea! let me file a ticket.

Edit: done!

- https://github.com/artichoke/artichoke/issues/1981

artichoke-backend/src/convert/implicit.rs
lopopolo: > the switch logic needs to handle 3 types in this above case

One thing that I think is a constant across all of the type conversions is that only one argument shape gets the implicit conversion treatment. So a method will either implicitly convert to string OR int, but not both.

That implicit conversion needs to go last in the chain so we can propagate the `TypeError`s.

b-n: > Since I don't need it here, I'd prefer to hold off on adding it.

Yes good call!

b-n: Likewise! Resolving :-D

artichoke-backend/src/extn/core/string/trampoline.rs
b-n: hah oof :grimacing:

b-n (approved):
LGTM!

PR #1969, opened 2022-07-24
Add a NUL byte to the end of the allocation from `mrb_sym_str`

lopopolo:
mruby assumes that the byte content associated with symbols are stored in memory as a null terminated `char*`s in the symbol table. Artichoke stores symbols as NUL terminated strings in the symbol table.

When creating a Ruby value from the symbol table with `mrb_sym_str`, mruby creates "static" `RString*`s from interned bytes that point at this NUL terminated memory. This behavior differs from Artichoke that does not have a static string/`Cow` optimization.

Sometimes mruby grabs the underlying `RSTRING_PTR` from these symbol table-backed `RString*`s and relies on the pointer being NUL terminated.

## Flow

In mruby, `Class#name` is implemented with a `__classname__` ivar on `Class` objects that points to an interned string via a `Symbol`.

In native code, mruby resolves this symbol to a byte buffer and wants to get access to the underlying `RSTRING_PTR`.

When mruby wants to pass around a byte buffer, it uses an `mrb_value` of type `MRB_TT_STRING` that is backed by an `RString*`.

In some cases, when getting `RSTRING_PTR`s from strings backed by the symbol table, mruby assumes the pointer is NUL terminated.

This assumption is made because mruby assumes that symbols are stored in memory as a null terminated `char*`s. When converting a symbol into string `mrb_value`, mruby creates "static" `RString`s from interned bytes that point at this NUL terminated memory.

### C Call Graph

`mrb_class_name` directly grabs the `RSTRING_PTR` from a String value:

https://github.com/artichoke/artichoke/blob/7bf3f033e020dd60f4b5fd5974dcb19af2617f6d/artichoke-backend/vendor/mruby/src/class.c#L2148

which searches the class path:

https://github.com/artichoke/artichoke/blob/7bf3f033e020dd60f4b5fd5974dcb19af2617f6d/artichoke-backend/vendor/mruby/src/class.c#L396

which is returns a class name from the symbol table:

https://github.com/artichoke/artichoke/blob/7bf3f033e020dd60f4b5fd5974dcb19af2617f6d/artichoke-backend/vendor/mruby/src/class.c#L2125

## Detection

This memory corruption was observed when printing out class names in ARGF specs. I observed this error while running the spec suite for `ARGF` where I got unexpected errors like:

```
An exception occurred during: loading core/argf/argf_spec.rb
ARGF is extended by the Enumerable module
ArgumentError: regex crate utf8 backend for Regexp only supports UTF-8 haystacks
```

This comes from the spec harness attempting to `Regexp` match on exception class names:

https://github.com/artichoke/artichoke/blob/7bf3f033e020dd60f4b5fd5974dcb19af2617f6d/spec-runner/src/spec_runner.rb#L133

After adding this diff:

```diff
diff --git i/spec-runner/src/spec_runner.rb w/spec-runner/src/spec_runner.rb
index 85b2cc8f327..51fa352ee86 100755
--- i/spec-runner/src/spec_runner.rb
+++ w/spec-runner/src/spec_runner.rb
@@ -130,6 +130,7 @@ module Artichoke
             skipped = true if state.message =~ /'untrusted\?'/
             skipped = true if state.message =~ /undefined method 'Rational'/
           when NameError
+            puts state.message.inspect
             skipped = true if state.message =~ /uninitialized constant Bignum/
           when SpecExpectationNotMetError
             skipped = true if state.it =~ /encoding/
```

Output showed this exception message:

```
"NameError: uninitialized constant #<Class:#<MSpecEnv\xFB\ae__:0x7fd6600c08a0>>::ARGF"
```

oh noes! memory unsafety 😧 

## Fix

Artichoke does not have a "static"/`Cow` string optimization, so `mrb_sym_str` allocates:

https://github.com/artichoke/artichoke/blob/7bf3f033e020dd60f4b5fd5974dcb19af2617f6d/artichoke-backend/src/extn/core/symbol/ffi.rs#L228-L232

In practice the call to `<[u8]>::clone` does not allocate excess capacity so the symbol `MSpecEnv` allocates a buffer with `capacity = len = 8`.

This means the memory beyond the `v` byte can have any value.

To ensure the buffer is "NUL terminated": Add a 0 byte to the end of the `Vec` to ensure that the byte at index `len + 1` is NUL.

https://github.com/artichoke/artichoke/blob/80f5397d0bf9025bb87b416e092b3a2a06e749d7/artichoke-backend/src/extn/core/symbol/ffi.rs#L237-L257

### DIFF
--- a/artichoke-backend/src/extn/core/symbol/ffi.rs
+++ b/artichoke-backend/src/extn/core/symbol/ffi.rs
@@ -229,7 +229,29 @@ unsafe extern "C" fn mrb_sym_str(mrb: *mut sys::mrb_state, sym: sys::mrb_sym) ->
     unwrap_interpreter!(mrb, to => guard);
 
     let value = if let Ok(Some(bytes)) = guard.lookup_symbol(sym) {
-        let bytes = bytes.to_vec();
+        let mut bytes = bytes.to_vec();
+
+        // SAFETY: ensure the byte buffer is NUL-terminated.
+        //
+        // Add a NUL byte to the end of the allocation for the byte buffer in
+        // the new `RString`.
+        //
+        // mruby assumes that symbols are stored in memory as a null terminated
+        // `char*`s and creates "static" `RString`s from interned bytes that
+        // point at this NUL terminated memory.
+        //
+        // Sometimes mruby grabs the `RString` pointer directly from value
+        // returned by this API call and assumes it is NUL terminated.
+        //
+        // Add a 0 byte to the end of the `Vec` to force an allocation and then
+        // set the length to the length of the byte content.
+        //
+        // See https://github.com/artichoke/artichoke/pull/1969.
+        let len = bytes.len();
+        bytes.reserve_exact(1);
+        bytes.push(0);
+        bytes.set_len(len);
+
         guard.try_convert_mut(bytes)
     } else {
         guard.try_convert_mut("")
--- a/artichoke-backend/src/extn/core/symbol/ffi.rs
+++ b/artichoke-backend/src/extn/core/symbol/ffi.rs
@@ -228,34 +228,35 @@ unsafe extern "C" fn mrb_sym_dump(mrb: *mut sys::mrb_state, sym: sys::mrb_sym) -
 unsafe extern "C" fn mrb_sym_str(mrb: *mut sys::mrb_state, sym: sys::mrb_sym) -> sys::mrb_value {
     unwrap_interpreter!(mrb, to => guard);
 
-    let value = if let Ok(Some(bytes)) = guard.lookup_symbol(sym) {
-        let mut bytes = bytes.to_vec();
-
-        // SAFETY: ensure the byte buffer is NUL-terminated.
-        //
-        // Add a NUL byte to the end of the allocation for the byte buffer in
-        // the new `RString`.
-        //
-        // mruby assumes that symbols are stored in memory as a null terminated
-        // `char*`s and creates "static" `RString`s from interned bytes that
-        // point at this NUL terminated memory.
-        //
-        // Sometimes mruby grabs the `RString` pointer directly from value
-        // returned by this API call and assumes it is NUL terminated.
-        //
-        // Add a 0 byte to the end of the `Vec` to force an allocation and then
-        // set the length to the length of the byte content.
-        //
-        // See https://github.com/artichoke/artichoke/pull/1969.
-        let len = bytes.len();
-        bytes.reserve_exact(1);
-        bytes.push(0);
-        bytes.set_len(len);
-
-        guard.try_convert_mut(bytes)
+    let mut bytes = if let Ok(Some(bytes)) = guard.lookup_symbol(sym) {
+        bytes.to_vec()
     } else {
-        guard.try_convert_mut("")
+        vec![]
     };
+
+    // SAFETY: ensure the byte buffer is NUL-terminated.
+    //
+    // Add a NUL byte to the end of the allocation for the byte buffer in
+    // the new `RString*`.
+    //
+    // mruby assumes that symbols are stored in memory as a null terminated
+    // `char*`s and creates "static" `RString`s from interned bytes that
+    // point at this NUL terminated memory.
+    //
+    // Sometimes mruby grabs the `RString` pointer directly from value
+    // returned by this API call and assumes it is NUL terminated.
+    //
+    // Add a 0 byte to the end of the `Vec` to ensure that the byte at index
+    // `len + 1` is NUL. Then set the length to the length of the `Vec` to
+    // hide the NUL byte.
+    //
+    // See https://github.com/artichoke/artichoke/pull/1969.
+    let len = bytes.len();
+    bytes.reserve_exact(1);
+    bytes.push(0);
+    bytes.set_len(len);
+
+    let value = guard.try_convert_mut(bytes);
     value.unwrap_or_default().inner()
 }
 


lopopolo:
While we're peeking at this code, I filed this upstream. This upstream bug also impacts Artichoke because strings returned by `mrb_sym_str` are not static:

- https://github.com/mruby/mruby/pull/5754

lopopolo:
Got an add hoc code review from @alex as of acb5be8065e60de438fef34637811d83d26eebe7:

- @lopopolo: is this bug report for an incidence of memory unsafety clear enough? https://github.com/artichoke/artichoke/pull/1969. also 😬 not proud of the fix
- @alex: Using `set_len` to "hide" the nul value strikes me as extremely fragile. Dunno if its actually relying on undefined behavior, but seems bad.
  - @lopopolo: I think it just relies on the vec not being modified, which is true in practice for legacy C that uses this API and relies on this property.
- @alex:
  > Add a 0 byte to the end of the Vec to force an allocation
  
  I don't think anything about this code actually guarantees that a new allocation will occur.
- @alex: This seems that the `Vec` returned by `bytes.to_vec()` allocates the exact size of `bytes`, but I don't think the API guarantees that.
  - @lopopolo: I think that's fine? if the allocation is bigger that's ok, I just need len + 1 to point to a `\0`.
  - @alex: Then the comment should be updated -- the comment suggests its relying on a new allocation happening.
  - @lopopolo: done. thanks for the feedback [NOTE: fixed in 80f5397d0bf9025bb87b416e092b3a2a06e749d7]
- @lopopolo: in practice uses in the legacy C code that rely on this property treat the C string as read only
- @alex: Also this bug probably applies to the `else` clause. static `""` isn't guaranteed to be NUL terminated either. [NOTE: fixed in 80f5397d0bf9025bb87b416e092b3a2a06e749d7]
- @lopopolo: this safety condition is the same as the safety condition in `Vec::as_mut_ptr`
  > I think it just relies on the vec not being modified, which is true in practice for legacy C that uses this API and relies on this property.

Thanks for the code review @alex!

lopopolo:
Despite the fragility of this fix, this is the best I can do without invasive changes to `spinoso-string` to add specialization for quasi-`'static` (~interpreter) lifetime strings. Merging for now.

One potential way to fix this is to make `spinoso-string` backed by [`bytes::Bytes`](https://docs.rs/bytes/latest/bytes/struct.Bytes.html) instead of `Vec<u8>`.

lopopolo:
Tweetstorm about this bug: https://twitter.com/artichokeruby/status/1551290796216705024

lopopolo:
This bugfix removed 6 of these exceptions from the `all-core-specs` run. From `spec-state` repo:

```console
$ rg 'ArgumentError: regex crate utf8 backend for Regexp only supports UTF-8 haystacks' reports/exceptions/2022/07/spec-exceptions-2022-07-22-rev-68cc73b3fbfed4517a535b77d4fe5c0474b19523.json | wc -l
      11
$ rg 'ArgumentError: regex crate utf8 backend for Regexp only supports UTF-8 haystacks' reports/exceptions/2022/07/spec-exceptions-2022-07-24-rev-9f446ac0bcd7f58f629e8c464520c08e08988f3c.json | wc -l
       5
```

Of the remaining ones, the `StringScanner` one appears to be a legitimate binary haystack. The ones involving `Method` and `UnboundMethod` are suspect:

```console
$ rg 'ArgumentError: regex crate utf8 backend for Regexp only supports UTF-8 haystacks' reports/exceptions/latest.json
3349:    "An exception occurred during: before :each\nMethod#inspect returns a String for methods defined with attr_accessor ERROR\nArgumentError: regex crate utf8 backend for Regexp only supports UTF-8 haystacks\n(eval):1082:in match\n(eval):1082:in sub\n/artichoke/virtual_root/src/lib/core/method/shared/to_s.rb:7:in protect\n(eval):165:in all?\n(eval):590:in each\n/artichoke/virtual_root/src/lib/core/method/inspect_spec.rb:6\n(eval):590:in each\n/artichoke/virtual_root/src/lib/spec_runner.rb:250:in run_specs\n(eval):1",
3366:    "An exception occurred during: before :each\nMethod#unbind returns a String containing the Module the method is referenced from ERROR\nArgumentError: regex crate utf8 backend for Regexp only supports UTF-8 haystacks\n(eval):1082:in match\n(eval):1082:in sub\n/artichoke/virtual_root/src/lib/core/method/unbind_spec.rb:10:in protect\n(eval):165:in all?\n(eval):590:in each\n/artichoke/virtual_root/src/lib/core/method/unbind_spec.rb:37\n(eval):590:in each\n/artichoke/virtual_root/src/lib/spec_runner.rb:250:in run_specs\n(eval):1",
7184:    "UnboundMethod#inspect the String shows the method name, Module defined in and Module extracted from ERROR\nArgumentError: regex crate utf8 backend for Regexp only supports UTF-8 haystacks\n(eval):239:in match\n(eval):239:in =~\n/artichoke/virtual_root/src/lib/core/unboundmethod/shared/to_s.rb:23:in protect\n(eval):165:in all?\n(eval):590:in each\n/artichoke/virtual_root/src/lib/core/unboundmethod/inspect_spec.rb:7\n(eval):590:in each\n/artichoke/virtual_root/src/lib/spec_runner.rb:250:in run_specs\n(eval):1",
7189:    "UnboundMethod#to_s the String reflects that this is an UnboundMethod object ERROR\nArgumentError: regex crate utf8 backend for Regexp only supports UTF-8 haystacks\n(eval):239:in match\n(eval):239:in =~\n/artichoke/virtual_root/src/lib/core/unboundmethod/shared/to_s.rb:17:in protect\n(eval):165:in all?\n(eval):590:in each\n/artichoke/virtual_root/src/lib/core/unboundmethod/to_s_spec.rb:7\n(eval):590:in each\n/artichoke/virtual_root/src/lib/spec_runner.rb:250:in run_specs\n(eval):1",
7199:    "StringScanner#getch is multi-byte character sensitive ERROR\nArgumentError: regex crate utf8 backend for Regexp only supports UTF-8 haystacks\n/artichoke/virtual_root/src/lib/strscan.rb:272:in match\n/artichoke/virtual_root/src/lib/strscan.rb:272:in scan_full\n/artichoke/virtual_root/src/lib/strscan.rb:264:in scan\n/artichoke/virtual_root/src/lib/strscan.rb:112:in getch\n/artichoke/virtual_root/src/lib/library/stringscanner/getch_spec.rb:19:in protect\n(eval):165:in all?\n(eval):590:in each\n/artichoke/virtual_root/src/lib/library/stringscanner/getch_spec.rb:35\n(eval):590:in each\n/artichoke/virtual_root/src/lib/spec_runner.rb:250:in run_specs\n(eval):1",
```

PR #1975, opened 2022-07-25
Add NUL bytes to the end of `String` allocations in `MRB_API` string FFI functions

lopopolo:
This PR is a continuation in the same theme as #1969. #1969 was an incomplete fix for the buffer overread that only targeted strings produced from symbols. This PR targets all strings to ensure they have a trailing NUL byte.

The upstream version of `mrb_str_new_capa` always allocates `capa + 1` bytes and sets the last byte of the allocation to zero.

Prior to this commit, Artichoke did not maintain this invariant. This results in the same style of buffer overread as in https://github.com/artichoke/artichoke/pull/1969 because `mrb_str_dup` is another plausible pathway to get an `RSTRING_PTR` from `mrb_class_path`:

https://github.com/artichoke/artichoke/blob/7bf3f033e020dd60f4b5fd5974dcb19af2617f6d/artichoke-backend/vendor/mruby/src/class.c#L2123-L2127

Commits in this PR ensure all `String` FFI routines route through `mrb_str_new_capa` to maintain this invariant, or otherwise ensure a trailing NUL byte is present similar to the mitigation in https://github.com/artichoke/artichoke/pull/1969.

This was observed as similar corruption/garbage bytes in class names in the spec suite:

For example, these subsequent runs as identified in https://github.com/artichoke/artichoke/pull/1969#issuecomment-1193400913:

```
Method#inspect: .XXXXXXX: "#<Method: MethodSpecs::MySub(MethodSpecs::MyMod)#bar>"
.XXXXXXX: "#<Method: MethodSpecs::MySub(MethodSpecs::MyMod)#bar>"
.XXXXXXX: "#<Method: MethodSpecs::MySub:\xB8\xF7\x7F(MethodSpecs::MyMod)#bar>"
X.XXXXXXX: "#<Method: MethodSpecs::MySub(MethodSpecs::MyMod)#bar>"
X.XXXXXXX: "#<Method: MethodSpecs::MySub(MethodSpecs::MyMod)#bar>"
.XXXXXXX: "#<Method: MethodSpecs::MySubW(MethodSpecs::MyMod)#bar>"
```

```
Method#inspect: .XXXXXXX: "#<Method: MethodSpecs::MySub(MethodSpecs::MyMod)#bar>"
.XXXXXXX: "#<Method: MethodSpecs::MySubmethods'-\xBC\xF7\x7F(MethodSpecs::MyMod)#bar>"
[...]

PR #1976, opened 2022-07-29
Add specialized byte buffer backends to `spinoso-string`

lopopolo:
This PR takes a more robust approach to fixing the buffer over-read memory corruption issues that were mitigated in #1969 and #1975. Rather than playing whack-a-mole at `String` allocation sites to ensure the byte content of the underlying `Vec<u8>` is NUL terminated, this PR bakes this invariant into the type system.

## Implement custom wrappers around `Vec` in `spinoso-string`

Issues https://github.com/artichoke/artichoke/pull/1969 and https://github.com/artichoke/artichoke/pull/1975 fixed memory unsafety due to a failure to uphold mruby's invariants that byte buffers stored in an `RString*` are always NUL terminated.

PRs https://github.com/artichoke/artichoke/pull/1969 and https://github.com/artichoke/artichoke/pull/1975 made tactical fixes at critical "FFI" portions of the `MRB_API` surface that are likely to be used directly by the mruby VM, but these tactical fixes are incomplete – not all `String` allocations went through these code paths, notably `TryConvertMut` implementations and trampoline methods that directly allocate `String` return values.

ca9644e07c007651ffd12e7a0069db9191eaaa4d adds two custom `Vec` types to `spinoso-string` in a new type called `Buf`. These crate-internal `Vec`s will be used by all encoded string types and can be specified at compile time by either enabling or disabling the `always-nul-terminated-c-string-compat` Cargo feature. The `always-nul-terminated-c-string-compat` is disabled by default because enabling it tends to incur extra allocations.

One variant added is a transparent wrapper around `Vec`; it does not change the semantics of any APIs or influence the way the inner `alloc::vec::Vec` allocates. The various encoded string variants in `spinoso-string` are updated to use the exported `Buf` type at `spinoso_string::buf::Buf` as their internal byte buffer.

A second variant is added that on construction (e.g. `new`, `with_capacity`, `From` impls) and mutation (methods that take `&mut self`) ensures that the buffer is NUL terminated in the spare capacity of the `Vec` by using the `Vec::spare_capacity_mut` and `Vec::reserve_exact` APIs. The NUL terminator will be present in the uninitialized memory that is part of the `Vec`'s allocation and follows the `len` of the `Vec`. This is the magic of asserting this invariant, which requires no unsafe code:

```rust
fn ensure_nul_terminated(vec: &mut Vec<u8>) {
    const NUL_BYTE: u8 = 0;

    let spare_capacity = vec.spare_capacity_mut();
    // If the vec has spare capacity, set the first and last bytes to NUL.
    match spare_capacity {
        [] => {}
        [next] => {
            next.write(NUL_BYTE);
            return;
        }
        [head, .., tail] => {
            head.write(NUL_BYTE);
            tail.write(NUL_BYTE);
            return;
        }
    }
    // Else `vec.len == vec.capacity`, so reserve an extra byte.
    vec.reserve_exact(1);
    let spare_capacity = vec.spare_capacity_mut();
    match spare_capacity {
        [] => panic!("Vec should have spare capacity"),
        [next] => {
            next.write(NUL_BYTE);
        }
        [head, .., tail] => {
            head.write(NUL_BYTE);
            tail.write(NUL_BYTE);
        }
    }
}
```

## Compatibility

Several `const` APIs have been made non-`const` to allow for future variants to allocate (for example, in `String::new`). The crate version has been bumped to 0.19.0.

When the `always-nul-terminated-c-string-compat` feature is enabled, Spinoso strings may allocate more than requested to enforce the NUL termination invariant.

The `Buf` variant that is activated when the `always-nul-terminated-c-string-compat` feature is enabled does not export methods for the `Drain` and `Splice` iterators. If this is required in a future iteration, wrappers around the types from `std` must be made to enforce the NUL termination invariant on `drop` of these iterator types.

## Memory Safety

The mitigations in #1969 and #1975 are mostly reverted.

This test from #1975 yields no results and was verified by reverting e37c53c10c41b016be22f5cda796fcf13c6b31e1 and verifying that the memory corruptions still reproduce.

```
cargo run --bin spec-runner -q -- all-core-specs.toml > dump.txt
rg 'ArgumentError: regex crate utf8 backend for Regexp only supports UTF-8 haystacks' dump.txt
```

### DIFF
--- a/artichoke-backend/src/extn/core/string/ffi.rs
+++ b/artichoke-backend/src/extn/core/string/ffi.rs
@@ -27,20 +27,7 @@ use crate::value::Value;
 #[no_mangle]
 unsafe extern "C" fn mrb_str_new_capa(mrb: *mut sys::mrb_state, capa: usize) -> sys::mrb_value {
     unwrap_interpreter!(mrb, to => guard);
-
-    let alloc_capacity = if let Some(capacity) = capa.checked_add(1) {
-        capacity
-    } else {
-        let err = ArgumentError::with_message("string capacity too large");
-        error::raise(guard, err);
-    };
-    // SAFETY: mruby assumes strings are allocated with `capacity = capa + 1`
-    // and that last byte is NUL.
-    let mut result = String::with_capacity(alloc_capacity);
-    let ptr = result.as_mut_ptr();
-    let last = ptr.add(capa);
-    *last = 0;
-
+    let result = String::with_capacity(capa);
     let result = String::alloc_value(result, &mut guard);
     match result {
         Ok(value) => value.inner(),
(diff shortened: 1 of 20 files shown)


lopopolo:
for those interested, this PR is best reviewed commit-by-commit.

artichoke-backend/src/extn/core/string/ffi.rs
```diff
@@ -27,20 +27,7 @@ use crate::value::Value;
...
-        error::raise(guard, err);
-    };
-    // SAFETY: mruby assumes strings are allocated with `capacity = capa + 1`
-    // and that last byte is NUL.
-    let mut result = String::with_capacity(alloc_capacity);
-    let ptr = result.as_mut_ptr();
-    let last = ptr.add(capa);
-    *last = 0;
-
+    let result = String::with_capacity(capa);
```
lopopolo, line 30: the behavior here is different than above and different than mruby.

mruby always sets the last byte of the allocation to NUL, whereas this code will set the first byte (the beginning of the spare capacity).

https://github.com/artichoke/artichoke/blob/16c869a9ad29acfe143bfcc011917ef442ccac54/artichoke-backend/vendor/mruby/src/string.c#L36-L38

I think probably the right fix is to set the first _and_ last bytes of the `Vec`'s spare capacity to zero.

lopopolo:
The _Check spinoso and scoalapasta crates_ job did not check the compiler errors in the latest build since it only runs `cargo check`, but these errors were errors in doctests.

PR #1984, opened 2022-07-30
Add tests and update patterns for parser `Context`

lopopolo:
- Use deref syntax instead of `as_ref()`.
- Add tests for filenames safety with `Context::new_unchecked` by calling `Context::new(...).unwrap()`.
- Collapse `cow.clone().into_owned()` into `cow.clone()` for APIs that take `Into<Vec<u8>>`.


lopopolo:
just a bit of micro refactoring

PR #1987, opened 2022-07-30
Cache the local timezone in a `static`

lopopolo:
Per the docs for `tzdb::local_tz`, it is suggested to cache the result of fetching the local timezone to avoid repeatedly accessing the filesystem on some platforms.

See: https://docs.rs/tzdb/latest/tzdb/fn.local_tz.html.

This commit stores the local TZ in a static member in `Offset::local`.


b-n (approved):
:bow: Thanks!

PR #1988, opened 2022-07-31
Fixup static offset parsing for tzrs times

lopopolo:
- Rather than check `Regex::is_match` and unwrapping `Regex::captures`, just call `Regex::captures` and pattern match on the returned `Option`.
- Use slice indexing on `Captures` since we know the groups are not optional and will be present in the match. This makes the code a bit easier to read by hiding the unwraps and `str` casting.
- Ensure only ASCII digits can match in the parsing `Regex`, add a test which failed previously.

### DIFF
--- a/spinoso-time/src/time/tzrs/offset.rs
+++ b/spinoso-time/src/time/tzrs/offset.rs
@@ -437,6 +437,22 @@ mod tests {
     }
 
     #[test]
+    fn from_str_fixed_strings_with_newlines() {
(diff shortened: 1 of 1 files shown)


lopopolo:
tweeted about this here: https://twitter.com/artichokeruby/status/1553535513742622720

b-n (approved):
Some excellent catching! It would be kinda nice if +१०:೩೬ would parse into i32 though  :grimacing:

PR #1990, opened 2022-07-31
Refactor `tzrs` offset parsing code for clarity

lopopolo:
- Use `(n..=m).contains(...)` to enforce range bounds on minutes and hours components of the offset for clarity and to prevent accidentally allowing negative offset parts.
- Remove `unicode-perl` feature from `regex` dependency in `spinoso-time` since #1988 removed use of the `\d` character class.


lopopolo:
we chatted about this on Discord so I'm not gonna block on your review @b-n

b-n (approved):
Makes all the more sense!

PR #2053, opened 2022-08-07
Extract `Kernel#Integer` to a new crate, `scolapasta-int-parse`

### LINKED ISSUE #2051: Split out `Kernel#Integer` parser into a `scolapasta` crate
The string + radix parser for `Kernel#Integer` is a chunky and largely standalone component in `artichoke-backend` that should be extracted into its own crate. `scolapasta-intparser` sounds good to me.

Relevant code includes `Radix`, `IntegerString`, and `extn::kernel::integer::method`:

https://github.com/artichoke/artichoke/blob/e808e43f574e672e78726e2a33375714349ec15f/artichoke-backend/src/extn/core/kernel/integer.rs#L11-L1190

This is a followup to https://github.com/artichoke/artichoke/pull/2043.

lopopolo:
Extract the parser and implementation of `Kernel#Integer` to its own crate, a new _scolapasta_ crate.

This PR performs the extraction 1:1, ports `kernel::trampoline` to the new crate, and then does a v0.2.0 rewrite to better handle the vagaries of radix parsing. The signature for `parse` looks like:

```rust
pub fn parse<T>(subject: &T, radix: Option<i64>) -> Result<i64, Error<'_>>
where
    T: AsRef<[u8]> + ?Sized,
{
    // ...
}
```

`scolapasta-int-parse` is `no_std` + `alloc`, has 100% documentation and example coverage, and a boatload of tests.

This new implementation handles more radix edge cases, has proper exception messages, and throws `RangeError` instead of `ArgumentError` when the radix is outside of `i32` range.

Examples of new radix behavior that is captured:

- `-1` parses to the `None` radix.
- Negative invalid radixes are ignored when string input has `0x...` style prefixes.
- Radixes `< i32::MIN` and `> i32::MAX` are `RangeError`s but have different error messages.
- Negative radixes are rounded to positive and reported as positive in "invalid radix" error messages.

Some additional FIXMEs have been identified where `Kernel#Integer` in `kernel::trampoline` needs to handle `i64` and `f64` arguments.

Closes https://github.com/artichoke/artichoke/issues/2051.


lopopolo:
markdown link check failure is expected since this is a new crate and the docs aren't pushed to GitHub Pages yet.

artichoke-backend/src/extn/core/kernel/trampoline.rs
```diff
@@ -1,14 +1,67 @@
...
-    let integer = kernel::integer::method(arg, base)?;
+    let base: Option<Radix> = interp.try_convert_mut(base)?;
+
+    // Implicit conversions are only performed if a non-nil radix argument is
+    // given:
+    //
+    // ```
+    // [3.1.2] > class A; def to_str; "1234"; end; end
+    // => :to_str
+    // [3.1.2] > Integer(A.new)
```
lopopolo, line 47: What about the case where an explicit `nil` is given?

artichoke-backend/src/extn/core/kernel/trampoline.rs
```diff
@@ -1,14 +1,67 @@
...
+//         from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `<main>'
+// [3.1.2] > Integer Float::INFINITY
+// (irb):44:in `Integer': Infinity (FloatDomainError)
+//         from (irb):44:in `<main>'
+//         from /usr/local/var/rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
+//         from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `load'
+//         from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `<main>'
+// ```
+pub fn integer(interp: &mut Artichoke, mut subject: Value, base: Option<Value>) -> Result<Value, Error> {
     let base = base.and_then(|base| interp.convert(base));
```
lopopolo, line 32: Add a comment about what this does: flatten arg not given and nil given to None.

scolapasta-int-parse/src/parser.rs
```diff
@@ -0,0 +1,81 @@
...
+        // [3.1.2] > -(2 ** 64)
+        // => -18446744073709551616
+        // [3.1.2] > (-(2 ** 64 - 1)).to_s.length
+        // => 21
+        // ```
+        //
+        // In bases below 10, the string repr for large numbers will be longer,
+        // but pre-allocating for these uncommon cases seems wasteful. The
+        // `String` will reallocate if it needs to in these pathological cases.
+        const MAX_REPR_LENGTH: usize = 21;
```
lopopolo: Rename this constant. It's not the max.

scolapasta-int-parse/src/radix.rs
```diff
@@ -0,0 +1,252 @@
+use core::num::NonZeroU32;
+
+// Create a lookup table from each byte value (of which the ASCII range is
+// relevant) to the maximum minimum radix the character is valid for.
+#[allow(clippy::cast_possible_truncation)]
```
lopopolo: This lint can be removed if the loop counter is a `u8` and the as cast widens to `usize`

scolapasta-int-parse/src/radix.rs
```diff
@@ -0,0 +1,252 @@
...
+const fn radix_table() -> [u32; 256] {
+    // `u32::MAX` is used as a sentinel such that no valid radix can use this
+    // byte.
+    let mut table = [u32::MAX; 256];
+    let mut idx = 0_usize;
+    loop {
+        if idx >= table.len() {
```
lopopolo: Move this check to right before the index increment to prevent overflowing `u8::MAX`

artichoke-backend/src/extn/core/kernel/trampoline.rs
lopopolo: added a snippet

artichoke-backend/src/extn/core/kernel/trampoline.rs
lopopolo: done and verified the behavior in MRI

scolapasta-int-parse/src/parser.rs
lopopolo: Renamed to `PRE_ALLOCATED_DIGIT_CAPACITY`.

scolapasta-int-parse/src/radix.rs
lopopolo: done

scolapasta-int-parse/src/radix.rs
lopopolo: done

lopopolo:
Some additional FIXMEs have been identified where `Kernel#Integer` in `kernel::trampoline` needs to handle `i64` and `f64` arguments.

PR #2069, opened 2022-08-10
Fix `Kernel#Integer` failing specs

### LINKED ISSUE #2067: `Kernel#Integer` no longer passing `rake spec`
I was just running spec running and came across the following errors (redacted for brevity):

```
Kernel.Integer ignores trailing whitespace
ArgumentError: invalid value for Integer(): "1 "
/artichoke/virtual_root/src/lib/core/kernel/Integer_spec.rb:185:in protect
...

Kernel.Integer parses the value as 0 if the string consists of a single zero character
ArgumentError: invalid value for Integer(): "0"
/artichoke/virtual_root/src/lib/core/kernel/Integer_spec.rb:278:in protect
...

Kernel.Integer ignores trailing whitespace
ArgumentError: invalid value for Integer(): "16 "
/artichoke/virtual_root/src/lib/core/kernel/Integer_spec.rb:390:in protect
...

Kernel#Integer ignores trailing whitespace
ArgumentError: invalid value for Integer(): "1 "
/artichoke/virtual_root/src/lib/core/kernel/Integer_spec.rb:185:in protect
...

Kernel#Integer parses the value as 0 if the string consists of a single zero character
ArgumentError: invalid value for Integer(): "0"
/artichoke/virtual_root/src/lib/core/kernel/Integer_spec.rb:278:in protect
...

Kernel#Integer ignores trailing whitespace
ArgumentError: invalid value for Integer(): "16 "
/artichoke/virtual_root/src/lib/core/kernel/Integer_spec.rb:390:in protect
...

MRI 3.1.2

```irb
irb(main):001:0> Integer("1 ")
=> 1
irb(main):002:0> Integer("0")
=> 0
```

# Analysis

I assume this was introduced in #2053

The following is not supported in `scolapasta-int-parse`:
- Trailing whitespace
- `"0"` string (I assume because of removing leading zeros)

I added these tests to `lib.rs` and got failures, so this might be something to test against:

```rust
    #[test]
    fn decimal_zero() {
        let result = parse("0", None);
        assert_eq!(result.unwrap(), 0);
    }

    #[test]
    fn trailing_whitespace() {
        let result = parse("1 ", None);
        assert_eq!(result.unwrap(), 1);
    }
```

lopopolo:
- Trim leading _and_ trailing ASCII whitespace.
- Trim _any_ ASCII whitespace, not just spaces (` `).
- Treat a single (potentially whitespace padded) `0` as parsing to `0`.

Fixes https://github.com/artichoke/artichoke/issues/2067.

Thanks for the report @b-n.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -609,7 +609,7 @@ version = "0.2.0"
 
 [[package]]
 name = "scolapasta-int-parse"
-version = "0.2.0"
+version = "0.2.1"
 dependencies = [
  "scolapasta-string-escape",
 ]
--- a/artichoke-backend/Cargo.toml
+++ b/artichoke-backend/Cargo.toml
@@ -25,7 +25,7 @@ qed = "1.3.0"
 # See: CVE-2022-24713
 # https://github.com/artichoke/artichoke/pull/1729
 regex = "1.5.5"
-scolapasta-int-parse = { version = "0.2.0", path = "../scolapasta-int-parse", default-features = false }
+scolapasta-int-parse = { version = "0.2.1", path = "../scolapasta-int-parse", default-features = false }
 scolapasta-string-escape = { version = "0.3.0", path = "../scolapasta-string-escape", default-features = false }
 spinoso-array = { version = "0.9.0", path = "../spinoso-array", default-features = false }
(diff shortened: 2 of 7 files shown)


b-n (approved):
Approving before CI run, but all changes make sense!

lopopolo:
markdown checks are being rate limited

PR #2102, opened 2022-08-24
Push fast paths in String::char_len down into encoded variants 

### LINKED ISSUE #1722: Push fast paths in `String::char_len` down into encoded variants
I agree with your instinct here. It'd also allow us to only pay the branching cost once, for example we branch in `char_len` above in the short circuiting logic.

Happy for this to be addressed in a followup.

I think the short circuiting logic is most useful for UTF-8 case since that is the expensive one to compute.

_Originally posted by @lopopolo in https://github.com/artichoke/artichoke/pull/1678#r820131264_

not only in the `char_len`, but also in `len()` too (double cost). So yeah, #future for sure. It feels like `String` in an ideal world would just redirect all calls to `EncodedString`, and if the logic is duplicated in the encoded type, then so be it. This will be much easier/clearer in a follow up PR though as you said :+1:

_Originally posted by @b-n in https://github.com/artichoke/artichoke/pull/1678#r820231870_

choznerol:
Close #1722


spinoso-string/src/lib.rs
```diff
@@ -1916,6 +1916,20 @@ impl String {
...
+    /// [3.0.1] > "aaa"[2..1]
+    /// => ""
+    /// ```
+    ///
+    /// but only if `start` is within the string.
+    /// ```
+    /// [3.0.1] > "aaa"[10..0]
+    /// => nil
+    /// ```
+    ///
```
choznerol: The comments were very helpful at least for me to understand the early return logic, but I was hesitant to duplicate them in all 3 encoded variants, so eventually I deleted most of them.

spinoso-string/src/lib.rs
```diff
@@ -1933,6 +1947,30 @@ impl String {
     /// assert_eq!(s.get_char_slice(Range { start: 15, end: 10 }), None);
     /// assert_eq!(s.get_char_slice(Range { start: 15, end:  1 }), None);
     /// assert_eq!(s.get_char_slice(Range { start:  4, end:  1 }), Some(&b""[..]));
+    ///
+    /// let s = String::ascii(b"abcd".to_vec());
```
choznerol: I added related tests to the doctest of `String#get_char_slice` before refactoring. If unit tests (for each encoded variant?) are preferred in this case, please let me know, and I would be happy to refactor.

spinoso-string/src/enc/ascii/mod.rs
```diff
@@ -198,7 +198,15 @@ impl AsciiString {
...
         let Range { start, end } = range;
 
-        self.inner.get(start..end).or_else(|| self.inner.get(start..))
+        if start > self.len() || start > self.char_len() {
+            return None;
+        }
+
+        return match end as isize - start as isize {
+            d if d <= 0 => Some(&[]),
+            1 => self.get_char(start),
```
choznerol: I added `as isize` otherwise `end - start` would be `usize` and the `<= 0` branch would not work due to `warning: comparison is useless due to type limits`.

spinoso-string/src/enc/ascii/mod.rs
```diff
@@ -198,7 +198,15 @@ impl AsciiString {
     pub fn get_char_slice(&self, range: Range<usize>) -> Option<&'_ [u8]> {
         let Range { start, end } = range;
 
-        self.inner.get(start..end).or_else(|| self.inner.get(start..))
```
lopopolo, line 201: The fast path is only really useful for UTF-8 encoded strings because it allows us to avoid a full UTF-8 decode of the bye buffer.

It looks like these changes to the `Ascii` and `Binary` variants don't change the behavior but add code.

Can we add unit tests for `Ascii` and `Binary` to check this behavior and then revert the changes in these variants?

spinoso-string/src/enc/ascii/mod.rs
lopopolo: I'm hesitant to use `as isize` casting here when the previous version of these fast paths did not.

spinoso-string/src/enc/binary/mod.rs
```diff
@@ -196,7 +196,15 @@ impl BinaryString {
     pub fn get_char_slice(&self, range: Range<usize>) -> Option<&'_ [u8]> {
         let Range { start, end } = range;
 
-        self.inner.get(start..end).or_else(|| self.inner.get(start..))
```
lopopolo, line 199: Same feedback as the `Ascii` variant: I don't think this fast path adds any behavior we don't have with the slice indexing. Slice indexing is very cheap, so I don't think we need to bloat this method with more code to achieve the same thing.

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -287,6 +287,17 @@ impl Utf8String {
     pub fn get_char_slice(&self, range: Range<usize>) -> Option<&'_ [u8]> {
         // TODO: use fast path rejection from `get_char` here too
         let Range { start, end } = range;
+
+        if start > self.len() || start > self.char_len() {
+            return None;
+        }
```
lopopolo: in the old code, this check was nested within a `if end < start` condition so we can avoid a UTF-8 decode unless we know we're short circuiting.

```suggestion
        // Fast path the lookup if the end of the range is before the start.
        if end < start {
            // Yes, these types of ranges are allowed and they return `""`,
            // but only if `start` is within the string.
            //
            // Attempt to short-circuit with a cheap length retrieval.
            if start > self.len() || start > self.char_len() {
                return None;
            }
            return Some(&[]);
        }
```

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -287,6 +287,17 @@ impl Utf8String {
...
+
+        match end as isize - start as isize {
+            d if d <= 0 => return Some(&[]),
```
lopopolo: because we wanted to avoid a UTF-8 decode above, we don't need to check for the end before start case and don't need to cast to `isize`.

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -287,6 +287,17 @@ impl Utf8String {
...
+            1 => return self.get_char(start),
+            _ => {}
+        };
+
```
lopopolo, line 397: I think we'll want to preserve this fast path:

```rust
        // If the start of the range is beyond the character count of the
        // string, the whole lookup must fail.
        //
        // Slice lookups where the start is just beyond the last character index
        // always return an empty slice.
        //
        // Fast path rejection for indexes beyond bytesize, which is cheap to
        // retrieve.
        if start > self.len() {
            return None;
        }
```

lopopolo: and this one:

```rust
        match self.char_len() {
            char_length if start > char_length => return None,
            char_length if start == char_length => return Some(&[]),
            _ => {}
        }

        // The span is guaranteed to at least partially overlap now.
        match end - start {
            0 => return Some(&[]),
            1 => return self.get_char(start),
            _ => {}
        }
```

spinoso-string/src/lib.rs
lopopolo: thanks for adding them to the docs!

spinoso-string/src/lib.rs
lopopolo: I think since we've pushed the logic into the encoded variants, I'd prefer these to be unit tests in each of the encoded buf types.

Ideally the doc tests would only show examples of how to use the APIs. In addition to moving all of the doc tests to unit tests, do you want to take a stab at simplifying the examples to maybe show only one `Some` and `None` case for UTF-8 and ASCII?

lopopolo (changes requested):
Thanks for the PR @choznerol. I think the UTF-8 fast path has different behavior than the code today and I like your idea to move the doc tests to unit tests on each encoded variant.

spinoso-string/src/enc/ascii/mod.rs
choznerol: That's surely a better approach! I added unit tests for `Ascii` and `Binary` in 1909d91dfbdc1e9bf2192df3932799620643e474 but it looks like there is one edge case not considered by the current implementation, that is, `"abc"[2..1]` would be `"c"` instead of `""`. I added 78e6f49c97ff02448ffcba48f59b7673f8702588 to address this case.

spinoso-string/src/enc/binary/mod.rs
choznerol: Same as above https://github.com/artichoke/artichoke/pull/2102#discussion_r962317754

spinoso-string/src/lib.rs
choznerol: I have added unit tests for the three encoded variants, and then simplify the doc tests in  b4c8746c6d8ed4329a15036e52a8f0aee0e7ee78.

choznerol:
Thanks for the detailed feedback! After realizing that the early return logic is only meant for the UTF8 variant, I redo the commits to preserve most of the original logic and the helpful comment only in the UTF8 variant. However, after adding more tests, I believe some changes still need to be made on Ascii and Binary, but this time I tried to do them in a more concise way.

P.s. the [PR close](https://github.com/artichoke/artichoke/pull/2102#event-7318518359) was an accident :sweat_smile:

spinoso-string/src/enc/ascii/mod.rs
```diff
@@ -198,7 +198,15 @@ impl AsciiString {
...
+        self.inner.get(start..end).or_else(|| {
+            if start > self.inner.len() {
+                None
+            } else if end <= start {
+                Some(&b""[..])
```
lopopolo: ooo nice catch!

spinoso-string/src/enc/ascii/mod.rs
```diff
@@ -517,4 +526,26 @@ mod tests {
...
+    }
+
+    #[test]
+    fn get_char_slice_invalid_range() {
+        let s = AsciiString::from("abc");
+        assert_eq!(s.get_char_slice(Range { start: 4, end: 5 }), None);
+        assert_eq!(s.get_char_slice(Range { start: 4, end: 1 }), None);
+        assert_eq!(s.get_char_slice(Range { start: 3, end: 1 }), Some(&b""[..]));
+        assert_eq!(s.get_char_slice(Range { start: 2, end: 1 }), Some(&b""[..]));
+    }
```
lopopolo: thank you!

spinoso-string/src/enc/ascii/mod.rs
lopopolo: This removes a few symbols to make this line read clearer.

```suggestion
                Some(&[])
```

spinoso-string/src/enc/binary/mod.rs
```diff
@@ -196,7 +196,15 @@ impl BinaryString {
...
+        self.inner.get(start..end).or_else(|| {
+            if start > self.inner.len() {
+                None
+            } else if end <= start {
+                Some(&b""[..])
```
lopopolo: This removes a few symbols to make this line read clearer.

```suggestion
                Some(&[])
```

spinoso-string/src/lib.rs
```diff
@@ -1922,17 +1922,13 @@ impl String {
...
     ///
-    /// let s = String::utf8(b"abc\xF0\x9F\x92\x8E\xFF".to_vec()); // "abc💎\xFF"
-    /// assert_eq!(s.get_char_slice(Range { start:  0, end:  1 }), Some(&b"a"[..]));
-    /// assert_eq!(s.get_char_slice(Range { start:  0, end:  3 }), Some(&b"abc"[..]));
-    /// assert_eq!(s.get_char_slice(Range { start:  0, end:  4 }), Some("abc💎".as_bytes()));
-    /// assert_eq!(s.get_char_slice(Range { start:  0, end:  5 }), Some(&b"abc\xF0\x9F\x92\x8E\xFF"[..]));
-    /// assert_eq!(s.get_char_slice(Range { start:  3, end: 10 }), Some(&b"\xF0\x9F\x92\x8E\xFF"[..]));
-    /// assert_eq!(s.get_char_slice(Range { start:  4, end: 10 }), Some(&b"\xFF"[..]));
+    /// let s = String::ascii(b"abc".to_vec());
+    /// assert_eq!(s.get_char_slice(Range { start:  1, end:  3 }), Some("bc".as_bytes()));
```
lopopolo: Thanks for simplifying these tests. Now that I'm taking another peek here, I think we can also make an opportunistic refactor to replace the range structs with range literals. Something like this:

```suggestion
    /// assert_eq!(s.get_char_slice(1..3), Some("bc".as_bytes()));
```

spinoso-string/src/lib.rs
```diff
@@ -1922,17 +1922,13 @@ impl String {
     /// use core::ops::Range;
```
lopopolo: if we make the change to remove the range struct instantiation, let's remove this import in the doc test

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -1082,4 +1191,35 @@ mod tests {
...
+        assert_eq!(
+            s.get_char_slice(Range { start: 1, end: 3 }),
+            Some(&b"\xF0\x9F\x92\x8E\xFF"[..])
+        );
+    }
+
+    #[test]
+    fn get_char_slice_invalid_range() {
+        let s = Utf8String::from(b"a\xF0\x9F\x92\x8E\xFF".to_vec()); // "a💎\xFF"
+        assert_eq!(s.get_char_slice(Range { start: 4, end: 5 }), None);
```
lopopolo: can we add the following test cases to ensure we have branch coverage on all parts of the fast path?

- `Range { start: 7, end: 10 }`
- `Range { start: 10, end: 8 }`
- `Range { start: 10, end: 5 }`
- `Range { start: 10, end: 2 }`

lopopolo (changes requested):
Thanks for reworking this PR @choznerol. I like the way it looks 💯. I left a couple of suggestions and I'd appreciate it if you could make another small change to the doc tests to make them a bit more readable by using range literals like `1..3`.

spinoso-string/src/lib.rs
lopopolo: thank you!

spinoso-string/src/enc/ascii/mod.rs
lopopolo: awesome and thanks for catching that bug! Glad to have the added tests.

spinoso-string/src/enc/utf8/mod.rs
choznerol: Added in d25a1c2

spinoso-string/src/enc/ascii/mod.rs
choznerol: Fixed [here](https://github.com/artichoke/artichoke/pull/2102/files#diff-1578dd864d1a9bac39c945d985837664c4c4655d41f16fdba68ea16ecef49c4aR205)

spinoso-string/src/enc/binary/mod.rs
choznerol: Fixed [here](https://github.com/artichoke/artichoke/pull/2102/files#diff-9a6cae435431760640daec73e708f6067634165194ea91e3c7f63213f8b48c28R203)

spinoso-string/src/lib.rs
choznerol: Fixed [here](https://github.com/artichoke/artichoke/pull/2102/files?diff=split&w=0#diff-a2d2fa818d5b61fb277dd19f5a3a58b696944224b071fe8faee5bca3c19103d3R1925-R1930)

Thanks for the suggestion! I also updated the [unit tests](https://github.com/artichoke/artichoke/pull/2102/files?diff=split&w=0#diff-edc577dd37d438dce282a0d8476d38f20897d1052364a5e8ef495bf09aa69852R1197-R1204) to use Range literals.

spinoso-string/src/lib.rs
choznerol: Oops! Thanks for catching this. Removed in the amended 681d619

spinoso-string/src/lib.rs
choznerol: > I also updated the [unit tests](https://github.com/artichoke/artichoke/pull/2102/files?diff=split&w=0#diff-edc577dd37d438dce282a0d8476d38f20897d1052364a5e8ef495bf09aa69852R1197-R1204) to use Range literals.

Opps I realized Clippy will [complain](https://github.com/artichoke/artichoke/runs/8207578734?check_suite_focus=true) for [reversed_empty_ranges](https://rust-lang.github.io/rust-clippy/master/index.html#reversed_empty_ranges) when I use `3..1` instead of `Range { start: 3, end: 1 }`. I added [`allow(clippy::reversed_empty_ranges)`](https://github.com/artichoke/artichoke/pull/2102/commits/229c2454034463cc7cc438037aea04bbec9c173e#diff-1578dd864d1a9bac39c945d985837664c4c4655d41f16fdba68ea16ecef49c4aR535) for now for all three `get_char_slice_invalid_range` unit tests, but please let me know if reverting back to using `Range` is preferred!

choznerol:
Thanks for the review! I have fix according to the comments and also cleaned up the old commits.

lopopolo (approved):
Thank you for the patch @choznerol. Performance improvements and tests and better docs and cleaner code. Big win!

PR #2104, opened 2022-08-27
Strip binaries in release mode with Cargo

lopopolo:
https://doc.rust-lang.org/cargo/reference/profiles.html#strip

Support for this landed in Rust toolchain version 1.59.0, but was not enabled at the time:

- https://github.com/artichoke/artichoke/pull/1753


lopopolo:
The corresponding change in the nightly builder landed here:

- https://github.com/artichoke/nightly/pull/73

PR #2105, opened 2022-08-27
Extract path helpers from `mezzaluna-feature-loader` to a new `scolapasta-path` crate

lopopolo:


lopopolo:
link check failed because docs aren't published at https://artichoke.github.io/artichoke/scolapasta_path/ yet.

lopopolo:
This should have bumped the version of `mezzaluna-feature-loader` but did not.

PR #2107, opened 2022-08-27
Remove `artichoke_backend::string`, add `artichoke_backend::fmt`

lopopolo:
Remove thin wrapper around `scolapasta_string_escape::format_debug_escape_into`. This only left `WriteError` in `artichoke_backend::string`. `WriteError` exists to bridge `std::fmt` to Artichoke's `Error` type and exception hierarchy. Rename `artichoke_backend::string` to `artichoke_backend::fmt` to reflect this.

`artichoke` crate gains a direct dependency on `scolapasta-string-escape`.


lopopolo:
This commit removes a bunch of `impl From<...> for Error` impls for various combinations of boxed `WriteError`.

PR #2166, opened 2022-09-02
Remove chrono backend from spinoso-time

lopopolo:
All of the recent investment in `spinoso-time` has been poured into the
`tz-rs` backend and we've got a high-quality time implementation with
the `tzrs` feature.

The `chrono` backend has atrophied since it was introduced. The API has
diverged from the standard set by the `tz-rs` backend (many APIs which
are fallible in the `tz-rs` backend are infallible in the `chrono`
backend). Full support for the offset API is not something that is
easily achievable with `chrono`.

`chrono` itself is difficult to work with. For example, many APIs panic
on out of range inputs. The API is overly generic which forces
`spinoso-time` to deconstruct a `DateTime` into its raw parts and use
`chrono` types as "views" into the calendar in each accessor method.

`chrono` with time zone support is also a heavy dependency. Removing
this backend results in 13 crates being dropped from `Cargo.lock`.

Especially given the good collaboration we've had with @x-hgg-x (`tz-rs`)
and @Kijewski (`tzdb`, `iana-time-zone`), I don't see a compelling reason
to continue maintaining the `chrono` backend in `spinoso-time`.


lopopolo:
@b-n I hope you're enjoying your holiday. No rush on this one, but I'd love to get your input and review on this before merging.

lopopolo:
check links job is not cooperating, but no links in markdown sources are broken (or changed?) as far as I know

Cargo.toml
```diff
@@ -133,8 +133,8 @@ core-regexp = ["artichoke-backend/core-regexp"]
 # backend they execute on. The `regex` crate backend remains the default as long
 # as it can parse the given pattern.
 core-regexp-oniguruma = ["core-regexp", "artichoke-backend/core-regexp-oniguruma"]
-# Implement the `Time` core class. This feature adds dependencies on `chrono`
-# and `chrono-tz`.
+# Implement the `Time` core class. This feature adds dependencies on `tz-rs` and
```
b-n, line 136: Good catch!

b-n, line 136: Good catch!

artichoke-backend/src/extn/core/time/mod.rs
```diff
@@ -14,11 +14,11 @@
 //! ```
 //!
 //! This implementation of `Time` supports the system clock via the
-//! [`chrono`] and [`chrono-tz`] crates.
+//! [`tz-rs`] and [`tzdb`] crates.
```
b-n: I wonder if just a reference to `spinoso-time` itself is better than the packages that `spinoso-time` relies on. But just a minor nit /shrug

spinoso-time/Cargo.toml
```diff
@@ -14,14 +14,11 @@ keywords = ["artichoke", "date", "datetime", "spinoso", "time"]
 categories = ["date-and-time"]
 
 [features]
-default = ["chrono", "tzrs", "tzrs-local"]
-chrono = ["dep:chrono", "dep:chrono-tz"]
+default = ["tzrs", "tzrs-local"]
```
b-n, line 17: ❤️

spinoso-time/README.md
```diff
@@ -38,32 +37,30 @@ Add this to your `Cargo.toml`:
...
 // Extract the Unix timestamp.
 let timestamp = utc.to_int();
 ```
 
 ## Crate features
 
-This crate supports two backends which are independent of each other. The
+This crate supports several backends which are independent of each other. The
 availability of different backends is controlled by Cargo features, all of which
 are enabled by default:
```
b-n:
```suggestion
This crate can support several backends, which are designed to be independent
of each other. The availability of different backends is controlled by Cargo
features, all of which are enabled by default:
```

spinoso-time/src/lib.rs
```diff
@@ -34,18 +34,15 @@
...
+//! information, and combines with the [`tz-rs`] crate to generate the time.
 //!
 //! # Crate features
 //!
-//! This crate supports two backends which are independent of each other. The
-//! availability of different backends is controlled by Cargo features, all of
-//! which are enabled by default:
+//! This crate supports several backends which are independent of each other.
+//! The availability of different backends is controlled by Cargo features, all
+//! of which are enabled by default:
```
b-n:
```suggestion
//! This crate can support several backends, which are designed to be
//! independent of each other. The availability of different backends is
//! controlled by Cargo features, all of which are enabled by default:
```

spinoso-time/src/time/mod.rs
```diff
@@ -5,25 +5,18 @@
 //! a timezone offset. These timestamps can be used to implement the Ruby `Time`
 //! core class.
 //!
-//! There are two independent backends which can be selected by specifying the
-//! `chrono` or `tzrs` feature:
+//! There are several independent backends which can be selected by specifying
+//! the appropriate feature:
```
b-n, line 9:
```suggestion
//! There can be several independent backends which can be selected by
//! specifying the appropriate feature:
```

b-n (approved):
Something beautiful about so much 🟥. I've put a couple of suggestions around language of "can support" instead of "are" - but they're purely contextual nits, and not super critical IMO.

artichoke-backend/src/extn/core/time/mod.rs
lopopolo: done

spinoso-time/README.md
lopopolo: done locally

spinoso-time/src/time/mod.rs
lopopolo: done locally

PR #2170, opened 2022-09-04
Add wasi platform support to scolapasta-path

### LINKED ISSUE #2111: Add `wasi` platform support to `scolapasta-path`
Path and platform string routines were extracted to `scolapasta-path` in https://github.com/artichoke/artichoke/pull/2105 and https://github.com/artichoke/artichoke/pull/2110.

Both modules in `scolapasta-path` define platform-specific implementations for:

- `cfg(windows)`
- `cfg(unix)`
- "default" `cfg(not(any(windows, unix))`

The `unix` modules use special extension traits which enable lossless conversion from `OsStr`/`OsString` to bytes: https://doc.rust-lang.org/std/os/unix/ffi/index.html.

`wasi` platform has a similar set of extension traits that enable lossless conversions: https://doc.rust-lang.org/std/os/wasi/ffi/index.html.

Add `wasi` platform support using these extension traits and sorting out the `cfg`s.

I don't suspect we'll add `wasm32-wasi` builds to CI. Testing locally should be sufficient for merging a patch for this. These functions are fairly simple and I don't expect them to change much.

jcaesar:
Takes care of #2111 without code duplication.

Fixes #2111.

### DIFF
--- a/scolapasta-path/src/paths.rs
+++ b/scolapasta-path/src/paths.rs
@@ -9,15 +9,15 @@
 use std::path::{Path, PathBuf};
 
 mod default;
-#[cfg(unix)]
-mod unix;
+#[cfg(any(unix, target_os = "wasi"))]
+mod unix_wasi;
 #[cfg(windows)]
 mod windows;
 
-#[cfg(not(any(unix, windows)))]
+#[cfg(not(any(unix, windows, target_os = "wasi")))]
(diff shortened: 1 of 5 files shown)


lopopolo (approved):
This is awesome! I like that you've found a way to combine the unix module with this new wasi code.

Thank you! 💎

PR #2182, opened 2022-09-14
Add `Time#{strftime|inspect|to_s|asctime|ctime}` implementations!

b-n:
These are now able to be added with the creation of `strftime-ruby` and the recent changes to `spinoso-time` to enable these fns.

`strftime-ruby` becomes a first level dep of artichoke since spinoso-time isn't wrapping the error enums there. If we don't want this direct dep, then we would need to wrap the error enum directly in spinoso-time and/or re-export it.

## RuntimeError

The `RuntimeError` is the closest guess I could think to what would be nice to have here. (I didn't look, but I believe this is a int bounds/memory bounds problem :grimacing:)

MRI 3.1.2.
```console
(irb):1:in `strftime': Numerical result out of range - %099999999a (Errno::ERANGE)
	from (irb):1:in `<main>'                                              
	from /home/ben/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
	from /home/ben/.rbenv/versions/3.1.2/bin/irb:25:in `load'             
	from /home/ben/.rbenv/versions/3.1.2/bin/irb:25:in `<main>'
```

With these changes:
```console
>>> Time.now.strftime("%099999999a")
	2: from (airb):3
	1: from (airb):3:in strftime
RuntimeError (Unexpected failure)
```

### DIFF
--- a/artichoke-backend/Cargo.toml
+++ b/artichoke-backend/Cargo.toml
@@ -39,7 +39,6 @@ spinoso-securerandom = { version = "0.2.0", path = "../spinoso-securerandom", op
 spinoso-string = { version = "0.20.0", path = "../spinoso-string", features = ["always-nul-terminated-c-string-compat"] }
 spinoso-symbol = { version = "0.3.0", path = "../spinoso-symbol" }
 spinoso-time = { version = "0.7.0", path = "../spinoso-time", features = ["tzrs"], default-features = false, optional = true }
-strftime-ruby = { version = "1.0.0", optional = true }
 
 [dev-dependencies]
 quickcheck = { version = "1.0.3", default-features = false }
@@ -70,7 +69,7 @@ core-math-full = ["core-math", "spinoso-math?/full"]
 core-random = ["dep:spinoso-random"]
 core-regexp = ["dep:spinoso-regexp"]
 core-regexp-oniguruma = ["core-regexp", "spinoso-regexp?/oniguruma", "dep:onig"]
-core-time = ["dep:spinoso-time", "dep:strftime-ruby"]
+core-time = ["dep:spinoso-time"]
 
 load-path-native-file-system-loader = ["artichoke-load-path/native-file-system-loader"]
 load-path-rubylib-native-file-system-loader = ["load-path-native-file-system-loader", "artichoke-load-path/rubylib-native-file-system-loader"]
--- a/artichoke-backend/Cargo.toml
+++ b/artichoke-backend/Cargo.toml
@@ -39,6 +39,7 @@ spinoso-securerandom = { version = "0.2.0", path = "../spinoso-securerandom", op
 spinoso-string = { version = "0.20.0", path = "../spinoso-string", features = ["always-nul-terminated-c-string-compat"] }
 spinoso-symbol = { version = "0.3.0", path = "../spinoso-symbol" }
 spinoso-time = { version = "0.7.0", path = "../spinoso-time", features = ["tzrs"], default-features = false, optional = true }
+strftime-ruby = { version = "1.0.0", optional = true }
 
 [dev-dependencies]
 quickcheck = { version = "1.0.3", default-features = false }
@@ -69,7 +70,7 @@ core-math-full = ["core-math", "spinoso-math?/full"]
 core-random = ["dep:spinoso-random"]
 core-regexp = ["dep:spinoso-regexp"]
 core-regexp-oniguruma = ["core-regexp", "spinoso-regexp?/oniguruma", "dep:onig"]
-core-time = ["dep:spinoso-time"]
+core-time = ["dep:spinoso-time", "dep:strftime-ruby"]
 
 load-path-native-file-system-loader = ["artichoke-load-path/native-file-system-loader"]
 load-path-rubylib-native-file-system-loader = ["load-path-native-file-system-loader", "artichoke-load-path/rubylib-native-file-system-loader"]
--- a/artichoke-backend/src/extn/core/time/trampoline.rs
+++ b/artichoke-backend/src/extn/core/time/trampoline.rs
@@ -1,6 +1,8 @@
 //! Glue between mruby FFI and `Time` Rust implementation.
 
-use spinoso_time::strftime::Error::InvalidFormatString;
+use spinoso_time::strftime::{
+    Error::FormattedStringTooLarge, Error::InvalidFormatString, Error::WriteZero, ASCTIME_FORMAT_STRING,
+};
 
 use crate::convert::implicitly_convert_to_int;
 use crate::convert::to_str;
@@ -488,11 +490,42 @@ fn strftime_with_encoding(
         // ```
         //
         // Note: The errors which are re-thrown as RuntimeError include (but is
-        // not limited to: `InvalidTime`, `FormattedStringTooLarge`,
-        // `WriteZero`, `FmtError(Error)`, `OutOfMemory(TryReserveError)`
+        // not limited to: `InvalidTime`, `FmtError(Error)`,
+        // `OutOfMemory(TryReserveError)`
+        #[allow(clippy::match_same_arms)]
         match e {
             InvalidFormatString => {
                 let mut message = br#"invalid format: "#.to_vec();
+                message.extend_from_slice(format);
+                Error::from(ArgumentError::from(message))
+            }
+            FormattedStringTooLarge => {
+                // TODO: This should be an `Errno::ERANGE` not an ArgumentError
+                //
+                // ```console
+                // [3.1.2] > Time.now.strftime "%4718593m"
+                // (irb):28:in `strftime': Result too large - %4718593m (Errno::ERANGE)
+                //      from (irb):28:in `<main>'
+                //      from /usr/local/var/rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
+                //      from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `load'
+                //      from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `<main>'
+                // ```
+                let mut message = br#"Result too large - "#.to_vec();
+                message.extend_from_slice(format);
+                Error::from(ArgumentError::from(message))
+            }
+            WriteZero => {
+                // TODO: This should be an `Errno::ERANGE` not an ArgumentError
+                //
+                // ```console
+                // [3.1.2] > Time.now.strftime "%2147483647m"
+                // (irb):28:in `strftime': Result too large - %2147483647m (Errno::ERANGE)
+                //      from (irb):29:in `<main>'
+                //      from /usr/local/var/rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
+                //      from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `load'
+                //      from /usr/local/var/rbenv/versions/3.1.2/bin/irb:25:in `<main>'
+                // ```
+                let mut message = br#"Result too large - "#.to_vec();
                 message.extend_from_slice(format);
                 Error::from(ArgumentError::from(message))
             }
--- a/artichoke-backend/src/extn/core/time/trampoline.rs
+++ b/artichoke-backend/src/extn/core/time/trampoline.rs
@@ -1,7 +1,8 @@
 //! Glue between mruby FFI and `Time` Rust implementation.
 
 use spinoso_time::strftime::{
-    Error::FormattedStringTooLarge, Error::InvalidFormatString, Error::WriteZero, ASCTIME_FORMAT_STRING,
+    Error::{FormattedStringTooLarge, InvalidFormatString, WriteZero},
+    ASCTIME_FORMAT_STRING,
 };
 
 use crate::convert::implicitly_convert_to_int;
--- a/artichoke-backend/src/extn/core/time/trampoline.rs
+++ b/artichoke-backend/src/extn/core/time/trampoline.rs
@@ -226,16 +226,7 @@ pub fn as_utc(interp: &mut Artichoke, mut time: Value) -> Result<Value, Error> {
 // Inspect
 
 pub fn asctime(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
-    // %c is the same as `%a %b %e %T %Y`, where:
-    // - `%a` is a capitalized abbreviated day of week name, e.g. `Sun`.
-    // - `%b` is a capitalized abbreviated month name, e.g. `Sep`.
-    // - `%e` is a a space padded right align day of month, e.g. ` 6`.
-    // - `%T` is a composite field which represents `%H:%M:%S` which represents
-    //   the 24 hour time padded with zeros, e.g. `01:02:03`.
-    // - `%Y` is the current year, padded to 4 digits, e.g. `0512`.
-    let format = "%c";
-
-    strftime_with_encoding(interp, time, format.as_bytes(), Encoding::Utf8)
+    strftime_with_encoding(interp, time, ASCTIME_FORMAT_STRING.as_bytes(), Encoding::Utf8)
 }
 
 pub fn to_string(interp: &mut Artichoke, mut time: Value) -> Result<Value, Error> {
--- a/artichoke-backend/src/extn/core/time/trampoline.rs
+++ b/artichoke-backend/src/extn/core/time/trampoline.rs
@@ -3,6 +3,8 @@
 use spinoso_time::strftime::Error::InvalidFormatString;
 
 use crate::convert::implicitly_convert_to_int;
+use crate::convert::to_str;
+use crate::extn::core::string::{Encoding, String};
 use crate::extn::core::time::{subsec::Subsec, Offset, Time};
 use crate::extn::prelude::*;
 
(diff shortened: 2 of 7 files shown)


artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -472,10 +467,30 @@ pub fn subsec(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
...
+        //
+        // ```console
+        // [3.1.2]> Time.now.strftime("%")
+        // (irb):1:in `strftime': invalid format: % (ArgumentError)
+        //      from (irb):1:in `<main>'
+        //	    from /home/ben/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
+        //	    from /home/ben/.rbenv/versions/3.1.2/bin/irb:25:in `load'
+        //	    from /home/ben/.rbenv/versions/3.1.2/bin/irb:25:in `<main>'
+        // ```
+        if let strftime::Error::InvalidFormatString = e {
```
lopopolo: Can we restructure this as an "exhaustive" (I think the error enum is marked non exhaustive) match? I'd feel more comfortable with this squash into runtime error approach if we are explicitly enumerating the cases we are squashing (at the very least it means we don't have to open the docs to see).

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -472,10 +467,30 @@ pub fn subsec(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
...
+            let mut message = br#"invalid format: "#.to_vec();
+            message.extend_from_slice(format);
+            Error::from(ArgumentError::from(message))
+        } else {
+            Error::from(RuntimeError::with_message("Unexpected failure"))
+        }
+    })?;
+
+    interp.try_convert_mut(&bytes[..])
```
lopopolo: Couple of things:

When using the converters to string, passing the owned vec if possible saves an allocation.

I think we need to preserve the encoding of the format string in the output string which means we need to unbox the format as a spinoso string directly (after getting a value using the new `to_str` MRI converter?) and then allocate and box a spinoso string with the correct encoding.

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -472,10 +467,30 @@ pub fn subsec(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
 
 // Time format
 
-pub fn strftime(interp: &mut Artichoke, time: Value, format: Value) -> Result<Value, Error> {
-    let _ = interp;
-    let _ = time;
-    let _ = format;
-    // Requires a parser.
-    Err(NotImplementedError::new().into())
+pub fn strftime(interp: &mut Artichoke, mut time: Value, format: Value) -> Result<Value, Error> {
```
lopopolo: Do you think we can add `asctime` and `ctime` in the same PR using the route through strftime approach that `to_s` does?

lopopolo (changes requested):
I like the approach you've taken with mapping the strftime errors, direct dep seems fine. Some questions about encodings and I think we can knock out a couple of more APIs too.

How do the specs look after this change? 📊

lopopolo:
Re the out of range error, that error is probably directly set in C's `errno` from `strftime` in libc.

I think strftime-ruby returns a specific error for these cases, but artichoke doesn't have `Errno` family of exceptions yet. Let's handle the case specifically and mark it as a FIXME to get the right exception class in there. `RuntimeError` sounds fine for now.

lopopolo:
The `Errno` thing is yet another spot where MRI is a thin layer on top of POSIX APIs 😅

b-n:
Re: Spec changes. I realised that a lot of the `N` are not what I assumed before, and are actually tests that aren't passing. Lots of the spec tests are run with `Time#utc` etc which isn't implemented yet. I've got another branch in local WIP which is going to be #SuperFun. `args.rs` coming soon.

Re: Encoding. Yes very good calls all round. I was somewhat reluctant to use the String, not sure why, but you raise very valid points. I had a thought of using the `to_s` etc, but this worked so I stopped haha. I'll give it a crack. Just checking, those conversions still return a `Value` so I'm assuming we still just use the unboxing stuff to get the value out?

Re: POSIX wrapping. 😅

artichoke-backend/src/extn/core/time/trampoline.rs
b-n: I had a feeling I was missing some, yes will add!

artichoke-backend/src/extn/core/time/trampoline.rs
b-n: Yes good point, it would make it a lot more explicit (good thing). I'm imagine a match block with each enum value, is that what you had in mind too?

artichoke-backend/src/extn/core/time/trampoline.rs
b-n: This should all be achievable by constructing a String with particular encoding with the Vec right?

Specifically on boxing the owned Vec, I noticed there wasnt a `try_convert_mut` for `Vec<u8>`, thus this slice conversion. I assume in future we might even want to reduce and/stop usage of that kind of conversion right?

artichoke-backend/src/extn/core/time/trampoline.rs
lopopolo: yup that's what I had in mind. 👍

artichoke-backend/src/extn/core/time/trampoline.rs
lopopolo: there is a `TryConvertMut` for vec:

https://github.com/artichoke/artichoke/blob/b261606b1de84f7ce4670e7ddda53f5738cde6cb/artichoke-backend/src/convert/bytes.rs#L14-L22

But yes long term I'd love to move in the direction of always going through spinoso-string. Here's an example of some bugs we've had due to these encoding issues:

- fac36b9f45b036ef2535673a95064afcc2776f73
- https://github.com/artichoke/artichoke/pull/2089

lopopolo:
> Just checking, those conversions still return a `Value` so I'm assuming we still just use the unboxing stuff to get the value out?

yup that's right but we should be able to just `expect` on the unboxing since the MRI converters guarantee the value has the right type. I kinda like this because it removes some `transmute` to get the lifetimes to match up.

artichoke-backend/Cargo.toml
```diff
@@ -39,6 +39,7 @@ spinoso-securerandom = { version = "0.2.0", path = "../spinoso-securerandom", op
 spinoso-string = { version = "0.20.0", path = "../spinoso-string", features = ["always-nul-terminated-c-string-compat"] }
 spinoso-symbol = { version = "0.3.0", path = "../spinoso-symbol" }
 spinoso-time = { version = "0.7.0", path = "../spinoso-time", features = ["tzrs"], default-features = false, optional = true }
+strftime-ruby = { version = "1.0.0", optional = true }
```
lopopolo: `spinoso-time` actually re-exports `strftime-ruby` so I don't think we need to add this direct dependency.

https://artichoke.github.io/artichoke/spinoso_time/#reexports

artichoke-backend/src/extn/core/time/trampoline.rs
lopopolo: let's add a `use spinoso_time::strftime` at the top of this file to bring this into scope.

b-n: Some limitations with regards to the non-exhaustive list, however I've added a comment to cater for this. I could have ignored the clippy error instead, but will resolve this for now.

artichoke-backend/Cargo.toml
b-n: Ah yes, missed this, easy fix!

artichoke-backend/src/extn/core/time/trampoline.rs
b-n: I'm mostly certain that this is fixed now. I'll resolve this for now, but I'll put a question in a place which I'm not 100% sure about it's safety.

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -471,11 +475,49 @@ pub fn subsec(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
...
+            InvalidFormatString => {
+                let mut message = br#"invalid format: "#.to_vec();
+                message.extend_from_slice(format);
+                Error::from(ArgumentError::from(message))
+            }
+            _ => Error::from(RuntimeError::with_message("Unexpected failure")),
+        }
+    })?;
+
+    let result = String::with_bytes_and_encoding(bytes, encoding);
```
b-n, line 537: For `[as]ctime` and `to_s|inspect`, it's always assumed it'll be valid utf-8.

For `Time#strftime` this might be a little bit more interesting. It's definitely making the assumption that it can enforce the coding of the provided format string. I'm mostly certain this should be a fair assumption. e.g. I assume it's impossible to create a RString with Utf8 encoding with invalid Utf8 chars right? And also, it's assuming that the `strftime-ruby` crate will never provide non ASCII chars back in place of the placeholders (This would do some weird things if the passed in format string was ASCII for example).

Do you think there should be any additional safe guards?

lopopolo: It is true that `strftime-ruby` will only write ASCII bytes to for the various format specifiers.

> e.g. I assume it's impossible to create a RString with Utf8 encoding with invalid Utf8 chars right?

This is not true. `RString`s can be constructed with arbitrary bytes and any encoding. The encoding is only conventional.

It looks like when the format string is valid per its encoding, the format string's encoding is used:

```
[3.1.2] > x = "xyz %c".b
=> "xyz %c"
[3.1.2] > x.encoding
=> #<Encoding:ASCII-8BIT>
[3.1.2] > Time.now.strftime x
=> "xyz Wed Sep 21 08:57:52 2022"
[3.1.2] > Time.now.strftime(x).encoding
=> #<Encoding:ASCII-8BIT>
[3.1.2] > z = "xyz \xFF %c".b
=> "xyz \xFF %c"
[3.1.2] > z.encoding
=> #<Encoding:ASCII-8BIT>
[3.1.2] > Time.now.strftime z
=> "xyz \xFF Wed Sep 21 09:00:36 2022"
[3.1.2] > Time.now.strftime(z).encoding
=> #<Encoding:ASCII-8BIT>
```

Additionally, when the format string is not valid per its encoding, invalid bytes _and_ the encoding are preserved:

```
[3.1.2] > y = "abc \xFF %c"
=> "abc \xFF %c"
[3.1.2] > y.encoding
=> #<Encoding:UTF-8>
[3.1.2] > y.valid_encoding?
=> false
[3.1.2] > Time.now.strftime y
=> "abc \xFF Wed Sep 21 09:01:43 2022"
[3.1.2] > Time.now.strftime(y).encoding
=> #<Encoding:UTF-8>
```

lopopolo: It'd be great for us to add some functional tests for the above ^. Do we have a `time_test.rb` yet?

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -224,9 +224,18 @@ pub fn as_utc(interp: &mut Artichoke, mut time: Value) -> Result<Value, Error> {
...
-    let _ = time;
-    Err(NotImplementedError::new().into())
+    // %c is the same as `%a %b %e %T %Y`, where:
+    // - `%a` is a capitalized abbreviated day of week name, e.g. `Sun`.
+    // - `%b` is a capitalized abbreviated month name, e.g. `Sep`.
+    // - `%e` is a a space padded right align day of month, e.g. ` 6`.
+    // - `%T` is a composite field which represents `%H:%M:%S` which represents
+    //   the 24 hour time padded with zeros, e.g. `01:02:03`.
+    // - `%Y` is the current year, padded to 4 digits, e.g. `0512`.
+    let format = "%c";
```
lopopolo: there is a constant for this in `strftime`: https://artichoke.github.io/artichoke/strftime/constant.ASCTIME_FORMAT_STRING.html

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -508,5 +509,15 @@ pub fn strftime(interp: &mut Artichoke, mut time: Value, format: Value) -> Resul
...
+
+    String::alloc_value(result, interp)
+}
+
+pub fn strftime(interp: &mut Artichoke, time: Value, format: Value) -> Result<Value, Error> {
+    let mut format = to_str(interp, format)?;
+
+    let format = unsafe { String::unbox_from_value(&mut format, interp)? };
+
+    strftime_with_encoding(interp, time, &format, format.encoding())
```
lopopolo, line 547: this is awesome. I love this refactor.

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -471,11 +475,49 @@ pub fn subsec(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
...
+        //
+        // Note: The errors which are re-thrown as RuntimeError include (but is
+        // not limited to: `InvalidTime`, `FormattedStringTooLarge`,
+        // `WriteZero`, `FmtError(Error)`, `OutOfMemory(TryReserveError)`
+        match e {
+            InvalidFormatString => {
+                let mut message = br#"invalid format: "#.to_vec();
+                message.extend_from_slice(format);
+                Error::from(ArgumentError::from(message))
+            }
```
lopopolo, line 502: `FormattedStringTooLarge` is a dedicated error:

```
[3.1.2] > Time.now.strftime "%4718593m"
(irb):28:in `strftime': Result too large - %4718593m (Errno::ERANGE)
        from (irb):28:in `<main>'
```

We don't have `Errno::Range`, but can we set the right message on an `ArgumentError` and add a TODO to fix the exception class?

lopopolo: `WriteZero` also shares this failure mode:

https://github.com/artichoke/strftime-ruby/blob/ccc82452080efa815e31dfe80149e16fd847e9e3/src/tests/format.rs#L875-L876

```
(irb):29:in `strftime': Result too large - %2147483647m (Errno::ERANGE)
        from (irb):29:in `<main>'
```

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -471,11 +475,49 @@ pub fn subsec(interp: &mut Artichoke, time: Value) -> Result<Value, Error> {
...
+        // (irb):1:in `strftime': invalid format: % (ArgumentError)
+        //      from (irb):1:in `<main>'
+        //	    from /home/ben/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
+        //	    from /home/ben/.rbenv/versions/3.1.2/bin/irb:25:in `load'
+        //	    from /home/ben/.rbenv/versions/3.1.2/bin/irb:25:in `<main>'
+        // ```
+        //
+        // Note: The errors which are re-thrown as RuntimeError include (but is
+        // not limited to: `InvalidTime`, `FormattedStringTooLarge`,
+        // `WriteZero`, `FmtError(Error)`, `OutOfMemory(TryReserveError)`
```
lopopolo: Technically we can handle the out of memory variant too, but I think it's ok to skip this one:

https://github.com/artichoke/artichoke/blob/589da905df31470ca6abfa17a33b645d46670f03/artichoke-backend/src/extn/core/string/ffi.rs#L222-L228

lopopolo (approved):
This looks good to me. I'm happy to have this merged as is, but I think I'd like it if we could follow up on these two things:

- Add some functional tests for `Time` to test the `strftime` encoding of result matches encoding of format string invariant.
- Handle "result too large" errors from `strftime-ruby`.

artichoke-backend/src/extn/core/time/trampoline.rs
b-n: Ahhh yes, I see what you mean here. We could add a `time_test.rb`, however it appears there still isn't support for encodings on String yet, which makes it kind of hard to write tests for :-(.

```
artichoke 0.1.0-pre.0 (2022-09-14 revision 6305) [x86_64-unknown-linux-gnu]
[rustc 1.63.0 (4b91a6ea7 2022-08-08) on x86_64-unknown-linux-gnu]
>>> "a".b.encoding
=> #<Encoding:UTF-8>
>>> "a".force_encoding(Encoding::ASCII).encoding
=> #<Encoding:UTF-8>
```

In part due to this: https://github.com/artichoke/artichoke/blob/trunk/artichoke-backend/src/extn/core/string/string.rb#L606-L610

These might be some quick wins - I'll raise some issues for it.

lopopolo: At least we can do UTF-8 and binary (`"abc".b`).

Ahh I see though, `String#encoding` isn't implemented. Maybe we can fake it by using an emoji in the format string and asserting on the length? For binary strings length will be bytesize.

artichoke-backend/src/extn/core/time/trampoline.rs
b-n: Yeah, I had a feeling that was going to be a harder one. That comment leads me to believe that we would be in much bigger trouble if we actually got that error.

b-n:
@lopopolo How do you feel about: 0c4bd57 ?

Note, the length checks (as you mentioned) are because of this below (`String#==` does not entirely match MRI)

```
>>> f = "%c \u{1F600}"
=> "%c 😀"
>>> f = "%c \u{1F600}".b
=> "%c \xF0\x9F\x98\x80"
>>> t.strftime(f)
=> "Thu Jan  1 00:16:40 1970 \xF0\x9F\x98\x80"
>>> t.strftime(f) == 'Thu Jan  1 00:16:40 1970 😀'
=> true
```

This is because in `airb`:

```
>>> "\u{1F600}".b == "\u{1F600}"
=> true
>>> "A".b == "A"
=> true
```

Where as MRI:

```
[3.1.2]> "\u{1F600}".b == "\u{1F600}"
=> false
[3.1.2]> "A".b == "A"
=> true
```

e.g. `String#==` is does some "fun" things :-(

lopopolo (approved):
This looks good to go for me. Thanks for adding the functional tests.

PR #2183, opened 2022-09-20
Refactor String#chop to use String#ends_with directly

### LINKED ISSUE #1721: Push down `String#ends_with?` implementation to encoded `String` variants
maybe a new `ends_with()` fn on the encoded implementations would make this even nicer

_Originally posted by @b-n in https://github.com/artichoke/artichoke/pull/1678#r816221491_

choznerol:
Close #1721

The original discussion in https://github.com/artichoke/artichoke/pull/1678#discussion_r816221491 seems like `String#ends_with` was not exist back then, but it actually [does exist here](https://github.com/b-n/artichoke/blob/b5ca546d817081cc3e1087322b3f98317ac7f78f/spinoso-string/src/enc/mod.rs#L503-L511). Please let me know if I misunderstand #1721.

### DIFF
--- a/spinoso-string/src/lib.rs
+++ b/spinoso-string/src/lib.rs
@@ -1544,7 +1544,7 @@ impl String {
         if self.is_empty() {
             return false;
         }
-        let bytes_to_remove = if self.inner.as_slice().ends_with(b"\r\n") {
+        let bytes_to_remove = if self.inner.ends_with(b"\r\n") {
             2
         } else if let Encoding::Utf8 = self.encoding() {
             let (ch, size) = bstr::decode_last_utf8(&self.as_slice());


spinoso-string/src/lib.rs
```diff
@@ -1544,7 +1544,7 @@ impl String {
         if self.is_empty() {
             return false;
         }
-        let bytes_to_remove = if self.inner.as_slice().ends_with(b"\r\n") {
+        let bytes_to_remove = if self.inner.ends_with(b"\r\n") {
```
choznerol, line 1547: I did not add additional tests because existing doc tests already cover this 'if' branch. If I change the following the doc test of `String#chop` would fail.
```diff
-        let bytes_to_remove = if self.inner.ends_with(b"\r\n") {
+        let bytes_to_remove = if self.inner.ends_with(b"abc\r\n") {
```

lopopolo (approved):
thank you!

spinoso-string/src/lib.rs
lopopolo: awesome thanks for verifying.

PR #2187, opened 2022-09-21
Update vendored reference ruby implementation

b-n:
This includes the 2 patches to `lib/uri/http.rb` and `lib/uri/generic.rb`


lopopolo:
Fork updated here: https://github.com/artichoke/ruby/tree/artichoke-vendor

lopopolo (approved):

PR #2191, opened 2022-09-24
Remove an allocation when computing a fixed time zone offset

lopopolo:
All valid timezone fixed time zone offsets result in an offset string of the format +HHMM or -HHMM. Both forms are exactly 5 bytes. We can use a stack-allocated buffer to write the bytes into and avoid allocating a `String`.

Exhaustive tests are added to assert the invariant that the buffer is exactly the right size.

### DIFF
--- a/spinoso-time/src/time/tzrs/offset.rs
+++ b/spinoso-time/src/time/tzrs/offset.rs
@@ -1,3 +1,4 @@
+use std::io::{self, Write as _};
 use std::slice;
 use std::str;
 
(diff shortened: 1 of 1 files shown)


spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -64,15 +65,14 @@ fn local_time_zone() -> TimeZoneRef<'static> {
...
-fn offset_hhmm_from_seconds(seconds: i32) -> String {
+fn offset_hhmm_from_seconds(seconds: i32, buf: &mut [u8; 5]) -> io::Result<()> {
     let flag = if seconds < 0 { '-' } else { '+' };
     let minutes = seconds.abs() / 60;
 
     let offset_hours = minutes / 60;
     let offset_minutes = minutes - (offset_hours * 60);
 
-    format!("{}{:0>2}{:0>2}", flag, offset_hours, offset_minutes)
+    write!(buf.as_mut_slice(), "{flag}{offset_hours:0>2}{offset_minutes:0>2}")
```
b-n, line 75: Ahhhh, this is pretty cool. I see why you're writing to the passed in buffer vs. attempting to return a `[u8;5]`. I assume the `io::Result` would give an error on overflow of the buffer, so that's how you're able to safely write to the 5 byte buffer sort of thing. Nice.

b-n (approved):
Clever! It took me a moment to realise why you'd pass in a buffer instead of just returning a `[u8;5]`, but it makes total sense. Nice.

PR #2196, opened 2022-09-29
Update `artichoke` CLI to clap v4

lopopolo:
https://epage.github.io/blog/2022/09/clap4/

`--help` output is changed, so update the ui-test snapshots.

One of the neat new things in clap v4 is the ability to add a test to assert that the `clap::Command` is configured correctly. When a package has both a `lib.rs` and a binary target, I think `#[test]`s don't run in the `main.rs`, so this PR also takes the chance to migrate all of the `artichoke` clap CLI setup to a new `artichoke::ruby::cli` module.

### DIFF
--- a/src/ruby/cli.rs
+++ b/src/ruby/cli.rs
@@ -2,6 +2,7 @@
 
 use std::env;
 use std::ffi::OsString;
+use std::iter;
 use std::path::PathBuf;
 use std::process;
 
@@ -56,8 +57,7 @@ pub fn parse_args() -> Args {
         args = args.with_commands(commands);
         if let Some(first_arg) = matches.get_one::<PathBuf>("programfile").cloned() {
             if let Some(argv) = matches.get_many::<OsString>("arguments") {
-                let ruby_program_argv = [OsString::from(first_arg)]
-                    .into_iter()
+                let ruby_program_argv = iter::once(OsString::from(first_arg))
                     .chain(argv.map(Clone::clone))
                     .collect::<Vec<_>>();
                 args = args.with_argv(ruby_program_argv);
--- a/src/ruby/cli.rs
+++ b/src/ruby/cli.rs
@@ -1,0 +1,153 @@
+//! Command line interface parser for the `ruby` binary.
+
+use std::env;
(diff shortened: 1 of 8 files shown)


src/ruby/cli.rs
```diff
@@ -0,0 +1,153 @@
...
+    };
+    if err.use_stderr() {
+        return Err(err.into());
+    }
+    // Explicitly ignore any error returned by write!. The most likely error
+    // at this point is a broken pipe error, in which case, we want to ignore
+    // it and exit quietly.
+    //
+    // (This is the point of this helper function. clap's functionality for
+    // doing this will panic on a broken pipe error.)
```
epage: I thought the broken pipe issue has been fixed for a while...

lopopolo: I think it has been, but this code has worked the whole time and doesn't have that much maintenance cost for me to carry. I also prefer having visibility into all calls to `process::exit` instead of having them happen in a library if I can help it.

lopopolo: But I can definitely clean up this comment since the clap side of things has been fixed.

lopopolo: I just pushed a commit that removes this parenthetical comment. Thanks for catching that!

epage: You can also use `err.print()` so you can control the exiting but get clap to do the printing, including colored output.  If you want more control over that and are fine with ANSI escape codes, you can now also call `err.render().ansi()`.

lopopolo: Oh that's awesome!

## Before

Screen Shot 2022-09-28 at 8 23 53 PM

## After

Screen Shot 2022-09-28 at 8 24 06 PM

lopopolo:
Thanks for the code review @epage 🎉 ⚙️

src/ruby/cli.rs
epage: Huh, noticed a couple issues in that.  Addressed in https://github.com/clap-rs/clap/pull/4287 which should be released soon

lopopolo: Glad to be of help!

PR #2217, opened 2022-10-15
Use a match guard clause for unwrapping the option to avoid right drift

b-n:
This is more of a proposal for long blocks like what is used in the `try_convert_mut` in the `time/subsec.rs` class. I do like the minimized overhead of guard clauses (seem to be very idiomatic in Ruby in my experience), and in this case I think it's worth it. One day we might get [`let-else`](https://github.com/rust-lang/rfcs/blob/master/text/3137-let-else.md) which would make this syntax even smaller.

This change looks big, but it's purely indenting. Also happy to trash this PR since it's pure style preference :smile:


lopopolo (approved):
adding `?w=1` to the PR files view made this look pretty simple. Thanks for laying that out in the PR description. I like this a lot.

b-n:
Hah, TIL. Also, found this option for making it more permanent: 
Screenshot from 2022-10-16 09-08-15

PR #2218, opened 2022-10-15
Ensure `Time#at` only takes symbol based subsec_types

### LINKED ISSUE #2095: Raise `ArgumentError` when given non-`Symbol` argument as time unit in `Time::at`
if we fail to unpack as a symbol, we need to return an `ArgumentError` instead of the `TypeError` this unboxing will return:

```
[3.1.2] > Time.at(12, 15000, "usec")
<internal:timev>:274:in `at': unexpected unit: usec (ArgumentError)
        from (irb):1:in `<main>'
[3.1.2] > Time.at(12, 15000, Object.new)
<internal:timev>:274:in `at': unexpected unit: #<Object:0x000000010e0e7b80> (ArgumentError)
        from (irb):2:in `<main>'
```

_Originally posted by @lopopolo in https://github.com/artichoke/artichoke/pull/1956#discussion_r950889522_

b-n:
Fixes #2095 

This also implements the same match guard clauses as seen in #2217. Really easy to change to the other way.

I'm not entirely sure whether it makes sense to test that this fails, but the tests is there at least (e.g. I threw a test for `Object.new` which should be enough, so I think testing for this situation below isn't entirely necessary)

```console
[3.1.2] > class A; def to_sym; :usec; end; end
=> :to_sym
[3.1.2] > Time.at(1, 1, A.new)
<internal:timev>:274:in `at': unexpected unit: #<A:0x00007f66207eb430> (ArgumentError)
	from (irb):29:in `<main>'                                                          
	from /home/ben/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
	from /home/ben/.rbenv/versions/3.1.2/bin/irb:25:in `load'                          
	from /home/ben/.rbenv/versions/3.1.2/bin/irb:25:in `<main>'  
```

### DIFF
--- a/artichoke-backend/src/extn/core/time/subsec.rs
+++ b/artichoke-backend/src/extn/core/time/subsec.rs
@@ -44,20 +44,21 @@ impl TryConvertMut<Option<Value>, SubsecMultiplier> for Artichoke {
     type Error = Error;
 
     fn try_convert_mut(&mut self, subsec_type: Option<Value>) -> Result<SubsecMultiplier, Self::Error> {
-        if let Some(mut subsec_type) = subsec_type {
-            let subsec_symbol = unsafe { Symbol::unbox_from_value(&mut subsec_type, self)? }.bytes(self);
-            match subsec_symbol {
-                b"milliseconds" => Ok(SubsecMultiplier::Millis),
-                b"usec" => Ok(SubsecMultiplier::Micros),
-                b"nsec" => Ok(SubsecMultiplier::Nanos),
-                _ => {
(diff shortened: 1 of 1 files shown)


lopopolo (approved):
Awesome! And thanks for adding the tests.

b-n:
I was pretty happy I found [`value.inspect(&mut interp)`](https://github.com/artichoke/artichoke/blob/trunk/artichoke-backend/src/value.rs#L200). Makes the "receieved x" argument errors super easy to handle.

PR #2220, opened 2022-10-16
Add Rust implementation of start_with? and end_with?

### LINKED ISSUE #2186: Implement `String#end_with?` and `String#start_with?` in Rust
These two functions are written in Ruby. If moved to Rust, they can be implemented with fewer allocations.

It would also allow gating the `Regexp` paths behind a feature flag.

Be sure to remove the pure Ruby variants.

xulaus:
Adds Rust implementations for `String#start_with?` and `String#end_with?`

Also corrects some differences to MRI 3.1.2
 - `start_with?` clears regexp globals when it fails to match
 - `end_with?` doesn't understand regexp matching

Closes #2186

### DIFF
--- a/artichoke-backend/src/extn/core/string/mod.rs
+++ b/artichoke-backend/src/extn/core/string/mod.rs
@@ -179,6 +179,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "core-regexp")]
     fn start_with_regex() {
         let mut interp = interpreter();
         // Test that regexp matching using start_with? clear the relevant regexp globals
--- a/artichoke-backend/src/extn/core/string/mod.rs
+++ b/artichoke-backend/src/extn/core/string/mod.rs
@@ -183,11 +183,11 @@ mod tests {
         let mut interp = interpreter();
         // Test that regexp matching using start_with? clear the relevant regexp globals
         // This is not tested in the vendored MRI version hence why it is tested here
-        let test = "
-            raise 'start_with? gives incorrect result' unless 'abcd test-123'.start_with?(/test-(\\d+)/) == false;
+        let test = r#"
+            raise 'start_with? gives incorrect result' unless 'abcd test-123'.start_with?(/test-(\d+)/) == false;
             raise 'start_with? should clear Regexp.last_match' unless Regexp.last_match == nil
             raise 'start_with? should clear $1' unless $1 == nil
-        ";
+        "#;
         let result = interp.eval(test.as_bytes());
         unwrap_or_panic_with_backtrace(&mut interp, SUBJECT, result);
     }
--- a/artichoke-backend/src/extn/core/string/mod.rs
+++ b/artichoke-backend/src/extn/core/string/mod.rs
@@ -177,4 +177,18 @@ mod tests {
         let result = interp.eval(test.as_bytes());
         unwrap_or_panic_with_backtrace(&mut interp, SUBJECT, result);
     }
+
+    #[test]
+    fn start_with_regex() {
+        let mut interp = interpreter();
+        // Test that regexp matching using start_with? clear the relevant regexp globals
+        // This is not tested in the vendored MRI version hence why it is tested here
+        let test = "
+            raise 'start_with? gives incorrect result' unless 'abcd test-123'.start_with?(/test-(\\d+)/) == false;
+            raise 'start_with? should clear Regexp.last_match' unless Regexp.last_match == nil
+            raise 'start_with? should clear $1' unless $1 == nil
+        ";
+        let result = interp.eval(test.as_bytes());
+        unwrap_or_panic_with_backtrace(&mut interp, SUBJECT, result);
+    }
 }
--- a/artichoke-backend/src/extn/core/string/mruby.rs
+++ b/artichoke-backend/src/extn/core/string/mruby.rs
@@ -448,7 +448,7 @@ unsafe extern "C" fn string_end_with(mrb: *mut sys::mrb_state, slf: sys::mrb_val
     let suffixes = suffixes.iter().map(|&other| Value::from(other));
     let result = trampoline::end_with(&mut guard, value, suffixes);
     match result {
-        Ok(result) => return result.inner(),
+        Ok(result) => result.inner(),
         Err(exception) => error::raise(guard, exception),
     }
 }
@@ -682,7 +682,7 @@ unsafe extern "C" fn string_start_with(mrb: *mut sys::mrb_state, slf: sys::mrb_v
     let prefixes = prefixes.iter().map(|&other| Value::from(other));
     let result = trampoline::start_with(&mut guard, value, prefixes);
     match result {
-        Ok(result) => return result.inner(),
+        Ok(result) => result.inner(),
         Err(exception) => error::raise(guard, exception),
     }
 }
--- a/artichoke-backend/src/extn/core/string/mruby.rs
+++ b/artichoke-backend/src/extn/core/string/mruby.rs
@@ -442,18 +442,14 @@ unsafe extern "C" fn string_empty(mrb: *mut sys::mrb_state, slf: sys::mrb_value)
 }
 
 unsafe extern "C" fn string_end_with(mrb: *mut sys::mrb_state, slf: sys::mrb_value) -> sys::mrb_value {
-    let prefixes = mrb_get_args!(mrb, *args);
+    let suffixes = mrb_get_args!(mrb, *args);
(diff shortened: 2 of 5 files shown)


artichoke-backend/src/extn/core/string/mod.rs
```diff
@@ -177,4 +177,18 @@ mod tests {
         let result = interp.eval(test.as_bytes());
         unwrap_or_panic_with_backtrace(&mut interp, SUBJECT, result);
     }
+
+    #[test]
+    fn start_with_regex() {
```
lopopolo, line 183: this is a great test. thanks for adding it!

artichoke-backend/src/extn/core/string/mod.rs
```diff
@@ -177,4 +177,18 @@ mod tests {
...
+        let mut interp = interpreter();
+        // Test that regexp matching using start_with? clear the relevant regexp globals
+        // This is not tested in the vendored MRI version hence why it is tested here
+        let test = "
```
lopopolo: I think this could be made a bit more clear if we used a "raw string" which would avoid having to escape the `\d` in the regexp literal.

https://rahul-thakoor.github.io/rust-raw-string-literals/

artichoke-backend/src/extn/core/string/mruby.rs
```diff
@@ -439,6 +441,21 @@ unsafe extern "C" fn string_empty(mrb: *mut sys::mrb_state, slf: sys::mrb_value)
...
+    let value = Value::from(slf);
+
+    for prefix in prefixes.iter().map(|&other| Value::from(other)) {
+        match trampoline::end_with(&mut guard, value, prefix) {
+            Ok(true) => return guard.convert(true).inner(),
+            Ok(false) => {}
+            Err(exception) => error::raise(guard, exception),
+        }
+    }
+    guard.convert(false).inner()
```
lopopolo: I'd prefer if we moved the loop into the trampoline. You can make the trampoline take a generic iterator. Can you refactor this to use the same pattern as `Kernel#puts`?


https://github.com/artichoke/artichoke/blob/dac4c8c9280a00350ae2421628b6f1b1710171b7/artichoke-backend/src/extn/core/kernel/trampoline.rs#L129-L132

https://github.com/artichoke/artichoke/blob/dac4c8c9280a00350ae2421628b6f1b1710171b7/artichoke-backend/src/extn/core/kernel/mruby.rs#L93-L98

artichoke-backend/src/extn/core/string/mruby.rs
```diff
@@ -661,6 +678,21 @@ unsafe extern "C" fn string_split(mrb: *mut sys::mrb_state, slf: sys::mrb_value)
...
+    let value = Value::from(slf);
+
+    for prefix in prefixes.iter().map(|&other| Value::from(other)) {
+        match trampoline::start_with(&mut guard, value, prefix) {
+            Ok(true) => return guard.convert(true).inner(),
+            Ok(false) => {}
+            Err(exception) => error::raise(guard, exception),
+        }
+    }
+    guard.convert(false).inner()
```
lopopolo: Ditto, let's move the iterator into the trampoline so this mruby landing pad looks the same as all of the rest of them.

artichoke-backend/src/extn/core/string/string.rb
```diff
@@ -574,25 +574,10 @@ def encoding
...
-        return true if suffix.empty?
-        return true if self[-suffix.length..-1] == suffix
-      end
-    end
-    false
-  end
+  #
+  # NOTE: Implemented in native code.
+  #
+  # def end_with?; end
```
lopopolo, line 580: 😄 awesome thanks for picking up on this pattern.

artichoke-backend/src/extn/core/string/string.rb
```diff
@@ -574,25 +574,10 @@ def encoding
   end
 
   # https://ruby-doc.org/core-3.0.2/String.html#method-i-end_with-3F
-  def end_with?(*suffixes)
-    suffixes.each do |suffix|
-      case suffix
-      when String
-        return true if suffix.empty?
-        return true if self[-suffix.length..-1] == suffix
-      when Regexp
```
lopopolo, line 583: Ooo thank you. Can you add a test to the string functional tests to assert that `String#end_with?` raises `ArgumentError` when given a `Regexp`? You can run this whole file with MRI by invoking it `ruby path/to/string_test.rb` to ensure we have parity.

Maybe one case with a single regexp, one case with a regexp in the set of args where none match, and one case with a regexp in the set of args where an arg earlier in the sequence matches?

https://github.com/artichoke/artichoke/blob/dac4c8c9280a00350ae2421628b6f1b1710171b7/artichoke-backend/src/extn/core/string/string_test.rb#L17-L23

lopopolo (changes requested):

artichoke-backend/src/extn/core/string/string.rb
xulaus: I've added the test you've suggested to that file. It's made me doubt putting the `start_with?` regexp test in rust though. Should that new test have been in the `string_test.rb` too, or is it fine where it is?

artichoke-backend/src/extn/core/string/mod.rs
lopopolo: This is fine to keep in Rust

artichoke-backend/src/extn/core/string/mruby.rs
```diff
@@ -439,6 +441,17 @@ unsafe extern "C" fn string_empty(mrb: *mut sys::mrb_state, slf: sys::mrb_value)
     }
 }
 
+unsafe extern "C" fn string_end_with(mrb: *mut sys::mrb_state, slf: sys::mrb_value) -> sys::mrb_value {
+    let suffixes = mrb_get_args!(mrb, *args);
+    unwrap_interpreter!(mrb, to => guard);
+    let value = Value::from(slf);
+    let suffixes = suffixes.iter().map(|&other| Value::from(other));
+    match trampoline::end_with(&mut guard, value, suffixes) {
```
lopopolo: I'm not sure what putting this expression in the match means for lifetimes and things. Because the Err branch unwinds with a longjmp, I'd feel more comfortable if the result of the trampoline was assigned to a local binding like it is in the other mruby entry points.

artichoke-backend/src/extn/core/string/mruby.rs
```diff
@@ -661,6 +674,17 @@ unsafe extern "C" fn string_split(mrb: *mut sys::mrb_state, slf: sys::mrb_value)
     }
 }
 
+unsafe extern "C" fn string_start_with(mrb: *mut sys::mrb_state, slf: sys::mrb_value) -> sys::mrb_value {
+    let prefixes = mrb_get_args!(mrb, *args);
+    unwrap_interpreter!(mrb, to => guard);
+    let value = Value::from(slf);
+    let prefixes = prefixes.iter().map(|&other| Value::from(other));
+    match trampoline::start_with(&mut guard, value, prefixes) {
```
lopopolo: Ditto

artichoke-backend/src/extn/core/string/string.rb
lopopolo: I think it's fine where it is. It allows us to gate the test on the core-regexp feature.

artichoke-backend/src/extn/core/string/mod.rs
```diff
@@ -177,4 +177,18 @@ mod tests {
         let result = interp.eval(test.as_bytes());
         unwrap_or_panic_with_backtrace(&mut interp, SUBJECT, result);
     }
+
+    #[test]
```
lopopolo, line 181:
```suggestion
    #[test]
    #[cfg(feature = "core-regexp")]
```

This will ensure this test only runs if `Regexp` support is enabled.

lopopolo (changes requested):
This is almost there. Just a few small changes left I think.

artichoke-backend/src/extn/core/string/string_test.rb
```diff
@@ -182,4 +183,19 @@ def string_tr
   raise unless 'abcd'.tr('a-z', 'xxx') == 'xxxx'
 end
 
+def string_end_with
+  def assert_raise_type_error
```
lopopolo: RuboCop doesn't seem to like this being nested. Feel free to move this helper to the top level. It reads really nicely this way with the block

lopopolo (approved):
thank you!

Issue #2075, opened 2022-08-12
Extract `String#to_i` impl to `scolapasta-int-parse`


lopopolo:
Reuse the radix type. Add tests.

From what I can tell, ASCII whitespace is trimmed from front and back like `Kernel#Integer`.

```
[3.1.2] > "  \t\n\u000A\u000C            123abc                \t\n\u000A\u000C       ".to_i(16)
=> 1194684
[3.1.2] > "  \t\n\u000A\u000C            123abc                \t\n\u000A\u000C       ".to_i
=> 123
```

`0x`-style prefixes are allowed but only if they match the given base _or_ `x` is in the given base's alphabet. If the given base does not match the `0x`-style prefix, 0 is returned.

```
[3.1.2] > "0x123abc".to_i
=> 0
[3.1.2] > "0x123abc".to_i(16)
=> 1194684
[3.1.2] > "0x123abc".to_i(24)
=> 0
[3.1.2] > "0x123abc".to_i(36)
=> 71897795832
```

Numbers are parsed from the front and parsing stops when encountering an ASCII alphanumeric not in the given base's char set:

```
[3.1.2] > "123abc".to_i
=> 123
[3.1.2] > "123abc".to_i(16)
=> 1194684
```

PR #2224, opened 2022-10-29
Enable String#to_i spec

xulaus:
Relates to #1912

Currently `String#to_i` isn't in the enforced specs, and there deviations from MRI because of this.
This PR makes `to_i` spec compliant, asides from the cases where a `Bignum` is needed which now throw a `NotImplemented` error instead.

While doing this work I did attempt to use the `scolapasta-int-parse` implementation, however as `to_i` parses as much as possible before failure and scolapasta either passes or returns an error I would have had to make more changes that I felt comfortable with in that library to make the specs pass. This has lead to a duplication of logic, but there is already an issue to fix this (#2075)

### DIFF
--- a/artichoke-backend/src/extn/core/string/trampoline.rs
+++ b/artichoke-backend/src/extn/core/string/trampoline.rs
@@ -1508,6 +1508,14 @@ pub fn to_i(interp: &mut Artichoke, mut value: Value, base: Option<Value>) -> Re
         slice = &slice[1..];
     }
 
+    if slice.first() == Some(&b'_') {
+        return Ok(interp.convert(0));
+    }
+
+    let mut slice = std::borrow::Cow::from(slice);
+    if slice.find(&"_").is_some() {
+        slice.to_mut().retain(|&c| c != b'_');
+    }
     loop {
         use std::num::IntErrorKind;
         // Try to greedily parse the whole string as an int.
@@ -1522,8 +1530,14 @@ pub fn to_i(interp: &mut Artichoke, mut value: Value, base: Option<Value>) -> Re
             }
             _ => {
                 // if parsing failed, start discarding from the end one byte at a time.
-                let (_, head) = slice.split_last().unwrap();
-                slice = head;
+                match slice {
+                    std::borrow::Cow::Owned(ref mut data) => {
+                        data.pop();
+                    }
+                    std::borrow::Cow::Borrowed(data) => {
+                        slice = std::borrow::Cow::from(&data[..(data.len() - 1)]);
+                    }
+                }
             }
         }
     }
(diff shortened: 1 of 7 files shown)


artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -1478,54 +1478,113 @@ pub fn to_f(interp: &mut Artichoke, mut value: Value) -> Result<Value, Error> {
 }
 
 pub fn to_i(interp: &mut Artichoke, mut value: Value, base: Option<Value>) -> Result<Value, Error> {
-    fn try_parse(slice: &[u8], base: u32) -> Option<i64> {
-        let s = str::from_utf8(slice).ok()?;
-        i64::from_str_radix(s, base).ok()
+    let s = unsafe { super::String::unbox_from_value(&mut value, interp)? };
+    let mut slice = s.as_slice();
+    // ignore preceding whitespace
+    if let Some(start) = slice.iter().position(|c| !c.is_ascii_whitespace()) {
```
lopopolo: I believe this should use `posix-space` because `vertical-tab` is also stripped:

```console
[3.1.2] > "\x0B123".to_i
=> 123
[3.1.2] > "\x0B 123".to_i
=> 123
```

https://docs.rs/posix-space/latest/posix_space/fn.is_space.html

artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -1478,54 +1478,113 @@ pub fn to_f(interp: &mut Artichoke, mut value: Value) -> Result<Value, Error> {
...
+            // This can only be 2-36 inclusive in this branch, so unwrap is safe
+            u32::try_from(x).unwrap()
+        }
+    };
+
+    // Check string doesn't start with any special characters
+    //  '_' is invalid because they are stripped out elsewhere in the string, but cannot start a number
+    //  '+' and '-' invalid because we already have the sign prior to the prefix, but they are accepted
+    //  at the begining of a string by str_from_radix, and double sign doesn't make sense
+    if slice.first() == Some(&b'_') || slice.first() == Some(&b'+') || slice.first() == Some(&b'-') {
```
lopopolo: I like the way this can be cleaned up with `matches!`:

```suggestion
    if matches!(slice.first(), Some(&b'_') | Some(&b'+') | Some(&b'-')) {
```

artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -1478,54 +1478,113 @@ pub fn to_f(interp: &mut Artichoke, mut value: Value) -> Result<Value, Error> {
...
         return Ok(interp.convert(0));
     }
+
+    let mut slice = std::borrow::Cow::from(slice);
+    if slice.find(&"_").is_some() {
+        slice.to_mut().retain(|&c| c != b'_');
+    }
```
lopopolo: I don't think this short circuits when there are multiple consecutive underscores:

```console
[3.1.2] > "\x0B 1__23".to_i
=> 1
```

lopopolo (changes requested):
This is looking good so far and I'm excited to turn the specs on. I think there are some spots where this code still drifts from MRI.

For the cases that are not covered by the specs, can you add some integrations tests to:

https://github.com/artichoke/artichoke/blob/255dd13cf871e32d51658bc32fd302af2bc4652b/artichoke-backend/src/extn/core/string/string_test.rb#L5-L16

lopopolo (approved):
this looks great! thank you for the logic fixes, the new specs, and the new integration tests

PR #2265, opened 2022-11-08
Fix a daylight savings time breakage in spinoso-time doctests

lopopolo:
Doc test examples for `Time::to_local` failed when the current user's local time has a different UTC offset than the 2022-07-08T12:34:56 timestamp used in the test.

For me, located in `America/Los_Angeles`, this test started failing as of 2am local time on 2022-11-06.

Fix the test by forcibly constructing the local time with the same date time instead of using `Time::now`.


b-n:
Classic timezone problems :sweat:

PR #2308, opened 2022-12-09
Change PartialEq for EncodedString

### LINKED ISSUE #2193: `String#==` does not work the same way as MRI expects
Generated from https://github.com/artichoke/artichoke/pull/2182#issuecomment-1257154057

Ruby docs ref: https://ruby-doc.org/core-3.1.2/String.html#method-i-3D-3D. Relevant part:

> Returns true if object has the **same length** and content; as self; false otherwise.

And it looks like we're currently just comparing the internal `bytes` at present:
- [spinoso-string](https://github.com/artichoke/artichoke/blob/2db53033309c3e4f923fdea960d0884c4b47d1c8/spinoso-string/src/eq.rs)
- [artichoke-backend](https://github.com/artichoke/artichoke/blob/trunk/artichoke-backend/src/extn/core/string/trampoline.rs#L315-L333)

From MRI:

```console
[3.1.2]> "A" == "A"
=> true
[3.1.2]> "A".b == "A"
=> true
[3.1.2]> "\u{1F600}" == "\u{1F600}"
=> true
[3.1.2]> "\u{1F600}".b == "\u{1F600}"
=> false
```

Artichoke:
```console
=> true
>>> "A" == "A"
=> true
>>> "A".b == "A"
=> true
>>> "\u{1F600}" == "\u{1F600}"
=> true
>>> "\u{1F600}".b == "\u{1F600}"
=> true
```

## Solutions ideas

Rust psuedo-ish code change:

```rust
self.as_bytes() == other.as_bytes() && self.len() == other.len()
```

:point_up: I imagine `len()` would be fine since it is encoding aware, and should return the number of codepoints vs. bytes in the above.

The question I have is, should it be in `spinoso-string`, or in `string/trampoline.rs`. `spinoso-string` does mention `The Ruby String class`, which means we could assume that RStrings should do equality in this way - however it feels a little bit unexpected/hidden.

moozzi:
Fixes #2193

### DIFF
--- a/spinoso-string/src/enc/mod.rs
+++ b/spinoso-string/src/enc/mod.rs
@@ -69,7 +69,7 @@ impl PartialEq for EncodedString {
         // [3.0.2] > s == t
         // => true
         // ```
-        *self.as_slice() == *other.as_slice()
+        *self.as_slice() == *other.as_slice() && self.char_len() == other.char_len()
     }
 }
 
--- a/spinoso-string/src/enc/mod.rs
+++ b/spinoso-string/src/enc/mod.rs
@@ -69,7 +69,13 @@ impl PartialEq for EncodedString {
         // [3.0.2] > s == t
         // => true
         // ```
-        *self.as_slice() == *other.as_slice() && self.char_len() == other.char_len()
+        //
+        // See the functional tests in `string_test.rb` for more cases.
+        match (self.encoding(), other.encoding()) {
+            (self_enc, other_enc) if self_enc == other_enc => self.as_slice() == other.as_slice(),
+            _ if self.is_ascii_only() && other.is_ascii_only() => self.as_slice() == other.as_slice(),
+            _ => false,
+        }
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -818,7 +818,7 @@ dependencies = [
 
 [[package]]
 name = "spinoso-string"
-version = "0.20.0"
+version = "0.21.0"
 dependencies = [
  "bstr",
  "bytecount",
--- a/artichoke-backend/Cargo.toml
+++ b/artichoke-backend/Cargo.toml
@@ -34,7 +34,7 @@ spinoso-math = { version = "0.3.0", path = "../spinoso-math", optional = true, d
 spinoso-random = { version = "0.3.0", path = "../spinoso-random", optional = true }
 spinoso-regexp = { version = "0.4.0", path = "../spinoso-regexp", optional = true, default-features = false }
 spinoso-securerandom = { version = "0.2.0", path = "../spinoso-securerandom", optional = true }
-spinoso-string = { version = "0.20.0", path = "../spinoso-string", features = ["always-nul-terminated-c-string-compat"] }
+spinoso-string = { version = "0.21.0", path = "../spinoso-string", features = ["always-nul-terminated-c-string-compat"] }
 spinoso-symbol = { version = "0.3.0", path = "../spinoso-symbol" }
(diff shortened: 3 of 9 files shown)


spinoso-string/src/enc/mod.rs
```diff
@@ -69,7 +69,7 @@ impl PartialEq for EncodedString {
         // [3.0.2] > s == t
         // => true
         // ```
-        *self.as_slice() == *other.as_slice()
+        *self.as_slice() == *other.as_slice() && self.char_len() == other.char_len()
```
lopopolo: this test can be made cheaper by comparing the encodings I think, but this works for now.

lopopolo (changes requested):
Can you add an integration test for this here:

https://github.com/artichoke/artichoke/blob/84ac85cc0750ea066f7155574b4f7a136b623041/artichoke-backend/src/extn/core/string/string_test.rb#L5-L17

I think asserting that `"太贵了!!" != "太贵了!!".b` should do the trick:

```
[3.1.2] > "太贵了!!" == "太贵了!!".b
=> false
```

once the test is in that file, you `cargo test` will run it under artichoke and `ruby artichoke-backend/src/extn/core/string/string_test.rb` can be used to test it under MRI.

spinoso-string/src/enc/mod.rs
b-n:
```suggestion
        (*self.as_slice() == *other.as_slice() && self.encoding() == other.encoding()) ||
            (*self.as_slice() == * other.as_slice() && self.encoding() != self.encoding && self.is_ascii_only() && other.is_ascii_only())
```

:point_up: I didn't test this, so it's likely invalid syntax, however I think this is valid for at least the 3 encodings supported by `spinoso-string`

```console
irb(main):001:0> a = "\x7f"
=> "\u007F"
irb(main):002:0> b = a.b
=> "\x7F"
irb(main):003:0> a.encoding
=> #<Encoding:UTF-8>
irb(main):004:0> b.encoding
=> #<Encoding:ASCII-8BIT>
irb(main):005:0> a == b
=> true
irb(main):006:0> c = "\x80"
=> "\x80"
irb(main):007:0> d = c.b
=> "\x80"
irb(main):008:0> c == d
=> false
```

moozzi: @b-n You meant `self.encoding() != other.encoding()` not `self.encoding() != self.encoding` right?

b-n: Ah sorry, missed this question. I will reformat my suggestion since it's hard to see what it was trying to do :grimacing: 

```suggestion
        if self.encoding() == other.encoding() {
            *self.as_slice() == *other.as_slice()
        } else {
            *self_as_slice() == *other.as_slice() && self.is_ascii_only() && other.is_ascii_only()
        }
```

I noticed that if two ruby strings contain exactly the same characters, and so long as they are only ascii characters (0..=127) then it says it's the "same", even if the encoding doesn't match.

lopopolo:
sorry to let this PR languish @AI-Mozi. I hope you don't mind that I've fixed the RuboCop lint errors on your branch a rebased to latest trunk to get the build passing.

spinoso-string/src/enc/mod.rs
lopopolo: @b-n I've made this change. let me know what you think

lopopolo (approved):

spinoso-string/src/enc/mod.rs
b-n: Rust match block++, very very very nice.

b-n (approved):

lopopolo:
thanks for the teamwork on this one folks!

Issue #2115, opened 2022-08-28
Implement `Array#append` and `Array#push` in Rust


lopopolo:
`Array#append` and `Array#push` take a list of arguments and extend an `Array` with them. Values are appended as is, even if they are arrays:

- https://ruby-doc.org/core-3.1.2/Array.html#method-i-append
- https://ruby-doc.org/core-3.1.2/Array.html#method-i-push

MRI implementation:

- https://github.com/ruby/ruby/blob/v3_1_2/array.c#L1330-L1354
- https://github.com/ruby/ruby/blob/v3_1_2/array.c#L1320-L1328

```c
/*
 *  call-seq:
 *    array << object -> self
 *
 *  Appends +object+ to +self+; returns +self+:
 *
 *    a = [:foo, 'bar', 2]
 *    a << :baz # => [:foo, "bar", 2, :baz]
 *
 *  Appends +object+ as one element, even if it is another \Array:
 *
 *    a = [:foo, 'bar', 2]
 *    a1 = a << [3, 4]
 *    a1 # => [:foo, "bar", 2, [3, 4]]
 *
 */
```

`Array#push` is currently implemented in Ruby by delegating to `Array#concat`, which is incorrect because `push` should not flatten `Array` arguments:

https://github.com/artichoke/artichoke/blob/da852668b8588935929b2e6cfb655586da23d575/artichoke-backend/src/extn/core/array/array.rb#L925-L930

https://github.com/artichoke/artichoke/blob/da852668b8588935929b2e6cfb655586da23d575/artichoke-backend/src/extn/core/array/array.rb#L1368

anshulxyz:
Hi Ryan, I'd like to pick this up. If no has started working on it yet.

lopopolo:
Sounds great @anshulxyz. Let me know here if you have any questions.

anshulxyz:
I need to investigate source code. But what I'm immediately looking for are the test cases that check this feature. So that I will be running them while development.

lopopolo:
This might be a good opportunity to add some. There's a functional test file called `array_test.rb` right next to the trampoline. You can also use the Ruby specs with the harness in the spec-runner directory.

anshulxyz:
hi @lopopolo can you give me `airb` examples of "`push` should not flatten `Array` arguments"

because I do not see the flattening
```ruby
>>> [1,2,3].push([4])
=> [1, 2, 3, [4]]
```

lopopolo:
[...]

PR #1323, opened 2021-08-14
Rebox modified `Array` into `mrb_value` before alloc in `shift`

lopopolo:
If an `Integer` argument is given to `Array#shift`, multiple items are
removed from the front of the `Array`'s underlying vector. These removed
elements are returned from `Array#shift` as a new `Array`. `Array#shift`
with an `Integer` argument is similar to an inverted `Vec::split_off` from
the Rust std.

The the array trampoline in `artichoke-backend` failed to repack the raw
parts of the `Array` into the receiver `mrb_value` before allocating a
new `Array` for the shifted elements.

All allocations in the mruby heap can trigger a garbage collection. If
the allocation of the new `Array` for the shifted elements triggered a
GC, as it does in the `spec-runner` with `yaml` formatter as outlined in
issue #1320, the mruby GC will attempt to mark all of the children in
the original receiver. The pointers in the receiver's `RArray *` will be
garbage, so attempting to index into the raw pointer will trigger
undefined behavior, which was expressing as a segfault.

This commit re-orders the code to ensure the receiver is repacked before
attempting to allocate a new `Array` for the result of `shift`.

The following invocation of the spec-runner succeeds as of this commit:

    ./target/debug/spec-runner --format yaml all-core-specs.toml

Fixes #1320.

🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉 🧯 🎉

PR #2309, opened 2022-12-12
Implement Array#push in Rust

Tseyang:
Related to the thread in #2115 regarding implementing `Array#push` in Rust.

My understanding in the thread was that the existing implementation is actually
correct, but this operation would ideally be implemented in Rust.

I'm far from a Rust expert so I had some trouble getting types to line up when
trying to replicate the approach of calling into `concat` directly. Regardless,
I took an initial stab at this out of interest and am resorting to calling
`push_single` a bunch of times. Not sure if this is actually a bad approach so
open to feedback.

Also, am a fan of both languages so thank you for making this project which
piqued my interest enough to attempt an OSS PR for the first time :).

### DIFF
--- a/artichoke-backend/src/extn/core/array/trampoline.rs
+++ b/artichoke-backend/src/extn/core/array/trampoline.rs
@@ -151,11 +151,17 @@ where
         return Err(FrozenError::with_message("can't modify frozen Array").into());
     }
 
+    let mut array = unsafe { Array::unbox_from_value(&mut ary, interp)? };
+
+    let array_mut = unsafe { array.as_inner_mut() };
+
     for value in others {
-        match push_single(interp, ary, value) {
-            Ok(new_ary) => ary = new_ary,
-            Err(e) => return Err(e),
-        }
+        array_mut.push(value);
+    }
+
+    unsafe {
+        let inner = array.take();
+        Array::box_into_value(inner, ary, interp)?;
     }
 
     Ok(ary)
--- a/artichoke-backend/src/extn/core/array/trampoline.rs
+++ b/artichoke-backend/src/extn/core/array/trampoline.rs
@@ -142,6 +142,25 @@ pub fn clear(interp: &mut Artichoke, mut ary: Value) -> Result<Value, Error> {
     Ok(ary)
 }
 
+pub fn push<I>(interp: &mut Artichoke, mut ary: Value, others: I) -> Result<Value, Error>
+where
+    I: IntoIterator<Item = Value>,
+    I::IntoIter: Clone,
+{
+    if ary.is_frozen(interp) {
+        return Err(FrozenError::with_message("can't modify frozen Array").into());
+    }
+
+    for value in others {
+        match push_single(interp, ary, value) {
+            Ok(new_ary) => ary = new_ary,
+            Err(e) => return Err(e),
+        }
+    }
+
+    Ok(ary)
+}
+
 pub fn concat<I>(interp: &mut Artichoke, mut ary: Value, others: I) -> Result<Value, Error>
 where
     I: IntoIterator<Item = Value>,
--- a/artichoke-backend/src/extn/core/array/trampoline.rs
+++ b/artichoke-backend/src/extn/core/array/trampoline.rs
@@ -153,11 +153,11 @@ where
 
     let mut array = unsafe { Array::unbox_from_value(&mut ary, interp)? };
 
+    // SAFETY: The array is repacked without any intervening interpreter heap
+    // allocations.
     let array_mut = unsafe { array.as_inner_mut() };
 
-    for value in others {
-        array_mut.push(value);
-    }
+    array_mut.extend(others);
 
     unsafe {
         let inner = array.take();
--- a/artichoke-backend/src/extn/core/array/array.rb
+++ b/artichoke-backend/src/extn/core/array/array.rb
@@ -918,13 +918,6 @@ class Array
     raise NotImplementedError
   end
 
-  def push(*args)
-    raise FrozenError, "can't modify frozen Array" if frozen?
-
-    concat(args)
-    self
-  end
-
   def rassoc(obj)
     idx = 0
     len = length
--- a/artichoke-backend/src/extn/core/array/array_functional_test.rb
+++ b/artichoke-backend/src/extn/core/array/array_functional_test.rb
@@ -339,7 +339,7 @@ def push
   a = []
   b = [1, 2]
   a.push b
-  raise unless a == [[1,2]]
+  raise unless a == [[1, 2]]
 end
 
 def concat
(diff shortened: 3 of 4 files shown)


artichoke-backend/src/extn/core/array/trampoline.rs
```diff
@@ -142,6 +142,25 @@ pub fn clear(interp: &mut Artichoke, mut ary: Value) -> Result<Value, Error> {
...
+{
+    if ary.is_frozen(interp) {
+        return Err(FrozenError::with_message("can't modify frozen Array").into());
+    }
+
+    for value in others {
+        match push_single(interp, ary, value) {
+            Ok(new_ary) => ary = new_ary,
+            Err(e) => return Err(e),
+        }
```
b-n: Pretty clever idea, however i think it might cause extra/new allocations. `push_single` retrieves the array from memory, adds the item to the address is memory. I think it would be better to bring some of the `push_single` logic into this function. something like:

```rust
let array_mut = unsafe { array.as_inner_mut() };
for value in others {
  array_mut.push(value);
}
```

and then reboxing as per `push_single`.

b-n (dismissed):
Thanks for picking this up! Just a suggestion instead of defering to `push_single`.

artichoke-backend/src/extn/core/array/trampoline.rs
Tseyang: @b-n thanks for the feedback, yea I was just trying to stay away from `unsafe` even though I realize its kinda needed for this stuff.

I tried adjusting based on your suggestion and the way `push_single` was implemented.

Can I ask: conceptually, I understand that `Array` implements the Ruby `Array` API (in Rust), but in terms of the underlying data, how should I think of it? Is it better to perceive it as a pointer-type abstraction? The function names here seem to suggest that its a Box-ed item (e.g. `unbox_from_value` etc.).

lopopolo: The Ruby value stored in the array is a tagged union that can either be an integral Ruby type (like nil, true, false, a float, int, or symbol), an interpreter heap-allocated object like a String or Array, or a Rust allocated heap object like a Regexp. These types are stored in a pointer, int64, float64 C union.

The use of "boxed" here is more similar to that of a Java boxed `Integer` than Rust's `Box`.

artichoke-backend/src/extn/core/array/trampoline.rs
```diff
@@ -142,6 +142,31 @@ pub fn clear(interp: &mut Artichoke, mut ary: Value) -> Result<Value, Error> {
...
+{
+    if ary.is_frozen(interp) {
+        return Err(FrozenError::with_message("can't modify frozen Array").into());
+    }
+
+    let mut array = unsafe { Array::unbox_from_value(&mut ary, interp)? };
+
+    let array_mut = unsafe { array.as_inner_mut() };
+
+    for value in others {
```
lopopolo: This loop can be replaced with:

```
array_mut.extend(others)
```

artichoke-backend/src/extn/core/array/trampoline.rs
```diff
@@ -142,6 +142,31 @@ pub fn clear(interp: &mut Artichoke, mut ary: Value) -> Result<Value, Error> {
...
+    I: IntoIterator<Item = Value>,
+    I::IntoIter: Clone,
+{
+    if ary.is_frozen(interp) {
+        return Err(FrozenError::with_message("can't modify frozen Array").into());
+    }
+
+    let mut array = unsafe { Array::unbox_from_value(&mut ary, interp)? };
+
+    let array_mut = unsafe { array.as_inner_mut() };
```
lopopolo, line 158: Just to walk through what's happening here in the code you've written:

Grab a mutable reference to the underlying Rust array which we are viewing from a boxed Ruby value. This operation is unsafe because a mutable reference allows us to modify the array and cause a reallocation, which will invalidate the way the Array is stored in the boxed Ruby value and leave a dangling pointer.

Then we push some values into the array. This may cause the array to be reallocated.

The initial use of unsafe is safe though because immediately after a potential reallocation we consume the mutable reference and repack the arrays raw parts into the boxed Ruby value.

The important thing is we don't cause any interpreter heap allocations while the boxed Ruby value is potentially invalidated. If we were to do that, it could trigger a GC which would likely cause a crash when marking the array we are operating on.

Would you be able to add a `SAFETY ` comment similar to the other ones in this file that captures this proof of correctness?

lopopolo (changes requested):
Thanks for the PR. This is a good start. I've left a couple of suggestions.

artichoke-backend/src/extn/core/array/trampoline.rs
lopopolo: To use some Rust terminology, the Ruby value is an enum where one of its variants has a `Box<Vec<RubyValue>>` payload.

artichoke-backend/src/extn/core/array/trampoline.rs
Tseyang: Thanks for the detailed explanation. I think I had the idea abstractly but its been a while since I've had to actually think about pointers (I mainly use Ruby in my day-to-day now). 

Just to make sure I have the right understanding of the bad-path: incorrect behavior would be attempting to mutate the `mut array` in a manner that might cause reallocations before the repacking from `as_inner_mut` e.g. calling `array.extend` directly after the first `unsafe`?

lopopolo: there would be bad behavior if after the call to `array_mut.extend` but before reboxing `array` into `ary` we called something like `interp.try_convert_mut("abc")` and allocated a Ruby String. This allocation could cause a GC. When the GC visited `ary` to try and mark its children, `ary` would contain an invalid pointer.

For a concrete example and real Artichoke bug due to this, take a peek at:

- https://github.com/artichoke/artichoke/pull/1323

lopopolo (approved):

lopopolo:
it looks like RuboCop isn't quite happy. It looks like running an autofix should get this PR green. You can do this locally:

```shell
bundle install
bundle exec rake lint:rubocop:autocorrect
```

Tseyang:
ah I thought I had fixed that, pushed up a fix.

lopopolo:
@b-n it looks like you've got an ❌ on this PR. This LGTM; if @Tseyang has addressed the changes you wanted to see please 👍 and merge.

Tseyang:
@b-n bumping this, any other changes you want me to make?

Tseyang:
@lopopolo should we keep waiting on @b-n ?

lopopolo (approved):

lopopolo:
Thanks for the PR @Tseyang! and thanks for following up to poke me to merge this

PR #2356, opened 2023-01-04
Raise Ruby LoadError instead of Rust IoError when fail to load a program file

### LINKED ISSUE #1579: `artichoke` returns different errors than MRI when failing to load a program file

choznerol:
Fixes #1579

Demo:

The current implementation is still a little different from MRI in the `(os error 2)` or `(os error 12)`. I'm not sure if the `(os error x)` is part of the default behavior of underlying OS error because I can't really find related code in Artichoke codebase. If that's the case, would we still want to remove the `(os error x)` part with something like `.replace()`?

### DIFF
--- a/artichoke-backend/src/eval.rs
+++ b/artichoke-backend/src/eval.rs
@@ -2,17 +2,18 @@ use std::ffi::OsStr;
 use std::path::Path;
 
 use scolapasta_path::os_str_to_bytes;
-use spinoso_exception::{ArgumentError, Fatal};
+use spinoso_exception::{ArgumentError, Fatal, LoadError};
 
 use crate::core::{Eval, LoadSources, Parser};
 use crate::error::Error;
-use crate::exception_handler;
 use crate::ffi::InterpreterExtractError;
 use crate::state::parser::Context;
 use crate::sys;
 use crate::sys::protect;
 use crate::value::Value;
 use crate::Artichoke;
+use crate::{exception_handler, RubyException};
+use bstr::ByteSlice;
 
 impl Eval for Artichoke {
     type Value = Value;
@@ -56,7 +57,13 @@ impl Eval for Artichoke {
         let context = Context::new(os_str_to_bytes(file.as_os_str())?.to_vec())
             .ok_or_else(|| ArgumentError::with_message("path name contains null byte"))?;
         self.push_context(context)?;
-        let code = self.read_source_file_contents(file)?.into_owned();
+        let code = self
+            .read_source_file_contents(file)
+            .map_err(|err| {
+                let message = format!("ruby: {} -- {}", err.message().as_bstr(), file.to_string_lossy());
+                LoadError::from(message)
+            })?
+            .into_owned();
         let result = self.eval(code.as_slice());
         self.pop_context()?;
         result
@@ -65,6 +72,8 @@ impl Eval for Artichoke {
 
 #[cfg(test)]
 mod tests {
+    use std::path::Path;
+
     use bstr::ByteSlice;
 
     use crate::test::prelude::*;
@@ -219,5 +228,16 @@ mod tests {
             .unwrap();
         let err = interp.eval(b"require 'fail'").unwrap_err();
         assert_eq!("SyntaxError", err.name().as_ref());
+    }
+
+    #[test]
+    fn eval_file_error_file_not_found() {
+        let mut interp = interpreter();
+        let err = interp.eval_file(Path::new("no/such/file.rb")).unwrap_err();
+        assert_eq!("LoadError", err.name().as_ref());
+        assert_eq!(
+            "ruby: file not found in virtual file system -- no/such/file.rb",
+            err.message().as_ref().as_bstr()
+        );
     }
 }
--- a/artichoke-backend/src/eval.rs
+++ b/artichoke-backend/src/eval.rs
@@ -1,6 +1,7 @@
 use std::ffi::OsStr;
 use std::path::Path;
 
+use bstr::ByteSlice;
 use scolapasta_path::os_str_to_bytes;
 use spinoso_exception::{ArgumentError, Fatal, LoadError};
 
@@ -13,7 +14,6 @@ use crate::sys::protect;
 use crate::value::Value;
 use crate::Artichoke;
 use crate::{exception_handler, RubyException};
-use bstr::ByteSlice;
 
 impl Eval for Artichoke {
     type Value = Value;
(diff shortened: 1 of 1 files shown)


artichoke-backend/src/eval.rs
```diff
@@ -56,7 +57,13 @@ impl Eval for Artichoke {
         let context = Context::new(os_str_to_bytes(file.as_os_str())?.to_vec())
             .ok_or_else(|| ArgumentError::with_message("path name contains null byte"))?;
         self.push_context(context)?;
-        let code = self.read_source_file_contents(file)?.into_owned();
+        let code = self
+            .read_source_file_contents(file)
+            .map_err(|err| {
+                let message = format!("ruby: {} -- {}", err.message().as_bstr(), file.to_string_lossy());
```
lopopolo: rather than using `to_string_lossy` here, can we construct the message as a `Vec<u8>`?

artichoke-backend/src/eval.rs
```diff
@@ -220,4 +229,15 @@ mod tests {
         let err = interp.eval(b"require 'fail'").unwrap_err();
         assert_eq!("SyntaxError", err.name().as_ref());
     }
+
+    #[test]
+    fn eval_file_error_file_not_found() {
+        let mut interp = interpreter();
+        let err = interp.eval_file(Path::new("no/such/file.rb")).unwrap_err();
```
lopopolo, line 244: this is a really nifty way to test this. I thought we'd need a UI test. Thank you!

lopopolo (changes requested):
this looks good modulo the `Vec` ask. I'm ok with the difference between the error messages which includes the os error code in Rust.

artichoke-backend/src/eval.rs
choznerol: Since [`Path` is not guarantee to be valid `UTF-8`](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_str), I'm not sure if we have many choices here. I added a test to demonstrate the case when path contains invalid UTF-8: https://github.com/artichoke/artichoke/pull/2356/commits/19330239a7f1a17de10022abdcf4bc68d9065536

I'm not sure if I understand correctly, but displaying `Vec<u8>` will compile, but the message doesn't seem right https://github.com/choznerol/artichoke/commit/71c54d2c5852969c33becd952c7515f6fdfb769e

Using `.to_str()` will require `.unwrap()` and the newly added test case will panic: https://github.com/choznerol/artichoke/commit/5c2b973944ac386798c6fe31a425d83b416be4a0

artichoke-backend/src/eval.rs
choznerol: I was just copying `interp.eval()` from other test cases 😄

artichoke-backend/src/eval.rs
choznerol: Update: found that `.to_string_loussy()` can be replaced by `.display()` and produce the same effect. Pushed cff55d2.

artichoke-backend/src/eval.rs
```diff
@@ -56,7 +57,13 @@ impl Eval for Artichoke {
         let context = Context::new(os_str_to_bytes(file.as_os_str())?.to_vec())
             .ok_or_else(|| ArgumentError::with_message("path name contains null byte"))?;
         self.push_context(context)?;
-        let code = self.read_source_file_contents(file)?.into_owned();
+        let code = self
+            .read_source_file_contents(file)
+            .map_err(|err| {
+                let message = format!("ruby: {} -- {}", err.message().as_bstr(), file.display());
```
lopopolo: Sorry for not being clear on this. The underlying representation for a Ruby `String` and Artichoke exception messages is `Vec<u8>`. Because we are extracting the error message from an existing error, we end up with bytes, so we should stay with bytes.

Calling `as_bstr` on those bytes in the original message results in a transform if the message contains invalid UTF-8 byte sequences, which we'd like to avoid. Calling `display` on the path does the same destructive transform.

To help us stay in bytes, let's construct the new message like this (forgive the pseudocode, I'm on my mobile):

```rust
let mut message = b"ruby: ";
message.extend_from_slice(err.message());
message.extend_from_slice(b" - ");
message.extend_from_slice(scolapasta_path::os_str_to_bytes(path.into())?);
LoadError::from(message)
```

There is a fallible helper function in `scolapasta_path` which converts an `OsStr` to a byte slice and `Path` is freely convertible to `OsStr`. https://artichoke.github.io/artichoke/scolapasta_path/fn.os_str_to_bytes.html

lopopolo (changes requested):

artichoke-backend/src/eval.rs
choznerol: Oh I see. Thanks for the detailed explanation! I try to handle the `ConvertBytesError` the way [a similar case in `load.rs`](https://github.com/choznerol/artichoke/blob/c191d021bdd1f1953461ef99bbb1f824b8441f6b/artichoke-backend/src/load.rs#L101-L106) does.

Fixed in c191d02

artichoke-backend/src/eval.rs
```diff
@@ -220,4 +237,29 @@ mod tests {
...
+        assert_eq!("LoadError", err.name().as_ref());
+        assert_eq!(
+            b"ruby: file not found in virtual file system -- no/such/file.rb",
+            err.message().as_ref()
+        );
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn eval_file_error_invalid_path() {
```
lopopolo, line 254: thank you for adding this test!

lopopolo (approved):
this looks great!

artichoke-backend/src/eval.rs
lopopolo: thank you!

Issue #2223, opened 2022-10-27
Implement `Time.utc` and `Time.local`


b-n:
Normally I'd just create a PR for this, however it looks like a not insignificant amount of work in order to get there due to the way the way these functions in the title work.

## Docs and context

Ruby Docs Example: https://ruby-doc.org/core-3.1.2/Time.html#method-c-utc

Note: The function can take `1..8` args or 10 args. The errors in MRI for 9 args implies that the 10 arg variant does not exist:

```console
[3.1.2] > Time.utc(1,2,3,4,5,6,7,8,9)
(irb):33:in `utc': wrong number of arguments (given 9, expected 1..8) (ArgumentError)
	from (irb):33:in `<main>'                                
	from /home/ben/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
	from /home/ben/.rbenv/versions/3.1.2/bin/irb:25:in `load'
	from /home/ben/.rbenv/versions/3.1.2/bin/irb:25:in `<main>'
```

The reason for the 10 arg support is so the following function works:

```console
[3.1.2] > Time.utc(*Time.now.to_a)   #splat the array of 10 values into args
=> 2022-10-27 10:48:57 UTC
```

:warning: The big issue here is that the order of the args in the `1..8` version is different than in the 10 version variant. Specifically this:

```
utc(year, month=1, day=1, hour=0, min=0, sec_i=0, usec=0) → new_time
utc(sec_i, min, hour, day, month, year, dummy, dummy, dummy, dummy) → new_time 
```

## How MRI/Mruby handle this

MRI solves this as follows: https://github.com/artichoke/artichoke/blob/trunk/artichoke-backend/vendor/ruby/time.c#L2861-L2870 (just re-order the values if we have 10 args). Called from here: https://github.com/artichoke/artichoke/blob/trunk/artichoke-backend/vendor/ruby/time.c#L3362-L3369

MRuby just ignores the 10 var variant: https://github.com/artichoke/artichoke/blob/trunk/artichoke-backend/vendor/mruby/mrbgems/mruby-time/src/time.c#L492-L503

## Solution design

[...]

PR #2410, opened 2023-02-04
Add support for Time.utc and Time.local

b-n:
This makes progress towards #2223, however it does not yet close it. Breaking this work down to more atomic changes.

Notes:
- The new exclusions on the spec runner is due to needing to add even more support in parsing time Args. I believe MRI 3.x added support for the text values `"jan|feb|..."` etc
- changes to the math class is due to `TryConvertMut` supporting two targets from `&[Value]` so the compiler now needs hinting
- I've left the 10 arg impl empty for now as a step towards `Time.utc`. This should be easily addable later, but happy to do it here too.

Some design thoughts:
- I think the checking of the individual arguments should be handled within the `TryConvertMut` block, instead of in the accessor functions. I originally did it this way because I naively thought I could `to_int` all the things, but apparently not (month param + sec taking fractional params)
- If I did the above, I could make it just return a big tuple of the values (although this might be harder since `Time.new` supports a `zone` parameter from the looks of it)
- I have tried to leave the coercion of the args into `Value::value` in the `mruby.rs` file - since this feels more consistent with other implementations. That way, I can always say that trampoiline expects a RubyValue. However this limits the implementation a bit - e.g. right now we're doing some args length checking after coercion to RubyValues, where as I believe those checks should be done much earlier. Specificallly, I think it would be nicer if the new time `Args` was being generated directly from `mruby.rs` and being passed through. I'll comment in the code where I think this is applicable.

Misc note:
- Parts of this PR were commited/created in New Zealand. So Artichoke has now added and additional timezone of contribution :sweat_smile:)


artichoke-backend/src/extn/core/time/mruby.rs
```diff
@@ -130,7 +130,7 @@ unsafe extern "C" fn time_self_at(mrb: *mut sys::mrb_state, _slf: sys::mrb_value
 unsafe extern "C" fn time_self_mkutc(mrb: *mut sys::mrb_state, _slf: sys::mrb_value) -> sys::mrb_value {
     let args = mrb_get_args!(mrb, *args);
     unwrap_interpreter!(mrb, to => guard);
-    let args = args.iter().copied().map(Value::from);
+    let args = &args.iter().copied().map(Value::from).collect::<Vec<Value>>();
```
b-n: It's tempting to pass this iterator through as it was done previously, however the length of the args is a very quick validation, and makes the rest of the checks a whole lot easier - thus the collection into a Vec. 

However, I think this would be better if we actually just let the `Args` take the `mrb_get_args!` returned slice, and then do all of the operations directly in `Args` itself. The reason, we can avoid the conversion to `Value` fairly quickly by just checking the slice length etc. It could be duplicated logic, but IMO it makes more sense to just use `Args` to parse these values directly. Thoughts?

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -0,0 +1,86 @@
...
+
+        let raw_args = [];
+
+        let err = as_time_args(&mut interp, raw_args.into_iter()).unwrap_err();
+
+        assert_eq!(err.name(), "ArgumentError");
+        assert_eq!(
+            err.message().as_bstr(),
+            b"wrong number of arguments (given 0, expected 1..8)"
+                .as_slice()
```
lopopolo: bstr 1.2 has impls for `[u8; N]` so this `as_slice` isn't necessary anymore!

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
-    let _interp = interp;
-    let _args = args;
-    Ok(TimeArgs::default())
+impl TimeArgs {
+    pub fn year(&self) -> Result<i32, Error> {
+        i32::try_from(self.year).map_err(|_| ArgumentError::with_message("year out of range").into())
+    }
+
+    pub fn month(&self) -> Result<u8, Error> {
+        // 1..=12 is gauranteed to fit into u8, so this cast is safe
```
lopopolo: typo: gauranteed -> guaranteed

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+        match self.month {
+            #![allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
```
lopopolo: we can avoid the cast by structuring the code like this:

```rust
match u8::try_from(self.month) {
    Ok(month @ 1..12) => Ok(month),
    _ => Err(ArgumentError::with_message("mon out of range").into()),
}
```

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+            1..=12 => Ok(self.month as u8),
+            _ => Err(ArgumentError::with_message("mon out of range").into()),
+        }
+    }
+
+    pub fn day(&self) -> Result<u8, Error> {
+        // 1..=31 is gauranteed to fit into u8, so this cast is safe
```
lopopolo: typo: gauranteed -> guaranteed

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+        match self.day {
+            #![allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
```
lopopolo: ditto above, using a `try_from` allows us to avoid this cast and clippy lint allows.

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+            1..=31 => Ok(self.day as u8),
+            _ => Err(ArgumentError::with_message("mday out of range").into()),
+        }
+    }
+
+    pub fn hour(&self) -> Result<u8, Error> {
+        // 1..=23 is gauranteed to fit into u8, so this cast is safe
```
lopopolo: typo: gauranteed -> guaranteed

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+        match self.hour {
+            #![allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
```
lopopolo: ditto above

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+            0..=23 => Ok(self.hour as u8),
+            _ => Err(ArgumentError::with_message("hour out of range").into()),
+        }
+    }
+
+    pub fn minute(&self) -> Result<u8, Error> {
+        // 1..=59 is gauranteed to fit into u8, so this cast is safe
```
lopopolo: typo: gauranteed -> guaranteed

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+        match self.minute {
+            #![allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
```
lopopolo: ditto

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+            0..=59 => Ok(self.minute as u8),
+            _ => Err(ArgumentError::with_message("min out of range").into()),
+        }
+    }
+
+    pub fn second(&self) -> Result<u8, Error> {
+        // 1..=60 is gauranteed to fit into u8, so this cast is safe
```
lopopolo: typo: gauranteed -> guaranteed

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+        match self.second {
+            #![allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
```
lopopolo: ditto

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+            0..=60 => Ok(self.second as u8),
+            _ => Err(ArgumentError::with_message("sec out of range").into()),
+        }
+    }
+
+    pub fn nanoseconds(&self) -> Result<u32, Error> {
+        // TimeArgs take a micros parameter, not a nanos value. The below
+        // multiplication and casting is gauranteed to be inside a `u32`.
+        match self.micros {
+            #![allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
```
lopopolo: ditto

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+    }
+}
+
+pub fn as_time_args(interp: &mut Artichoke, args: &[Value]) -> Result<TimeArgs, Error> {
+    // TimeArgs are in order of year, month, day, hour, minute, second, micros.
+    // This is unless there are 10 arguments provided (`Time#to_a` format), at
+    // which points it is second, minute, hour, day, month, year. The number of
+    // expected parameters doesn't give this hint though.
+
+    match args.len() {
```
lopopolo: this is pretty cool

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -20,33 +21,140 @@ impl Default for TimeArgs {
...
+            }
+            Ok(result)
+        }
+        10 => todo!(),
+        _ => unreachable!(),
+    }
 }
 
 #[cfg(test)]
 mod tests {
```
lopopolo, line 156: these tests are looking awesome

artichoke-backend/src/extn/core/time/mruby.rs
lopopolo: there are a couple of other places in `mruby.rs` files that we allocate a `Vec` for a `*rest` args. I expect this to go away when we are not adapting mruby's value type to Artichoke's, so this seems fine for now.

artichoke-backend/src/extn/core/time/trampoline.rs
```diff
@@ -114,20 +114,18 @@ pub fn at(
     Time::alloc_value(time, interp)
 }
 
-pub fn mkutc<T>(interp: &mut Artichoke, args: T) -> Result<Value, Error>
-where
-    T: IntoIterator<Item = Value>,
+pub fn mkutc(interp: &mut Artichoke, args: &[Value]) -> Result<Value, Error>
```
lopopolo: :chef-kiss:

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -12,9 +12,9 @@ pub struct TimeArgs {
     micros: i64,
 }
 
-impl Default for TimeArgs {
-    fn default() -> TimeArgs {
-        TimeArgs {
+impl Default for Args {
+    fn default() -> Args {
+        Args {
```
lopopolo: Artichoke style is to use `Self` for these appearances of the type name.

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -87,78 +87,84 @@ impl TimeArgs {
...
+            0 | 9 | 11.. => {
+                let mut message = br#"wrong number of arguments (given "#.to_vec();
+                message.extend(args.len().to_string().bytes());
+                message.extend_from_slice(b", expected 1..8)");
+                Err(ArgumentError::from(message).into())
+            }
+            1..=8 => {
+                // For 0..=7 params, we need to validate to_int
+                let mut result = Args::default();
+                for (i, arg) in args.iter().enumerate() {
```
lopopolo: we can add a `&` pattern here to remove some dereference symbol noise from the code below:

```rust
for (i, &arg) in args.iter().enumerate() {
```

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -87,78 +87,84 @@ impl TimeArgs {
...
+                        }
+                        _ => unreachable!(),
                     }
-                    _ => unreachable!(),
                 }
+                Ok(result)
             }
-            Ok(result)
+            10 => todo!(),
+            _ => unreachable!(),
```
lopopolo, line 150: is rustc not smart enough to see that this match is exhaustive. 😢 if that's the case.

artichoke-backend/src/extn/core/time/mruby.rs
```diff
@@ -141,7 +141,7 @@ unsafe extern "C" fn time_self_mkutc(mrb: *mut sys::mrb_state, _slf: sys::mrb_va
 unsafe extern "C" fn time_self_mktime(mrb: *mut sys::mrb_state, _slf: sys::mrb_value) -> sys::mrb_value {
     let args = mrb_get_args!(mrb, *args);
     unwrap_interpreter!(mrb, to => guard);
-    let args = args.iter().copied().map(Value::from);
+    let args = &args.iter().copied().map(Value::from).collect::<Vec<Value>>();
```
lopopolo: I think it would be a bit more clear (and not rely on rustc promotion) to take the borrow on the vec in the below line:

```rust
let args = args.iter().copied().map(Value::from).collect::<Vec<Value>>();
let result = trampoline::mktime(&mut guard, &args);
```

lopopolo (changes requested):
This looks pretty awesome. I left some comments as I went through commit by commit.

I'll take another pass at the PR as a whole.

lopopolo (commented):
I have no concerns with the approach you've taken @b-n. It looks readable and well structured. The `Args` type keeps all the parsing local and easy to reason about, especially for these TODOs you've called out about supporting some of the polymorphic parameters.

Most of my comments are nits, not to do with the logic itself.

artichoke-backend/src/extn/core/time/args.rs
b-n: `Abolish gaurantee{,s,d} guarantee{}` to the rescue!

Note: With the `try_from` I think the comment isn't necessary anymore, but still good to have this abolished.

artichoke-backend/src/extn/core/time/args.rs
b-n: Hah indeed. It's maybe just one level of abstraction too far for it to know that args.len() and the index on the iterator are the same thing I guess :shrug:

lopopolo (approved):
Looks great.

PR #2411, opened 2023-02-06
Enable parsing of Time Args with 10 parameters

b-n:
Quick follow on from #2410 

Adds ten argument support (e.g. `Time.local(*Time.now.to_a)`)
Refactor the existing logic to do all type and range checking within the initial parsing. (See the commit details for some more info).

This solves a bunch of problems I had in my head about how to structure this code. It might be somewhat reusable for the likes of `Time::new`, but will have to see.

Related to: #2223

### DIFF
--- a/artichoke-backend/src/extn/core/time/args.rs
+++ b/artichoke-backend/src/extn/core/time/args.rs
@@ -58,7 +58,7 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
             match i {
                 0 => {
                     let arg = to_int(self, arg)?;
-                    let arg: i64 = arg.try_convert_into::<Option<i64>>(self)?.unwrap();
+                    let arg: i64 = arg.try_convert_into(self)?;
 
                     result.year = i32::try_from(arg).map_err(|_| ArgumentError::with_message("year out of range"))?;
                 }
@@ -66,7 +66,7 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
                     // TODO: This should support 3 letter month names
                     // as per the docs. https://ruby-doc.org/3.1.2/Time.html#method-c-new
                     let arg = to_int(self, arg)?;
-                    let arg: i64 = arg.try_convert_into::<Option<i64>>(self)?.unwrap();
+                    let arg: i64 = arg.try_convert_into(self)?;
 
                     result.month = match u8::try_from(arg) {
                         Ok(month @ 1..=12) => Ok(month),
@@ -75,7 +75,7 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
                 }
                 2 => {
                     let arg = to_int(self, arg)?;
-                    let arg: i64 = arg.try_convert_into::<Option<i64>>(self)?.unwrap();
+                    let arg: i64 = arg.try_convert_into(self)?;
 
                     result.day = match u8::try_from(arg) {
                         Ok(day @ 1..=31) => Ok(day),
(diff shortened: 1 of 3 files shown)


artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -21,134 +21,133 @@ impl Default for Args {
...
         }
-    }
 
-    pub fn day(&self) -> Result<u8, Error> {
-        match u8::try_from(self.day) {
-            Ok(day @ 1..=31) => Ok(day),
-            _ => Err(ArgumentError::with_message("mday out of range").into()),
-        }
-    }
+        let mut args = args.to_vec();
```
b-n: Note, we need to change this to an owned structure because we're mutating it in the 10 argument variant. I'm sure there's a way to do this just with slices but couldn't figure it out. Alternative is to just make the input `mut` since it's already an artichoke owned piece of memory from `mruby.rs` anyway :shrug:

lopopolo: I'd rather us take a `&mut [Value]` for the `TryConvertMut` impl and push the allocation to callers if they need to. The API design principle here is _ask for what you need_ and indeed, callers of this converter already have access to a mutable slice, so we can save an allocation.

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -21,134 +21,133 @@ impl Default for Args {
...
+        // Args are in order of year, month, day, hour, minute, second, micros.
+        // This is unless there are 10 arguments provided (`Time#to_a` format),
+        // at which points it is second, minute, hour, day, month, year.
+        if args.len() == 10 {
+            args.swap(0, 5);
+            args.swap(1, 4);
+            args.swap(2, 3);
+            // All arguments after position 5 are ignored in the 10 argument
+            // variant.
+            args.truncate(6);
```
lopopolo: when converting this code to take a slice, we'd just truncate the slice like `args = &args[..6]`, right?

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -21,134 +21,133 @@ impl Default for Args {
...
-        match u8::try_from(self.second) {
-            Ok(second @ 0..=60) => Ok(second),
-            _ => Err(ArgumentError::with_message("sec out of range").into()),
-        }
-    }
+        for (i, &arg) in args.iter().enumerate() {
+            match i {
+                0 => {
+                    let arg = to_int(self, arg)?;
+                    let arg: i64 = arg.try_convert_into::<Option<i64>>(self)?.unwrap();
```
lopopolo: I'm a bit confused why we convert to `Option<i64>` and then unwrap. Can we replace this with this:

```suggestion
                    let arg = arg.try_convert_into::<i64>(self)?;
```

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -21,134 +21,133 @@ impl Default for Args {
...
-        }
-    }
-}
+                    result.year = i32::try_from(arg).map_err(|_| ArgumentError::with_message("year out of range"))?;
+                }
+                1 => {
+                    // TODO: This should support 3 letter month names
+                    // as per the docs. https://ruby-doc.org/3.1.2/Time.html#method-c-new
+                    let arg = to_int(self, arg)?;
+                    let arg: i64 = arg.try_convert_into::<Option<i64>>(self)?.unwrap();
```
lopopolo: ditto for all of these

lopopolo (changes requested):

artichoke-backend/src/extn/core/time/args.rs
b-n: Apart from some `mut` here and there, that's exactly correct!

artichoke-backend/src/extn/core/time/args.rs
b-n: Yes, good to know! It looks a whole lot better than a random new allocation hanging around.

artichoke-backend/src/extn/core/time/args.rs
b-n: You know what, I'm also confused as to why I put it there in the original code too :grimacing:. I assume `to_int` should never result in a `nil`, so easy fix!

lopopolo (approved):
lookgs good to me!

PR #2412, opened 2023-02-08
WIP - Attempt to convert args with `TryConvertMut`

b-n:
Context:

`Time::utc`,`Time::local` and friends takes variable amounts of arguments. These are usually year, month, day, hour, minute second - however they flip when there are 10 args, giving second, minute, hour, day month year (the rest are ignored). In addition to this, each parameter can be accessed in a fun variety of ways. Some more quirky ones:

```console
3.1.2 > class A; def to_int; 3; end; end
=> :to_int
3.1.2 > Time.utc(2023, A.new)
=> 2022-03-01 00:00:00 +0100
3.1.2 > Time.utc(2022, "feb")
=> 2022-02-01 00:00:00 UTC
3.1.2 > Time.utc(2022, "mar")
=> 2022-03-01 00:00:00 UTC
```

:point_up: This is tricky when the order of the parameters change. Additionally, `Time::new` supports other parameters which are not available in `Time::utc` and friends, however it still supports the same `RubyValue`s.

This PR is attempting to use `TryConvertMut` to coerce the `RubyValue`s of the parameters into the appropriate form. e.g. the order of the arguments is only relevant for ensuring the parameters are in the correct place, however after that point, the actual conversion of the `RubyValue` into a `Month` value should always be the same.

### DIFF
--- a/artichoke-backend/src/extn/core/time/args.rs
+++ b/artichoke-backend/src/extn/core/time/args.rs
@@ -1,76 +1,189 @@
 use crate::convert::to_int;
 use crate::extn::prelude::*;
 
-#[derive(Debug, Copy, Clone)]
+pub type Year = i32;
+pub type Month = u8;
+pub type Day = u8;
+pub type Hour = u8;
+pub type Minute = u8;
+pub type Second = u8;
+pub type Nanoseconds = u32;
+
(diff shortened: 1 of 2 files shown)


b-n:
Closing since #2411 has (IMO) a much better refactor than what I was attempting here.

PR #2413, opened 2023-02-18
Time Args support for month string values

b-n:
Another follow on from #2410 and #2411.
Related to #2223

`Time#utc`, `Time#local`, and friends all support specifying a month value from the shortened 3 character english names for the month.

Ref: https://ruby-doc.org/3.1.2/Time.html#method-c-utc

### DIFF
--- a/artichoke-backend/src/extn/core/time/args.rs
+++ b/artichoke-backend/src/extn/core/time/args.rs
@@ -1,4 +1,4 @@
-use crate::convert::to_int;
+use crate::convert::{to_int, to_str};
 use crate::extn::prelude::*;
 
 #[derive(Debug, Copy, Clone)]
@@ -57,18 +57,52 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
         for (i, &arg) in args.iter().enumerate() {
             match i {
                 0 => {
-                    let arg = to_int(self, arg)?;
-                    let arg: i64 = arg.try_convert_into(self)?;
+                    let arg: i64 = to_int(self, arg).and_then(|arg| arg.try_convert_into(self))?;
 
                     result.year = i32::try_from(arg).map_err(|_| ArgumentError::with_message("year out of range"))?;
                 }
                 1 => {
-                    // TODO: This should support 3 letter month names
-                    // as per the docs. https://ruby-doc.org/3.1.2/Time.html#method-c-new
-                    let arg = to_int(self, arg)?;
-                    let arg: i64 = arg.try_convert_into(self)?;
+                    // ```irb
+                    // 3.1.2 => Time.utc(2022, 2).month
+                    // => 2
+                    // 3.1.2 => class I; def to_int; 2; end; end
+                    // => :to_int
+                    // 3.1.2 => Time.utc(2022, I.new).month
+                    // => 2
+                    // 3.1.2 > Time.utc(2022, "feb").month
+                    // => 2
+                    // 3.1.2 > class A; def to_str; "feb"; end; end
+                    // => :to_str
+                    // 3.1.2 > Time.utc(2022, A.new).month
+                    // => 2
+                    // ```
+                    let month: i64 = if let Ok(month) = to_str(self, arg).and_then(|arg| {
+                        let month_str: &str = arg.try_convert_into_mut(self)?;
+                        match month_str {
+                            "jan" => Ok(1),
+                            "feb" => Ok(2),
+                            "mar" => Ok(3),
+                            "apr" => Ok(4),
+                            "may" => Ok(5),
+                            "jun" => Ok(6),
+                            "jul" => Ok(7),
+                            "aug" => Ok(8),
+                            "sep" => Ok(9),
+                            "oct" => Ok(10),
+                            "nov" => Ok(11),
+                            "dec" => Ok(12),
+                            // This error will never be seen due to the
+                            // fallback of to_int conversion below
+                            _ => Err(ArgumentError::with_message("unsupported mon value").into()),
+                        }
+                    }) {
+                        Ok(month)
+                    } else {
+                        let arg = to_int(self, arg)?;
+                        arg.try_convert_into(self)
+                    }?;
 
-                    result.month = match u8::try_from(arg) {
+                    result.month = match u8::try_from(month) {
                         Ok(month @ 1..=12) => Ok(month),
                         _ => Err(ArgumentError::with_message("mon out of range")),
                     }?;
@@ -298,6 +332,76 @@ mod tests {
         assert_eq!(0, result.minute);
         assert_eq!(0, result.second);
         assert_eq!(0, result.nanoseconds);
+    }
+
+    #[test]
+    fn month_supports_string_values() {
+        let mut interp = interpreter();
+
+        let table = [
+            (b"[2022, \"jan\"]", 1),
+            (b"[2022, \"feb\"]", 2),
+            (b"[2022, \"mar\"]", 3),
+            (b"[2022, \"apr\"]", 4),
+            (b"[2022, \"may\"]", 5),
+            (b"[2022, \"jun\"]", 6),
+            (b"[2022, \"jul\"]", 7),
+            (b"[2022, \"aug\"]", 8),
+            (b"[2022, \"sep\"]", 9),
+            (b"[2022, \"oct\"]", 10),
+            (b"[2022, \"nov\"]", 11),
+            (b"[2022, \"dec\"]", 12),
+        ];
+
+        for (input, expected_month) in table {
+            let args = interp.eval(input).unwrap();
+            let mut ary_args: Vec<Value> = interp.try_convert_mut(args).unwrap();
+            let result: Args = interp.try_convert_mut(ary_args.as_mut_slice()).unwrap();
+
+            assert_eq!(expected_month, result.month);
+        }
+    }
+
+    #[test]
+    fn month_supports_string_like_values() {
+        let mut interp = interpreter();
+
+        let args = interp
+            .eval(b"class A; def to_str; \"feb\"; end; end; [2022, A.new]")
+            .unwrap();
+        let mut ary_args: Vec<Value> = interp.try_convert_mut(args).unwrap();
+        let result: Args = interp.try_convert_mut(ary_args.as_mut_slice()).unwrap();
+
+        assert_eq!(2, result.month);
+    }
+
+    #[test]
+    fn month_supports_int_like_values() {
+        let mut interp = interpreter();
+
+        let args = interp.eval(b"class A; def to_int; 2; end; end; [2022, A.new]").unwrap();
+        let mut ary_args: Vec<Value> = interp.try_convert_mut(args).unwrap();
+        let result: Args = interp.try_convert_mut(ary_args.as_mut_slice()).unwrap();
+
+        assert_eq!(2, result.month);
+    }
+
+    #[test]
+    fn invalid_month_string_responds_with_int_conversion_error() {
+        let mut interp = interpreter();
+
+        let args = interp
+            .eval(b"class A; def to_str; \"aaa\"; end; end; [2022, A.new]")
+            .unwrap();
+        let mut ary_args: Vec<Value> = interp.try_convert_mut(args).unwrap();
+        let result: Result<Args, Error> = interp.try_convert_mut(ary_args.as_mut_slice());
+        let error = result.unwrap_err();
+
+        assert_eq!(
+            error.message().as_bstr(),
+            b"no implicit conversion of A into Integer".as_bstr()
+        );
+        assert_eq!(error.name(), "TypeError");
     }
 
     #[test]
--- a/artichoke-backend/src/extn/core/time/args.rs
+++ b/artichoke-backend/src/extn/core/time/args.rs
@@ -339,18 +339,19 @@ mod tests {
         let mut interp = interpreter();
 
         let table = [
-            (b"[2022, \"jan\"]", 1),
-            (b"[2022, \"feb\"]", 2),
-            (b"[2022, \"mar\"]", 3),
-            (b"[2022, \"apr\"]", 4),
-            (b"[2022, \"may\"]", 5),
-            (b"[2022, \"jun\"]", 6),
-            (b"[2022, \"jul\"]", 7),
-            (b"[2022, \"aug\"]", 8),
-            (b"[2022, \"sep\"]", 9),
-            (b"[2022, \"oct\"]", 10),
-            (b"[2022, \"nov\"]", 11),
-            (b"[2022, \"dec\"]", 12),
+            (b"[2022, 'jan']", 1),
+            (b"[2022, 'feb']", 2),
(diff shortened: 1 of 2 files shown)


artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -63,12 +63,47 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
...
+                            // This error will never be seen due to the
+                            // fallback of to_int conversion below
+                            _ => Err(ArgumentError::with_message("unsupported mon value").into()),
+                        }
+                    }) {
+                        Ok(month)
+                    } else {
+                        let arg = to_int(self, arg)?;
+                        arg.try_convert_into(self)
+                    }?;
```
b-n: I spent a little time puzzling over the nicest way to do this. It could be split up a little instead of chaining the `and_then` into the `if let`, however I think the logic still is "correct".

It's also very tempting to short circuit this with a type check on `Ruby::FixNum` since it's very likely the majority of the usage.

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -57,18 +57,52 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
...
+                            "mar" => Ok(3),
+                            "apr" => Ok(4),
+                            "may" => Ok(5),
+                            "jun" => Ok(6),
+                            "jul" => Ok(7),
+                            "aug" => Ok(8),
+                            "sep" => Ok(9),
+                            "oct" => Ok(10),
+                            "nov" => Ok(11),
+                            "dec" => Ok(12),
```
lopopolo: Per the docs:

> A 3-character string that matches regular expression `/jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec/i`.

which seems to indicate this string is case insensitive. testing this out seems to confirm that:

```console
[3.1.2] > Time.utc(2022, "FEB").month
=> 2
[3.1.2] > Time.utc(2022, "FeB").month
=> 2
[3.1.2] > Time.utc(2022, "Feb").month
=> 2
[3.1.2] > Time.utc(2022, "fEB").month
=> 2
```

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -57,18 +57,52 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
...
+                    // 3.1.2 => Time.utc(2022, I.new).month
+                    // => 2
+                    // 3.1.2 > Time.utc(2022, "feb").month
+                    // => 2
+                    // 3.1.2 > class A; def to_str; "feb"; end; end
+                    // => :to_str
+                    // 3.1.2 > Time.utc(2022, A.new).month
+                    // => 2
+                    // ```
+                    let month: i64 = if let Ok(month) = to_str(self, arg).and_then(|arg| {
```
lopopolo: this is a bit too intense in the `if let` clause for me with the `and_then` block. is there a way to restructure this code so it is a bit more imperative?

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -57,18 +57,52 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
...
+                        let month_str: &str = arg.try_convert_into_mut(self)?;
```
lopopolo: since we're checking against an ASCII allowlist, we can use `&[u8]` and `b"jan"` byte literals here to avoid a UTF-8 decode.

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -300,6 +334,76 @@ mod tests {
...
+            (b"[2022, \"apr\"]", 4),
+            (b"[2022, \"may\"]", 5),
+            (b"[2022, \"jun\"]", 6),
+            (b"[2022, \"jul\"]", 7),
+            (b"[2022, \"aug\"]", 8),
+            (b"[2022, \"sep\"]", 9),
+            (b"[2022, \"oct\"]", 10),
+            (b"[2022, \"nov\"]", 11),
+            (b"[2022, \"dec\"]", 12),
+        ];
```
lopopolo:
```suggestion
        let table = [
            (br#"[2022, "jan"]"#, 1),
            (br#"[2022, "feb"]"#, 2),
            (br#"[2022, "mar"]"#, 3),
            (br#"[2022, "apr"]"#, 4),
            (br#"[2022, "may"]"#, 5),
            (br#"[2022, "jun"]"#, 6),
            (br#"[2022, "jul"]"#, 7),
            (br#"[2022, "aug"]"#, 8),
            (br#"[2022, "sep"]"#, 9),
            (br#"[2022, "oct"]"#, 10),
            (br#"[2022, "nov"]"#, 11),
            (br#"[2022, "dec"]"#, 12),
        ];
```

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -300,6 +334,76 @@ mod tests {
         assert_eq!(0, result.nanoseconds);
     }
 
+    #[test]
+    fn month_supports_string_values() {
+        let mut interp = interpreter();
+
+        let table = [
```
lopopolo, line 358: we'll need to add some mixed case and uppercase test cases to this table.

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -300,6 +334,76 @@ mod tests {
...
+            assert_eq!(expected_month, result.month);
+        }
+    }
+
+    #[test]
+    fn month_supports_string_like_values() {
+        let mut interp = interpreter();
+
+        let args = interp
+            .eval(b"class A; def to_str; \"feb\"; end; end; [2022, A.new]")
```
lopopolo: we can make the ruby code look like Ruby code here by using different quotes.

```suggestion
            .eval(b"class A; def to_str; 'feb'; end; end; [2022, A.new]")
```

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -300,6 +334,76 @@ mod tests {
...
+
+        assert_eq!(2, result.month);
+    }
+
+    #[test]
+    fn invalid_month_string_responds_with_int_conversion_error() {
+        let mut interp = interpreter();
+
+        let args = interp
+            .eval(b"class A; def to_str; \"aaa\"; end; end; [2022, A.new]")
```
lopopolo:
```suggestion
            .eval(b"class A; def to_str; 'aaa'; end; end; [2022, A.new]")
```

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -300,6 +334,76 @@ mod tests {
...
+
+        let args = interp.eval(b"class A; def to_int; 2; end; end; [2022, A.new]").unwrap();
+        let mut ary_args: Vec<Value> = interp.try_convert_mut(args).unwrap();
+        let result: Args = interp.try_convert_mut(ary_args.as_mut_slice()).unwrap();
+
+        assert_eq!(2, result.month);
+    }
+
+    #[test]
+    fn invalid_month_string_responds_with_int_conversion_error() {
```
lopopolo, line 444: ❤️

lopopolo (changes requested):
love that we're getting so close to parity here. I find the structure of the `if let` hard to follow. It reminds me of this pattern in golang which I also find hard to read:

```go
if err := func() {
    // do
    // the
    // thing
    return err
    // do
    // do
    // do
    return nil
}(); err != nil {
    return nil
}
```

I think the simplest way to do the case insensitivity is with an allocation to `Vec<u8>` combined with [`BStr::make_ascii_lowercase`](https://docs.rs/bstr/latest/bstr/trait.ByteSlice.html#method.make_ascii_lowercase). Maybe we can short circuit if the length of the byteslice != 3?

artichoke-backend/src/extn/core/time/args.rs
b-n: I remember reading and swiftly ignoring the `/i` part of that regex since I was so focused on the values :sweat_smile:. Agree with the other comment though, comparing lowercase should fix it!

artichoke-backend/src/extn/core/time/args.rs
b-n: I'll give it a whack! I was trying to DRY it too much by relying on the same `to_int` logic since failing the string match ends with a numeric match on the string. Speaking of which, I need to add a test case:

```console
3.1.2 > class A; def to_str; "2"; end; end
=> :to_str
3.1.2 > Time.utc(2022, A.new)
=> 2022-02-01 00:00:00 UTC
```

Note: I think this would break the idea of a shortcircuit on the byteslice length :disappointed:

artichoke-backend/src/extn/core/time/args.rs
b-n: `%s/\\\"/\'/g` :-D

artichoke-backend/src/extn/core/time/args.rs
b-n: `make_ascii_lowercase()` - which nicely does convert in place - needed a `&mut [u8]` conversion for Artichoke (which doesn't exist currently). I assume that's for a reason rather than adding my own - so the only way this was possible was to make it convert to a Vec<u8>, and do the comparison of byte slices with `as_slice()`.

Is there any particular reason there isn't a conversion to `&mut [u8]` as of yet?

artichoke-backend/src/extn/core/time/args.rs
b-n: Added as another test

artichoke-backend/src/extn/core/time/args.rs
b-n: Except I had the error incorrect :grimacing:. But it's accurate now!

artichoke-backend/src/extn/core/time/args.rs
b-n: Note:

```rust
impl<'a> TryConvertMut<Value, &'a mut [u8]> for Artichoke {
    type Error = Error;

    fn try_convert_mut(&mut self, mut value: Value) -> Result<&'a mut [u8], Self::Error> {
        self.protect(value);
        let mut s = unsafe { String::unbox_from_value(&mut value, self)? };
        // SAFETY: This transmute modifies the lifetime of the byte slice pulled
        // out of the boxed `String`. This requires that no garbage collections
        // that reclaim `value` occur while this slice is alive. This is
        // enforced for at least this entry from an mruby trampoline by the call
        // to `protect` above.
        //
        // FIXME: does this unbound lifetime and transmute below allow
        // extracting `&'static [u8]`?
        let slice = unsafe { mem::transmute(s.as_inner_mut().as_mut_slice()) };
        Ok(slice)
    }
}
```

appears to be "valid rust". I am very okay in admitting that I'm not entirely sure if this makes 100% rust/memory sense (still a topic I need to get a properly deep understanding of)

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -57,18 +58,64 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
...
+                    // 3.1.2 > class A; def to_str; "feb"; end; end
+                    // => :to_str
+                    // 3.1.2 > Time.utc(2022, A.new).month
+                    // => 2
+                    // 3.1.2 > class I; def to_str; "2"; end; end
+                    // => :to_str
+                    // 3.1.2 > Time.utc(2022, I.new).month
+                    // => 2
+                    // ```
+                    let month: i64 = if Ruby::Fixnum == arg.ruby_type() {
```
lopopolo: if you want to save some rightward drift, we can make one of the arms in the match:

```rust
                 1 if arg.ruby_type() == Ruby::Fixnum => {
                    // Short circuit to avoid string checking
                     let arg = to_int(self, arg)?;
                     arg.try_convert_into(self)?
                 }
```

Consider this suggestion non-blocking.

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -57,18 +58,64 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
...
+                                    // Delegate to `Kernel#Integer` as last
+                                    // resort to handle Integer strings.
+                                    let arg = integer(self, arg, None)?;
+                                    arg.try_convert_into(self)
+                                }
+                            }
+                        } else {
+                            let arg = to_int(self, arg)?;
+                            arg.try_convert_into(self)
+                        }?
```
lopopolo: most elsewhere we prefer having if/else or if let/else expressions return a bare type with each branch making sure to use `?` to return errors. This would avoid all of the `Ok` wrapping in the happy path above. Keeping the `?` local to the fallible operations makes it a bit easier to see which operations are indeed fallible (as opposed to highlighting those that aren't with `Ok` wrapping).

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -57,18 +58,64 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
...
+                    };
+
+                    result.month = match u8::try_from(month) {
                         Ok(month @ 1..=12) => Ok(month),
                         _ => Err(ArgumentError::with_message("mon out of range")),
```
lopopolo: similarly, this and other `match`es in this file would be structured:

```suggestion
                        Ok(month @ 1..=12) => month,
                        _ => return Err(ArgumentError::with_message("mon out of range").into()),
```

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -300,6 +347,112 @@ mod tests {
...
+            (b"[2022, 'fEB']", 2),
+            (b"[2022, 'FEB']", 2),
+        ];
+
+        for (input, expected_month) in table {
+            let args = interp.eval(input).unwrap();
+            let mut ary_args: Vec<Value> = interp.try_convert_mut(args).unwrap();
+            let result: Args = interp.try_convert_mut(ary_args.as_mut_slice()).unwrap();
+
+            assert_eq!(expected_month, result.month);
```
lopopolo, line 401: it seems cheap to add an assert on `result.year` too if you think it might be valuable.

lopopolo (approved):
I left some style suggestions which you don't need to take. Logic LGTM.

artichoke-backend/src/extn/core/time/args.rs
b-n: IMO, this is good to add. You never know if someone tries `14` expecting it to wrap and removing other conditions etc!

artichoke-backend/src/extn/core/time/args.rs
b-n: Genius!

artichoke-backend/src/extn/core/time/args.rs
b-n: Ah yes! I was fumbling a bit with types here, but this totally makes sense. I know you've said this before too, but this time it's going to stick!

lopopolo (approved):

PR #2415, opened 2023-02-19
Remove `TryConvertMut` fully-qualified turbofishes

lopopolo:
- Simplify a trampoline by using `Value::try_convert_into` in an `if let` condition.
- Reimplement `TryConvertMut<Value, String>` to go through the `Vec<u8>` owned impl, instead of the `&str` -> `&[u8]` borrowed impl. This saves a stack frame and uses a direct clone of the underlying `Vec<u8>`.

Thanks for poking me about this and getting me to look at this old cold again @b-n!


lopopolo:
@b-n should I interpret your 🚀 reaction as an approval?

b-n (approved):
Here is a clearer approval! :rocket:

PR #2417, opened 2023-02-19
Use short-circuit return instead of Ok wrapping in Time args

lopopolo:
Followup to #2413.

### DIFF
--- a/artichoke-backend/src/extn/core/time/args.rs
+++ b/artichoke-backend/src/extn/core/time/args.rs
@@ -129,27 +129,27 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
                     let arg: i64 = arg.try_convert_into(self)?;
 
                     result.day = match u8::try_from(arg) {
-                        Ok(day @ 1..=31) => Ok(day),
-                        _ => Err(ArgumentError::with_message("mday out of range")),
-                    }?;
+                        Ok(day @ 1..=31) => day,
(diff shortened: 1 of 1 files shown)


b-n (approved):
:sweat_smile: 

Now I'm wondering if there could ever be a lint rule for this!

artichoke-backend/src/extn/core/time/args.rs
```diff
@@ -174,9 +174,9 @@ impl TryConvertMut<&mut [Value], Args> for Artichoke {
                     // therefore we must multiply the value by 1_000. This is
                     // gaurnateed to fit in a u32.
                     result.nanoseconds = match u32::try_from(arg) {
-                        Ok(micros @ 0..=999_999) => Ok(micros * 1000),
-                        _ => Err(ArgumentError::with_message("subsecx out of range")),
-                    }?;
+                        Ok(micros @ 0..=999_999) => micros * 1000,
+                        _ => return Err(ArgumentError::with_message("subsecx out of range").into()),
```
lopopolo, line 182: this looked like a typo to me but it isn't. Adding a comment for clarity:

```suggestion
                        // ```
                        // [3.1.2] > Time.utc(2022, 12, 31, 13, 24, 55, 100000000000000000)
                        // (irb):3:in `utc': subsecx out of range (ArgumentError)
                        // ```
                        _ => return Err(ArgumentError::with_message("subsecx out of range").into()),
```

Issue #2361, opened 2023-01-28
Add `byteindex` and `byterindex` methods to `spinoso-string`


lopopolo:
The existing `spinoso_string::String::index` and `spinoso_string::String::rindex` methods return byte indexes, but should operate on characters.

Expose new methods `spinoso_string::String::byteindex` and `spinoso_string::String::byterindex` which preserve the existing behavior. These APIs match up with these in Ruby's `String`:

- https://ruby-doc.org/3.2.0/String.html#method-i-byteindex
- https://ruby-doc.org/3.2.0/String.html#method-i-byterindex

As in #2360, these new methods could use some tests which use multibyte UTF-8 strings.

See:

- https://github.com/artichoke/artichoke/issues/2360

PR #2450, opened 2023-03-21
`String#index` and `String#rindex` should return character offsets instead of byte offsets

### LINKED ISSUE #323: Implement String#index with bstr for String patterns

### LINKED ISSUE #2360: `String#index` and `String#rindex` should return character offsets instead of byte offsets
`spinoso_string::String:index` and `spinoso_string::String::rindex` return byte offsets. Modify these methods in `spinoso-string` to return character offsets which respect the string's underlying encoding.

This area could use some additional tests with multi-byte UTF-8 strings.

See:

- https://github.com/artichoke/artichoke/issues/2361

### Artichoke

```console
$ cargo run --bin artichoke
artichoke 0.1.0-pre.0 (2022-12-16 revision 6441) [x86_64-apple-darwin]
[rustc 1.66.0 (69f9c33d7 2022-12-12) on x86_64-apple-darwin]
>>> "via 💎 v3.1.2".index "2"
=> 14
```

### MRI

```console
$ irb
[3.1.2] > "via 💎 v3.1.2".index "2"
=> 11
```

choznerol:
Close #2360 
Close https://github.com/artichoke/artichoke/issues/323

~The branch of this PR is based on #2361 to avoid conflict.~

### DIFF
--- a/spinoso-string/src/enc/ascii/mod.rs
+++ b/spinoso-string/src/enc/ascii/mod.rs
@@ -356,7 +356,8 @@ impl AsciiString {
     #[inline]
     #[must_use]
     pub fn rindex(&self, needle: &[u8], offset: usize) -> Option<usize> {
-        let buf = self.get(..=offset)?;
+        let search_until = (offset + 1).min(self.len());
+        let buf = self.get(..search_until)?;
         let index = buf.rfind(needle)?;
         Some(index)
     }
@@ -596,6 +597,7 @@ mod tests {
     #[test]
     fn rindex_with_different_offset() {
         let s = AsciiString::from(b"foo");
+        assert_eq!(s.rindex("o".as_bytes(), 3), Some(2));
         assert_eq!(s.rindex("o".as_bytes(), 2), Some(2));
         assert_eq!(s.rindex("o".as_bytes(), 1), Some(1));
         assert_eq!(s.rindex("o".as_bytes(), 0), None);
--- a/spinoso-string/src/enc/ascii/mod.rs
+++ b/spinoso-string/src/enc/ascii/mod.rs
@@ -343,6 +343,25 @@ impl AsciiString {
     }
 }
 
+// Index
+impl AsciiString {
+    #[inline]
+    #[must_use]
+    pub fn index(&self, needle: &[u8], offset: usize) -> Option<usize> {
+        let buf = self.get(offset..)?;
+        let index = buf.find(needle)?;
+        Some(index + offset)
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn rindex(&self, needle: &[u8], offset: usize) -> Option<usize> {
+        let buf = self.get(..=offset)?;
+        let index = buf.rfind(needle)?;
+        Some(index)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use alloc::string::String;
@@ -546,5 +565,39 @@ mod tests {
         assert_eq!(s.get_char_slice(4..1), None);
         assert_eq!(s.get_char_slice(3..1), Some(&b""[..]));
         assert_eq!(s.get_char_slice(2..1), Some(&b""[..]));
+    }
+
+    #[test]
+    fn index_with_default_offset() {
+        let s = AsciiString::from(b"foo");
+        assert_eq!(s.index("f".as_bytes(), 0), Some(0));
+        assert_eq!(s.index("o".as_bytes(), 0), Some(1));
+        assert_eq!(s.index("oo".as_bytes(), 0), Some(1));
+        assert_eq!(s.index("ooo".as_bytes(), 0), None);
+    }
+
+    #[test]
+    fn index_with_different_offset() {
+        let s = AsciiString::from(b"foo");
+        assert_eq!(s.index("o".as_bytes(), 1), Some(1));
+        assert_eq!(s.index("o".as_bytes(), 2), Some(2));
+        assert_eq!(s.index("o".as_bytes(), 3), None);
+    }
+
+    #[test]
+    fn rindex_with_default_offset() {
+        let s = AsciiString::from(b"foo");
+        assert_eq!(s.rindex("f".as_bytes(), 2), Some(0));
+        assert_eq!(s.rindex("o".as_bytes(), 2), Some(2));
+        assert_eq!(s.rindex("oo".as_bytes(), 2), Some(1));
+        assert_eq!(s.rindex("ooo".as_bytes(), 2), None);
+    }
+
+    #[test]
+    fn rindex_with_different_offset() {
+        let s = AsciiString::from(b"foo");
+        assert_eq!(s.rindex("o".as_bytes(), 2), Some(2));
+        assert_eq!(s.rindex("o".as_bytes(), 1), Some(1));
+        assert_eq!(s.rindex("o".as_bytes(), 0), None);
     }
 }
--- a/spinoso-string/src/enc/utf8/mod.rs
+++ b/spinoso-string/src/enc/utf8/mod.rs
@@ -1274,6 +1274,7 @@ mod tests {
     #[test]
     fn rindex_with_different_offset() {
         let s = Utf8String::from("f💎oo");
+        assert_eq!(s.rindex("o".as_bytes(), 4), Some(3));
         assert_eq!(s.rindex("o".as_bytes(), 3), Some(3));
         assert_eq!(s.rindex("o".as_bytes(), 2), Some(2));
         assert_eq!(s.rindex("o".as_bytes(), 1), None);
--- a/spinoso-string/src/enc/utf8/mod.rs
+++ b/spinoso-string/src/enc/utf8/mod.rs
@@ -1280,4 +1280,28 @@ mod tests {
         assert_eq!(s.rindex("o".as_bytes(), 1), None);
         assert_eq!(s.rindex("o".as_bytes(), 0), None);
     }
+
+    #[test]
+    fn index_and_rindex_support_invalid_utf8_in_needle() {
+        // Invalid UTF-8 in needle
+        let needle = &"💎".as_bytes()[..3];
+
+        println!("{:?} rfinx {:?}", "f💎oo".as_bytes(), needle);
+
+        assert_eq!(Utf8String::from("f💎oo").index(needle, 0), None); // FIXME: Currently `Some(1)`
+        assert_eq!(Utf8String::from("f💎oo").rindex(needle, 3), None); // FIXME: Currently `Some(1)`
+    }
+
+    #[test]
+    fn index_and_rindex_support_invalid_utf8_in_haystack() {
+        // Invalid UTF-8 in haystack
+        let mut haystack = Vec::new();
+        haystack.extend_from_slice(b"f");
+        haystack.extend_from_slice(&"💎".as_bytes()[..2]);
+        haystack.extend_from_slice(b"oo");
+        let haystack = Utf8String::from(haystack);
+
+        assert_eq!(haystack.index("💎".as_bytes(), 0), None);
+        assert_eq!(haystack.rindex("💎".as_bytes(), 3), None);
+    }
 }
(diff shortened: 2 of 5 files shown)


lopopolo:
@choznerol let us know when this one is ready for review (I notice it's still in Draft)

choznerol:
> @choznerol let us know when this one is ready for review (I notice it's still in Draft)

Okay. Will let you know when I do. Thanks!

spinoso-string/src/lib.rs
```diff
@@ -1649,18 +1651,19 @@ impl String {
...
-                // a valid index of the underlying `Vec`.
-                //
-                // `self.buf.len() < isize::MAX` because `self.buf` is a `Vec` and
-                // `Vec` documents `isize::MAX` as its maximum allocation size.
-                Some(index + offset)
-            } else {
-                buf.find(needle)
+            let offset = offset.unwrap_or(0);
+            let buf = std::str::from_utf8(buf).ok()?;
+            let needle = std::str::from_utf8(needle).ok()?;
```
lopopolo: one thing I noticed here is that this requires strings to be valid UTF-8 which may not be true for strings of _any_ encoding.

I think this method is more appropriately delegated down through `EncodedString` to the individual string variants. Take a peek at `String::get_char_slice` to see how to implement something like that.

choznerol: Thanks! I did have the question in mind about when to implement things in encoded variants, but I think I get it now. Will let you know when I finish refactoring both methods and pass CI.

choznerol: Hi @lopopolo I've refactored most encoding-aware logic down to the encoded variants, so it's ready for review :)

The `enc::ascii::(r)index` and `enc::binary::(r)index` turn out to be quite duplicated, but it seems to be common across many `(Ascii|Binary)String` methods. Just want to confirm if there is anything I should do in this PR to DRY them?

lopopolo: yup that's right, no DRYing up needed. that's the way I was expecting it to look.

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -804,6 +804,33 @@ impl Utf8String {
     }
 }
 
+// Index
+impl Utf8String {
+    #[inline]
+    #[must_use]
+    pub fn index(&self, needle: &[u8], offset: usize) -> Option<usize> {
+        for byte_pos in self.find_iter(needle) {
+            let char_pos = self[..byte_pos].chars().count();
```
lopopolo: this does an unchecked index into a byte slice (ok) and calls `bstr::ByteSlice::chars()` on the result. `chars` replaces invalid UTF-8 byte sequences with the Unicode replacement character `�`. This means all of these cases return incorrect indexes:

- `Utf8String`s which contain invalid UTF-8 byte sequences.
- Needles which contain invalid UTF-8 byte sequences that match a prefix or suffix of a UTF-8 character in the `Utf8String`.

`Utf8String::index` and `Utf8String::rindex` should use the `bstr::decode_utf8` or `bstr::decode_last_utf8` functions to do a "character"-at-a-time decoding. These routines allow properly detecting whether an invalid UTF-8 byte sequence has been encountered and properly juggling the indexes.

See this example in `Utf8String::get_char` for how to use these APIs:

https://github.com/artichoke/artichoke/blob/b49699f43241f24b5781efe7de3f225b16c297ac/spinoso-string/src/enc/utf8/mod.rs#L237-L281

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -1216,4 +1243,40 @@ mod tests {
...
+        assert_eq!(s.index("f".as_bytes(), 0), Some(0));
+        assert_eq!(s.index("o".as_bytes(), 0), Some(2));
+        assert_eq!(s.index("oo".as_bytes(), 0), Some(2));
+        assert_eq!(s.index("ooo".as_bytes(), 0), None);
+    }
+
+    #[test]
+    fn index_with_different_offset() {
+        let s = Utf8String::from("f💎oo");
+        assert_eq!(s.index("o".as_bytes(), 1), Some(2));
```
lopopolo, line 1324:
```console
[3.1.2] > "💎".unpack "C*"
=> [240, 159, 146, 142]
```

I suspect adding some test cases with needles as various subslices of the UTF-8 encoding for `💎` will reveal the bugs in relying on `ByteSlice::chars` above.

spinoso-string/src/lib.rs
```diff
@@ -1640,32 +1640,20 @@ impl String {
...
     /// assert_eq!(s.index("💎", None), Some(4));
-    /// assert_eq!(s.index(".", None), Some(11)); // FIXME: Should be 8 (#2360)
+    /// assert_eq!(s.index(".", None), Some(8));
+    /// assert_eq!(s.index("v", Some(6)), Some(6));
+    /// assert_eq!(s.index("v", Some(7)), None);
     /// assert_eq!(s.index("X", None), None);
     /// ```
     ///
+    /// Related: String#rindex.
+    ///
```
lopopolo: please remove this

```suggestion
```

spinoso-string/src/lib.rs
```diff
@@ -1682,27 +1670,22 @@ impl String {
...
-        fn inner(buf: &[u8], needle: &[u8], offset: Option<usize>) -> Option<usize> {
-            if let Some(offset) = offset {
-                let end = buf.len().checked_sub(offset).unwrap_or_default();
-                let buf = buf.get(..end)?;
-                buf.rfind(needle)
-            } else {
-                buf.rfind(needle)
-            }
+        let default_max_offset = self.inner.char_len() - 1;
+        let mut offset = offset.unwrap_or(default_max_offset);
```
lopopolo: `char_len()` is potentially an expensive operation since it must do a full (UTF-8) decode of the string. Please move it to be lazily evaluated in:

```suggestion
        let mut offset = offset.unwrap_or_else(|| self.inner.char_len().checked_sub(1).unwrap_or_default());
```

I would have expected this underflow on an empty string to have resulted in a panic in tests. Can we add some tests for both empty receivers and empty needles?

lopopolo (changes requested):

lopopolo:
I've additionally linked this old issue which should be resolved by this PR:

- https://github.com/artichoke/artichoke/issues/323

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -809,10 +809,37 @@ impl Utf8String {
...
     pub fn index(&self, needle: &[u8], offset: usize) -> Option<usize> {
-        for byte_pos in self.find_iter(needle) {
-            let char_pos = self[..byte_pos].chars().count();
-            if char_pos >= offset {
-                return Some(char_pos);
+        // Decode needle
+        let needle_chars = match must_decode_utf8(needle) {
+            Some(chars) => chars,
+            // Needle containing any invalid UTF-8 should never match in MRI
+            None => return None,
```
choznerol, line 816: According to `irb`:
image

spinoso-string/src/lib.rs
choznerol: Thanks for pointing out. The underflow indeed can be reproduced with 1847a6a1b2e15dd171b75b2413bdebd179e8b576.

Fixed in 4cd5c98add24b9945b28627b77c69c05ff174250.

spinoso-string/src/enc/utf8/mod.rs
choznerol: Thanks! Indeed, I can reproduce the issue in 751be01.

spinoso-string/src/enc/utf8/mod.rs
choznerol: Thanks! I updated both `enc::utf8::index` and `enc::utf8::rindex` to use `bstr::decode_utf8` in b1d69c7.

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -804,6 +804,98 @@ impl Utf8String {
...
 
+// Index
+impl Utf8String {
+    #[inline]
+    #[must_use]
+    pub fn index(&self, needle: &[u8], offset: usize) -> Option<usize> {
+        // Decode needle
+        let needle_chars = match must_decode_utf8(needle) {
+            Some(chars) => chars,
+            // Needle containing any invalid UTF-8 should never match in MRI
```
choznerol, line 815: According to `irb`:
image

lopopolo: sweet. I played with this a bit too:

```console
$ irb
[3.2.2] > s = "abc\xFF"
=> "abc\xFF"
[3.2.2] > s.encoding
=> #<Encoding:UTF-8>
[3.2.2] > s.index s
=> nil
[3.2.2] > s.rindex s
=> nil
[3.2.2] > s.index "abc"
=> 0
[3.2.2] > s.rindex "abc"
=> 0
[3.2.2] > s.index "bc\xFF"
=> nil
[3.2.2] > s.rindex "bc\xFF"
=> nil
[3.2.2] > s.index "bc"
=> 1
[3.2.2] > s.rindex "bc"
=> 1
```

spinoso-string/src/enc/ascii/mod.rs
```diff
@@ -343,6 +343,26 @@ impl AsciiString {
...
+    pub fn index(&self, needle: &[u8], offset: usize) -> Option<usize> {
+        let buf = self.get(offset..)?;
+        let index = buf.find(needle)?;
+        Some(index + offset)
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn rindex(&self, needle: &[u8], offset: usize) -> Option<usize> {
+        let search_until = (offset + 1).min(self.len());
```
lopopolo, line 359: we should use a saturating add on `offset` here to avoid an overflow.

spinoso-string/src/enc/ascii/mod.rs
```diff
@@ -343,6 +343,26 @@ impl AsciiString {
...
+        let buf = self.get(..search_until)?;
```
lopopolo, line 360: this can be simplified to avoid a branch. combined with the above overflow comment, I think this works:

```suggestion
        let buf = self.get(..=search_until).unwrap_or_else(|| self.get(..));
```

lopopolo: this suggestion might need a special case and `#[test]` for when `buf` is empty.

lopopolo:
```console
[3.2.2] > "".rindex ""
=> 0
[3.2.2] > "".rindex "a"
=> nil
[3.2.2] > "a".rindex ""
=> 1
```

spinoso-string/src/enc/utf8/mod.rs
```diff
@@ -804,6 +804,98 @@ impl Utf8String {
     }
 }
 
+// Index
+impl Utf8String {
+    #[inline]
+    #[must_use]
+    pub fn index(&self, needle: &[u8], offset: usize) -> Option<usize> {
+        // Decode needle
+        let needle_chars = match must_decode_utf8(needle) {
```
lopopolo, line 813: I'll followup with a subsequent PR to see if we can do this without allocating

lopopolo (approved):
Looks good to me. I left a few comments, but I'm comfortable addressing them in a followup. Let's get this merged.

spinoso-string/src/enc/utf8/mod.rs
lopopolo: See https://github.com/artichoke/artichoke/pull/2554

spinoso-string/src/enc/ascii/mod.rs
lopopolo: I added some tests in https://github.com/artichoke/artichoke/pull/2554 which caught a few underflows.

spinoso-string/src/enc/ascii/mod.rs
lopopolo: I added some tests in https://github.com/artichoke/artichoke/pull/2554 and fixed this.

lopopolo:
@choznerol I landed those followups I mentioned in the comments here:

- https://github.com/artichoke/artichoke/pull/2554

PR #2497, opened 2023-04-10
Add `RB_FIXABLE` macro impl in a new crate

### LINKED ISSUE #2311: Build Rust crate for truncating floats to ints aka `RB_FIXABLE`
With the recent stabilization of [`Duration::try_from_secs_f64`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.try_from_secs_f64) in Rust 1.66.0, there is now an API in `std` that can losslessly truncate an `f64` to a `u64` like this:

```rust
use core::time::Duration::try_from_secs_f64;
use std::error::Error;

pub fn float_to_fix(f: f64) -> Result<u64, Box<dyn Error>> {
    let d = Duration::try_from_secs_f64(f).or_else(|| Duration::try_from_secs_f64(-f))?;
    let fix: u32 = d.as_secs();
    fix
}
```

This routine will fail if:

- `f` is `NAN`.
- `f` is `INF` or `NEG_INF`.
- `f` has magnitude greater than `u64::MAX`.

This routine returns the integral part of the float if it is in `u64` range.

Create a crate in the Artichoke organization that exposes this single function.

See the `RB_FIXABLE` macro in MRI for more details: https://github.com/ruby/ruby/blob/v3_1_3/include/ruby/internal/arithmetic/fixnum.h#L52.

lopopolo:
Add a new crate, `scolapasta-fixable` which can test and convert integers to "fixable" range.

Closes #2311.

This PR also includes some opportunistic fixes to tests and READMEs in other crates that were discovered while working on this branch.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -296,6 +296,14 @@ jobs:
         run: cargo check --verbose --all-features --all-targets --profile=test
         working-directory: "scolapasta-aref"
 
+      - name: Compile scolapasta-fixable with no default features
+        run: cargo check --verbose --no-default-features --all-targets --profile=test
+        working-directory: "scolapasta-fixable"
+
+      - name: Compile scolapasta-fixable with all features
+        run: cargo check --verbose --all-features --all-targets --profile=test
+        working-directory: "scolapasta-fixable"
+
       - name: Compile scolapasta-hex with no default features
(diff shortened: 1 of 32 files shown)


lopopolo:
markdown link check errors are the same flakes from #2498, ignoring these

PR #2498, opened 2023-04-10
Reduce scope of ignore list for markdown URLs

lopopolo:
Restrict ignore list for markdown URLs in markdown-link-check config to only ignore workspace crates in docs.rs and crates.io. All workspace crates have 'artichoke', 'mezzaluna', 'scolapasta', or 'spinoso' prefixes.


lopopolo:
failures are crates.io links for onig and cc. these pass locally, chalking this up to GitHub Actions runner config.

PR #2506, opened 2023-04-17
Split class and module spec registry data structure into a new mezzaluna crate

lopopolo:
This removes generic code from `artichoke-backend` and removes duplication between class and module specs.


lopopolo:
markdown link error is the GitHub Pages-hosted `mezzaluna-type-registry` rustdocs, which makes sense since the code hasn't been merged yet.

```
  [✖] https://artichoke.github.io/artichoke/mezzaluna_type_registry/ → Status: 404
```

PR #2517, opened 2023-04-25
Upgrade vendored emscripten headers to emsdk 3.1.22

lopopolo:
633d7851f01315d39a54b41edf8ad2eb68936aa7 addresses a assertion failure on 32-bit targets that was an oversight missed in https://github.com/artichoke/artichoke/pull/2515.


lopopolo:
verified the playground builds for me locally as of df355a05c80d026db6e84267522ae262678d0726. this is good to go

PR #2543, opened 2023-05-10
Pin `dirs-sys` to v0.4.0

lopopolo:
`dirs-sys` v0.4.1 brings in an additional dependency, `option-ext`. `option-ext` is licensed with MPL-2.0. MPL-2.0 is a vaguely copyleft license, all of which are banned in Artichoke by `cargo-deny` and `deny.toml`.

See:

- https://github.com/dirs-dev/dirs-sys-rs/issues/21
- https://github.com/dirs-dev/dirs-sys-rs/commit/e169da7af901eb621e5d244efe960f4da8ed150d

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,6 +77,7 @@ dependencies = [
  "bstr",
  "clap",
  "directories",
+ "dirs-sys",
  "log",
  "posix-space",
  "rustyline",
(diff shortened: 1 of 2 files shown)


pasabanov:
> all of which are banned in Artichoke

Why do you ban copyleft licenses?

I understand why someone would ban GPL family licenses due to their harsh requirements, but I don't understand why one would ban the MPL license, which essentially imposes the same amount of restrictions as Apache 2.0 or MIT unless you modify the code, in which case you just need to publish the modified version under the MPL.

These are very soft requirements, so why won't you accept them?

PR #2554, opened 2023-05-17
Followups to #2450, fix some bugs in `String::index` and `String::rindex`

lopopolo:
Followup to #2450 

Add additional tests which uncover some underflows and overflows.

Reimplement `Utf8String::index` and `Utf8String::rindex` to use the `find` and `rfind` routines in `bstr`. The new implementation does not allocate.

cc @choznerol


choznerol:
Thank you for the detailed commits & tests. I've learned a lot from the concise new implementation!

lopopolo:
Thanks for the ping here @choznerol! This prompted me to take another look and I think I've found some further improvements in `Utf8String`. I'd appreciate your review on https://github.com/artichoke/artichoke/pull/2566 if you have a moment.

choznerol:
Sure! Will take a look.

PR #2566, opened 2023-05-21
Significantly simplify and improve performance of Utf8String::index

lopopolo:
Upon review of the work @choznerol and @lopopolo did, I noticed that the structure of `Utf8String::index` could mirror that of the ASCII and binary implementations.

The ASCII and binary implementations did not look applicable at first pass because they did not have the initial step to compute the offset char length (they skip this because all characters are 1 byte).

Using `Utf8String::get_char_slice` though, we can convert the offset char count into a byte position, use `BStr::find` as we previously were, and then use the newly extracted `char_len` function on the matched portion of the string to find the char equivalent of that index. Converting the provided offset and found index to "char counts" allows the same basic structure of the routine to remain the same.

This is significantly cheaper because it allows us to delegate most of the char counting work to `bytecount`, which is SIMD optimized and should be much faster than doing a partial decode one character at a time with `bstr`.

### DIFF
--- a/spinoso-string/src/enc/utf8/mod.rs
+++ b/spinoso-string/src/enc/utf8/mod.rs
@@ -839,69 +839,11 @@ impl Utf8String {
             return None;
         }
 
-        let buf = self.get_char_slice(offset..usize::MAX)?;
-        let index = buf.find(needle)?;
+        let prefix = self.get_char_slice(0..offset)?;
+        let tail = &self[prefix.len()..];
+        let index = tail.find(needle)?;
 
-        let mut s = self.as_slice();
-        let mut char_count = 0;
-
-        // first decode `offset` number of characters.
-        let mut offset_chars_remaining = offset;
-        loop {
-            if offset_chars_remaining == 0 {
-                break;
-            }
-            match bstr::decode_utf8(s) {
-                (Some(_), size) => {
-                    s = &s[size..];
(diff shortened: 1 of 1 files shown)


choznerol (approved):
Thanks for the detailed explanation. It becomes so obvious after a second thought. I also didn't know that the char counting is `SIMD` optimized.

lopopolo:
> I also didn't know that the char counting is `SIMD` optimized.

yup! It's uses SIMD in 4 places:

https://github.com/artichoke/artichoke/blob/1ad29975f9fdd9f4b6c815a968a6a12690666e60/spinoso-string/src/enc/utf8/mod.rs#L146-L164

The string searching routines in `bstr` are generally SIMD optimzed `find_non_ascii_byte` on line 147 is one of those routines. This searches for the first non ASCII byte (if we don't find any, the char count is just the byte length).

Once a non-ASCII byte is found, we either are dealing with a tail that is fully valid UTF-8, in which case line 155 calls a SIMD-optimized UTF-8 validation routine `simdutf8::basic::from_utf8` and delegates to `bytecount` (a SIMD optimized UTF-8 character counter) on line 156.

If we have some malformed bytes in the string, we fall back to `bstr`'s `Utf8Chunks` iterator which allows us to again ru `bytecount` on the runs of valid UTF-8 characters.

PR #2571, opened 2023-05-23
Extract load path builders from `mezzaluna-feature-loader` to a new crate

lopopolo:
Split out load path builders and manipulation to a new crate: `mezzaluna-load-path`.

Remove the old code from `mezzaluna-feature-loader` and bump to v0.8.0.

The new crate is 100% documented and has more tests than the old code.

The new split setup moves in a direction of not conflating these 3 things together:

- Tracking loaded features
- Loading source code from disk
- Determining the load path

Eventually, `mezzaluna-feature-loader` will only track loaded features (and probably be renamed to `mezzaluna-loaded-features` along the way).

Splitting things up this way will make the conditional compilation a lot simpler and less NxM combinatorial explosion.


lopopolo:
markdown link check failure is expected. docs aren't published for the new crate yet:

```
  ERROR: 1 dead links found!
  [✖] https://artichoke.github.io/artichoke/mezzaluna_load_path/ → Status: 404
```

PR #2577, opened 2023-05-29
Add support for Regexp needles to String#byteindex

### LINKED ISSUE #2359: Add support for `Regexp` needles to `String#byteindex`
https://github.com/artichoke/artichoke/pull/2358 added the [`String#byteindex`] API but did not add support for `Regexp` patterns, like:

[`String#byteindex`]: https://ruby-doc.org/3.2.0/String.html#method-i-byteindex

```ruby
'foo'.byteindex(/f/) # => 0
'foo'.byteindex(/o/) # => 1
'foo'.byteindex(/oo/) # => 1
'foo'.byteindex(/ooo/) # => nil
```

I am not sure what the behavior of this function is when given a regexp and an offset arg.

The current code raises `NotImplementedError`. This is where the implementation must go:

https://github.com/artichoke/artichoke/blob/8460b3faef4c344dcf46f58086595e3f2b27660f/artichoke-backend/src/extn/core/string/trampoline.rs#L588-L591

See also:

- https://github.com/artichoke/artichoke/issues/2362

do-you-dare:
Part of https://github.com/artichoke/artichoke/issues/2359. This is my first time working with rust and artichoke, so all feedback is welcome.

Fixes https://github.com/artichoke/artichoke/issues/2359.

### DIFF
--- a/artichoke-backend/src/extn/core/string/trampoline.rs
+++ b/artichoke-backend/src/extn/core/string/trampoline.rs
@@ -585,23 +585,26 @@ pub fn byteindex(
     mut substring: Value,
     offset: Option<Value>,
 ) -> Result<Value, Error> {
-    #[cfg(feature = "core-regexp")]
     let s = unsafe { super::String::unbox_from_value(&mut value, interp)? };
+    #[cfg(feature = "core-regexp")]
     if let Ok(pattern) = unsafe { Regexp::unbox_from_value(&mut substring, interp) } {
-        let offset_number = match implicitly_convert_to_int(interp, offset.into()) {
-            Err(_) => None,
-            Ok(value) => Some(value),
+        let offset_number = if let Some(offset) = offset {
+            Some(implicitly_convert_to_int(interp, offset)?)
+        } else {
+            None
         };
 
+        // offset_number is a byte offset into the search string, not an offset into the capture
+        // group used. This is confirmed by playing around with MRI irb.
         let mut matches = pattern.match_(interp, Some(s.as_slice()), offset_number, None)?;
         if matches.is_nil() {
             return Ok(Value::nil());
         }
         let first_ocurrence = unsafe { MatchData::unbox_from_value(&mut matches, interp)? };
-        let ocurrence_index = first_ocurrence.begin(matchdata::Capture::GroupIndex(0)).unwrap();
+        let ocurrence_index = first_ocurrence.begin(matchdata::Capture::GroupIndex(0))?;
         match ocurrence_index {
+            Some(n) => return interp.try_convert(n),
             None => return Ok(Value::nil()),
-            Some(n) => return Ok(interp.convert(n as i64)),
         }
     }
     let needle = unsafe { implicitly_convert_to_string(interp, &mut substring)? };
--- a/artichoke-backend/src/extn/core/string/trampoline.rs
+++ b/artichoke-backend/src/extn/core/string/trampoline.rs
@@ -586,10 +586,24 @@ pub fn byteindex(
     offset: Option<Value>,
 ) -> Result<Value, Error> {
     #[cfg(feature = "core-regexp")]
-    if let Ok(_pattern) = unsafe { Regexp::unbox_from_value(&mut substring, interp) } {
-        return Err(NotImplementedError::from("String#byteindex with Regexp pattern").into());
-    }
     let s = unsafe { super::String::unbox_from_value(&mut value, interp)? };
+    if let Ok(pattern) = unsafe { Regexp::unbox_from_value(&mut substring, interp) } {
+        let offset_number = match implicitly_convert_to_int(interp, offset.into()) {
+            Err(_) => None,
+            Ok(value) => Some(value),
+        };
+
+        let mut matches = pattern.match_(interp, Some(s.as_slice()), offset_number, None)?;
+        if matches.is_nil() {
+            return Ok(Value::nil());
+        }
+        let first_ocurrence = unsafe { MatchData::unbox_from_value(&mut matches, interp)? };
+        let ocurrence_index = first_ocurrence.begin(matchdata::Capture::GroupIndex(0)).unwrap();
+        match ocurrence_index {
+            None => return Ok(Value::nil()),
+            Some(n) => return Ok(interp.convert(n as i64)),
+        }
+    }
     let needle = unsafe { implicitly_convert_to_string(interp, &mut substring)? };
     let offset = if let Some(offset) = offset {
         let offset = implicitly_convert_to_int(interp, offset)?;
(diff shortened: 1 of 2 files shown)


artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -586,10 +586,24 @@ pub fn byteindex(
...
     #[cfg(feature = "core-regexp")]
-    if let Ok(_pattern) = unsafe { Regexp::unbox_from_value(&mut substring, interp) } {
-        return Err(NotImplementedError::from("String#byteindex with Regexp pattern").into());
-    }
     let s = unsafe { super::String::unbox_from_value(&mut value, interp)? };
+    if let Ok(pattern) = unsafe { Regexp::unbox_from_value(&mut substring, interp) } {
+        let offset_number = match implicitly_convert_to_int(interp, offset.into()) {
+            Err(_) => None,
+            Ok(value) => Some(value),
+        };
```
lopopolo: calling `into()` on an `Option<Value>` flattens the option by turning the `None` case into a `nil` Ruby value. Further converting the `Err` return from `implicitly_convert_to_int` squashes this error handling and other errors that may be encountered in the implicit conversion (which can route back through Ruby code to call the `#to_int` method).

This statement doesn't properly handle these error conditions:

```console
$ irb
[3.2.2] > "foo".byteindex(/o/, nil)
(irb):1:in `byteindex': no implicit conversion from nil to integer (TypeError)
        from (irb):1:in `<main>'
        from /usr/local/var/rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'
        from /usr/local/var/rbenv/versions/3.2.2/bin/irb:25:in `load'
        from /usr/local/var/rbenv/versions/3.2.2/bin/irb:25:in `<main>'
[3.2.2] > class A; def to_int; 'abc' end; end
=> :to_int
[3.2.2] > "foo".byteindex(/o/, A.new)
(irb):3:in `byteindex': can't convert A to Integer (A#to_int gives String) (TypeError)
        from (irb):3:in `<main>'
        from /usr/local/var/rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'
        from /usr/local/var/rbenv/versions/3.2.2/bin/irb:25:in `load'
        from /usr/local/var/rbenv/versions/3.2.2/bin/irb:25:in `<main>'
[3.2.2] > class B; def to_int; raise 'oh no' end; end
=> :to_int
[3.2.2] > "foo".byteindex(/o/, B.new)
(irb):4:in `to_int': oh no (RuntimeError)
        from (irb):5:in `byteindex'
        from (irb):5:in `<main>'
        from /usr/local/var/rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'
        from /usr/local/var/rbenv/versions/3.2.2/bin/irb:25:in `load'
        from /usr/local/var/rbenv/versions/3.2.2/bin/irb:25:in `<main>'
```

When calling `implicitly_convert_to_int`, we must use the `?` operator to return errors.

What do you think about replacing this statement with:

```suggestion
        let offset_number = if let Some(offset) = offset {
            Some(implicitly_convert_to_int(interp, offset)?)
        } else {
            None
        };
```

artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -586,10 +586,24 @@ pub fn byteindex(
...
+
+        let mut matches = pattern.match_(interp, Some(s.as_slice()), offset_number, None)?;
+        if matches.is_nil() {
+            return Ok(Value::nil());
+        }
+        let first_ocurrence = unsafe { MatchData::unbox_from_value(&mut matches, interp)? };
+        let ocurrence_index = first_ocurrence.begin(matchdata::Capture::GroupIndex(0)).unwrap();
```
lopopolo: I was a bit confused here between line 596 and 601 on what `offset_number` meant. From playing around with this API (and what you've confirmed with the integration tests), `offset_number` is a byte offset into the "haystack" (string to search). It is NOT the capturing group to use.

This example made this clear to me:

```console
[3.2.2] > s = "abcdef"
=> "abcdef"
[3.2.2] > s.byteindex(/(c).*(f)/, 2)
=> 2
```

Could you add some comments to this code and the above case to `string_test.rb` to document this bit of the API?

lopopolo: Please don't unwrap here. I know this is very much "can never panic" code, but the error types are compatible and it removes this "dangerous" call from the source.

```suggestion
        let ocurrence_index = first_ocurrence.begin(matchdata::Capture::GroupIndex(0))?;
```

artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -586,10 +586,24 @@ pub fn byteindex(
...
+        match ocurrence_index {
+            None => return Ok(Value::nil()),
+            Some(n) => return Ok(interp.convert(n as i64)),
```
lopopolo: There are converters for most numeric types. We can avoid this `as` cast by using a fallible conversion.

```suggestion
            Some(n) => return interp.try_convert(n),
```

lopopolo: this is a style nit, but I prefer the `Some` arms to come first in a match like this.

artichoke-backend/src/extn/core/string/trampoline.rs
```diff
@@ -586,10 +586,24 @@ pub fn byteindex(
...
+        let offset_number = match implicitly_convert_to_int(interp, offset.into()) {
+            Err(_) => None,
+            Ok(value) => Some(value),
+        };
+
+        let mut matches = pattern.match_(interp, Some(s.as_slice()), offset_number, None)?;
+        if matches.is_nil() {
+            return Ok(Value::nil());
+        }
+        let first_ocurrence = unsafe { MatchData::unbox_from_value(&mut matches, interp)? };
```
lopopolo, line 603: this is super cool how you've delegated through to the `Regexp` and `MatchData` APIs to implement this!

lopopolo (changes requested):
This is looking great! I have a few suggestions for changes that I'd like to see addressed before we can merge this.

artichoke-backend/src/extn/core/string/trampoline.rs
do-you-dare: Thanks! I've added fixes for all points in 24e032e35764a16d8e6d657d786d79b919e4be38, I can squash the commits later if you prefer. Thanks for the feedback, it was very valuable to me!

lopopolo (approved):
this looks great! thanks for contributing 🙇 💎 🦀

Issue #183, opened 2019-08-20
Implement Encoding support


lopopolo:
Add support for [`Encoding`](https://ruby-doc.org/core-2.6.3/Encoding.html) class in Ruby core.

In Ruby, all `String`s have an encoding, which determines how character-oriented `String` methods behave. `Encoding` also enables transforming the raw byte contents of a `String` to another encoding.

One crate that we might use to implement this feature is [`encoding_rs`](https://docs.rs/encoding_rs/0.8.17/encoding_rs/).

Knowing a `String`'s encoding is required for GH-136.

For the initial implementation, let's add encoders for `Encoding::ASCII`, `Encoding::BINARY`, and `Encoding::UTF8`.

lopopolo:
Ruby has ~20 built in encodings that come from Onigmo:

https://github.com/ruby/ruby/blob/v3_0_2/include/ruby/onigmo.h#L184-L223

There are many aliases defined, for example:

https://github.com/ruby/ruby/blob/d92f09a5eea009fa28cd046e9d0eb698e3d94c5c/enc/x_emoji.h

PR #2578, opened 2023-05-31
Isolate `Encoding` logic to be extracted and enhanced upon

b-n:
This will likely be the first in a series of PRs.

I've gone this particular way of extracting `Encoding` to be in line with how the other `extn/core` types are implemented.

I haven't yet got plans on how this class/type will be backed, I firstly wanted to extract and isolate it (mostly to make sure it truly is isolated), and then iterate by moving the `encoding.rb` to `mruby.rs` piece by piece.

Relates: #183


b-n:
Some thoughts I'm having are around the linkage between `spinoso-string` and this particular piece of work. E.g. `spinoso-string` has [it's own `Encoding` enum](https://github.com/artichoke/artichoke/blob/trunk/spinoso-string/src/encoding.rs), and it has enough functionality to provide a number of functions that would be needed on the core `Encoding` type.

I think conceptually we could go crazy and extract a `spinoso-encoding` which is more of a database of encoding than actual implementation (that would be responsibility of spinoso-string still IMO) - but less is more, but just thinking out loud about where this could possibly go if we find the need one day.

b-n:
@lopopolo do you have a pref to bundle this into #2579, or to keep it separate?

lopopolo (approved):
Looks good to me.

I do kinda like the idea of a `spinoso-encoding` crate that is a database of _all_ encodings that MRI supports/Artichoke could support. And then it'd be up to `spinoso-string` to actually build in that support (and fallible conversions from `spinoso_encoding::Encoding` to `spinoso_string::Encoding` would facilitate that).

This would also get us to a point where one real enum would be able to be used for defining all of the encoding constants in the `Encoding` class.

lopopolo:
> @lopopolo do you have a pref to bundle this into #2579, or to keep it separate?

let's merge this separately. this is a nice refactor.

lopopolo:
> I haven't yet got plans on how this class/type will be backed

if we add a new `MRB_TT_ENCODING` type tag to mruby, the encodings can be converted to a `u32` and stuffed in the `sym` field of the `mrb_value`, which would let us pass around encodings without dynamic allocation (and allow us to pack the encoding into the `flags` of `String` `mrb_value`s like is currently done).

artichoke-backend/src/extn/core/encoding/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[cfg(test)]
+mod tests {
+    use crate::test::prelude::*;
+
+    const SUBJECT: &str = "Encoding";
+    const FUNCTIONAL_TEST: &[u8] = include_bytes!("encoding_test.rb");
+
+    #[test]
+    fn functional() {
```
lopopolo, line 26: I do love this pattern and love when new functional test suites get added. thanks for adding this.

b-n:
> I do kinda like the idea of a spinoso-encoding crate that is a database of all encodings that MRI supports/Artichoke could support. And then it'd be up to spinoso-string to actually build in that support (and fallible conversions from spinoso_encoding::Encoding to spinoso_string::Encoding would facilitate that).

You've read my mind a bit here. I was going towards this and then spotted the Encoding being available in spinoso-striing, and thought: "Well I might aswell begin there" haha.

PR #2579, opened 2023-05-31
[WIP] Implement `Encoding` in rust

b-n:
By extracting out `Encoding` into it's own module, we can start to rely on some of the implementations in `spinoso-string` to provide the functionality.

Some rough todos:

- [ ] Implement instance methods
- [ ] Implement constants and their binding to the interpreter

Relates: PR #2578
Relates: #183


artichoke-backend/src/convert.rs
```diff
@@ -13,7 +13,7 @@ use crate::Artichoke;
 
 mod array;
 mod boolean;
-mod boxing;
+pub(crate) mod boxing;
```
lopopolo: this shouldn't need to be made pub. all of the types are `pub use`'d below

artichoke-backend/src/encoding_registry.rs
```diff
@@ -0,0 +1,160 @@
...
+    type Value = Value;
+    type Error = Error;
+    type Spec = encoding::Spec;
+
+    fn def_encoding(&mut self, spec: Self::Spec) -> Result<(), Error> {
+        // Setup the class name regex
+        static ESCAPABLE_CONSTANT_CHARS: OnceLock<Regex> = OnceLock::new();
+        let escapable_const_char_regex = ESCAPABLE_CONSTANT_CHARS.get_or_init(|| {
+            // Match any `-` and `.` chars. These will be replaced with `_`
+            Regex::new(r"[-\.]+").unwrap()
```
lopopolo: regex is really heavy here. We should be able to use `ByteSlice::replace` from `bstr`: https://docs.rs/bstr/latest/bstr/trait.ByteSlice.html#method.replace

artichoke-backend/src/extn/core/encoding/trampoline.rs
```diff
@@ -0,0 +1,116 @@
...
+    let _ = interp;
+    Err(NotImplementedError::new().into())
+}
+
+pub fn name_list(interp: &mut Artichoke) -> Result<Value, Error> {
+    let _ = interp;
+    Err(NotImplementedError::new().into())
+}
+
+#[allow(clippy::unnecessary_wraps)]
```
lopopolo: this allow appears to be unused

artichoke-backend/src/state/mod.rs
```diff
@@ -32,6 +34,7 @@ pub struct State {
     pub parser: Option<parser::State>,
     pub classes: Registry<class::Spec>,
     pub modules: Registry<module::Spec>,
+    pub encodings: HashSet<encoding::Spec>,
```
lopopolo: I think a better design is to pass encodings around as symbols (i.e. a `u32`) both in the interpreter storage, `mrb_value` repr for `Encoding`s, and the flags of a `String` `mrb_value`.

If we end up doing that, we should use an `FxHashSet` here from `rustc_hash` which will perform much better

lopopolo (commented):
some early thoughts

artichoke-backend/src/encoding_registry.rs
b-n: Ah nice, that is indeed much better.

I was thinking of whether there is another way to create bare symbols. e.g.

```ruby
irb(main):001:0> constants = Encoding.constants
=> 
[:ISO8859_14,                                                         
...                                                                   
irb(main):002:0> Encoding::UTF_8.replicate("Super Special Encoding")
=> #<Encoding:Super Special Encoding>
irb(main):003:0> Encoding.constants - constants
=> [:Super_Special_Encoding, :SUPER_SPECIAL_ENCODING]
irb(main):004:0> Encoding::Super_Special_Encoding
=> #<Encoding:Super Special Encoding>
irb(main):005:0> Encoding::Super_Special_Encoding.name
=> "Super Special Encoding"
```

And check this out:

```ruby
irb(main):007:0> Encoding::UTF_8.replicate("\x7F")
=> #<Encoding:>
irb(main):008:0> Encoding::UTF_8.replicate("\x80")
(Object doesn't support #inspect)
=>                                                                                
irb(main):009:0> _.name
=> "\x80"
irb(main):010:0> _.encoding
=> #<Encoding:US-ASCII>
irb(main):011:0> Encoding.find("\x80")
(Object doesn't support #inspect)
=>                                                               
irb(main):012:0> _.name
=> "\x80"
```

Valid names apparently need to be asciicompat as per https://github.com/artichoke/artichoke/blob/trunk/artichoke-backend/vendor/ruby/include/ruby/internal/encoding/encoding.h#L766-L793 which is used here: https://github.com/artichoke/artichoke/blob/trunk/artichoke-backend/vendor/ruby/encoding.c#L288-L301

Eventually some more to look into maybe, but:

```ruby
irb(main):001:0> constants = Encoding.constants
=> 
[:ISO8859_14,                                                          
...                                                                    
irb(main):002:0> (1..127).map(&:chr).each_slice(32).map { |c| Encoding::UTF_8.replicate(c.join) }
=> 
[#<Encoding:                                                           
                                                                       
                                                                       
 >,                                                                    
 #<Encoding:!"#$%&'()*+,-./0123456789:;<=>?@>,                         
 #<Encoding:ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`>,                         
 #<Encoding:abcdefghijklmnopqrstuvwxyz{|}~>]                           
irb(main):003:0> Encoding.constants - constants
=> [:ABCDEFGHIJKLMNOPQRSTUVWXYZ______, :ABCDEFGHIJKLMNOPQRSTUVWXYZ_____]
```

(Had to chunk it because otherwise some issue with Constants and lengths that I think it was getting mad at).

artichoke-backend/src/extn/core/encoding/trampoline.rs
b-n: Yeah, heavy WIP. If it looks dirty, it'll be cleaned prior to requesting real review :smile:

artichoke-backend/src/encoding_registry.rs
b-n: Note: in the between time, going to take the easy Bstr replace approach.

Issue #1726, opened 2022-03-07
Implement `String::force_encoding` and `EncodedString::force_encoding`


lopopolo:
> ruby `force_encoding` is gong to be .... fun

`String#force_encoding` can be implemented like this:

```rust
impl EncodedString {
    pub fn force_encoding(&mut self, encoding: Encoding) {
        if encoding == self.encoding() {
            return;
        }
        let buf = mem::replace(self.as_mut_vec(), vec![]);
        match encoding {
            Encoding::Ascii => *self = Self::Ascii(AsciiString::new(buf)),
            Encoding::Ascii => *self = Self::Binary(BinaryString::new(buf)),
            Encoding::Ascii => *self = Self::Utf8(Utf8String::new(buf)),
        }
    }
}
```

_Originally posted by @lopopolo in https://github.com/artichoke/artichoke/pull/1678#discussion_r803376737_

PR #2637, opened 2023-08-05
Push `codepoints()` implementation down into encoded variants

### LINKED ISSUE #1728: Implement a `Codepoints` iterator for each encoded string variant in `spinoso-string`
Push the branching down into the encoded variants.

Followup to https://github.com/artichoke/artichoke/pull/1678.

choznerol:
Fix #1728

This PR moves the encoding-specific code from `codepoints.rs` to encoded variants.

- The original version [consumes](https://github.com/artichoke/artichoke/compare/choznerol/1728-Implement-a-Codepoints-iterator-for-each-encoded-string-variant-in-spinoso-string?expand=1#diff-a2d2fa818d5b61fb277dd19f5a3a58b696944224b071fe8faee5bca3c19103d3L1923) the `String` to produce a `Codepoints` iterator, while the new version doesn't consume the `String` but allocate a `Vec` to create iterator. I'm not sure if the previous approach is preferred, if so I can try to refactor it.
- I made 589977d and 24837dd separate commits, as they may or may not be the preferred approach.

### DIFF
--- a/spinoso-string/src/enc/ascii/mod.rs
+++ b/spinoso-string/src/enc/ascii/mod.rs
@@ -11,6 +11,7 @@ use crate::codepoints::{Codepoints, InvalidCodepointError};
 use crate::encoding::Encoding;
 use crate::iter::{Bytes, IntoIter, Iter, IterMut};
 use crate::ord::OrdError;
+use crate::CodepointsError;
 
 mod eq;
 mod impls;
@@ -283,15 +284,15 @@ impl AsciiString {
     }
 
     #[inline]
-    pub fn codepoints(&self) -> Codepoints {
+    pub fn codepoints(&self) -> Result<Codepoints, CodepointsError> {
         let iter = self
             .inner
             .as_slice()
             .bytes()
-            .map(u32::from)
+            .map(|b| b as u32)
             .collect::<Vec<_>>()
             .into_iter();
-        Codepoints::from(iter)
+        Ok(Codepoints::from(iter))
     }
 }
 
--- a/spinoso-string/src/enc/ascii/mod.rs
+++ b/spinoso-string/src/enc/ascii/mod.rs
@@ -1,5 +1,4 @@
 use alloc::collections::TryReserveError;
-use alloc::vec::Vec;
 use core::fmt;
 use core::ops::Range;
 use core::slice::SliceIndex;
@@ -7,11 +6,10 @@ use core::slice::SliceIndex;
 use bstr::ByteSlice;
 use scolapasta_strbuf::Buf;
 
-use crate::codepoints::{Codepoints, InvalidCodepointError};
+use crate::codepoints::InvalidCodepointError;
 use crate::encoding::Encoding;
 use crate::iter::{Bytes, IntoIter, Iter, IterMut};
 use crate::ord::OrdError;
-use crate::CodepointsError;
 
 mod eq;
 mod impls;
@@ -281,18 +279,6 @@ impl AsciiString {
     #[inline]
     pub fn extend_from_slice(&mut self, other: &[u8]) {
         self.inner.extend_from_slice(other);
-    }
-
-    #[inline]
-    pub fn codepoints(&self) -> Result<Codepoints, CodepointsError> {
-        let iter = self
-            .inner
-            .as_slice()
-            .bytes()
-            .map(|b| b as u32)
-            .collect::<Vec<_>>()
-            .into_iter();
-        Ok(Codepoints::from(iter))
     }
 }
 
--- a/spinoso-string/src/enc/ascii/mod.rs
+++ b/spinoso-string/src/enc/ascii/mod.rs
@@ -288,7 +288,7 @@ impl AsciiString {
             .inner
             .as_slice()
             .bytes()
-            .map(u32::from)
+            .map(char::from)
             .collect::<Vec<_>>()
             .into_iter();
         Codepoints::from(iter)
--- a/spinoso-string/src/enc/ascii/mod.rs
+++ b/spinoso-string/src/enc/ascii/mod.rs
@@ -11,7 +11,6 @@ use crate::codepoints::{Codepoints, InvalidCodepointError};
 use crate::encoding::Encoding;
 use crate::iter::{Bytes, IntoIter, Iter, IterMut};
 use crate::ord::OrdError;
-use crate::CodepointsError;
 
 mod eq;
 mod impls;
@@ -284,15 +283,15 @@ impl AsciiString {
     }
 
     #[inline]
-    pub fn codepoints(&self) -> Result<Codepoints, CodepointsError> {
+    pub fn codepoints(&self) -> Codepoints {
         let iter = self
             .inner
             .as_slice()
             .bytes()
-            .map(|b| b as u32)
+            .map(u32::from)
             .collect::<Vec<_>>()
             .into_iter();
-        Ok(Codepoints::from(iter))
+        Codepoints::from(iter)
     }
 }
 
--- a/spinoso-string/src/enc/ascii/mod.rs
+++ b/spinoso-string/src/enc/ascii/mod.rs
@@ -288,7 +288,7 @@ impl AsciiString {
             .inner
             .as_slice()
             .bytes()
-            .map(char::from)
+            .map(u32::from)
             .collect::<Vec<_>>()
             .into_iter();
         Codepoints::from(iter)
--- a/spinoso-string/src/enc/ascii/mod.rs
+++ b/spinoso-string/src/enc/ascii/mod.rs
@@ -11,12 +11,14 @@ use crate::encoding::Encoding;
 use crate::iter::{Bytes, IntoIter, Iter, IterMut};
 use crate::ord::OrdError;
 
+mod codepoints;
 mod eq;
 mod impls;
 mod inspect;
 #[cfg(feature = "std")]
 mod io;
 
+pub use codepoints::Codepoints;
 pub use inspect::Inspect;
 
 #[repr(transparent)]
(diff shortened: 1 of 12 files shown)


spinoso-string/src/enc/ascii/mod.rs
```diff
@@ -280,6 +281,18 @@ impl AsciiString {
...
+    pub fn codepoints(&self) -> Codepoints {
+        let iter = self
+            .inner
+            .as_slice()
+            .bytes()
+            .map(char::from)
+            .collect::<Vec<_>>()
+            .into_iter();
+        Codepoints::from(iter)
+    }
```
lopopolo: usually I'd implement this by having each encoding have their own ` CodePoints` iterator.

Then the codepoints iterator for string would be defined like this:

```rust
pub struct Codepoints<'a> {
    state: State<'a>,
}

enum State<'a> {
    Ascii(ascii::Codepoints<'a>),
    Binary(binary::Codepoints<'a>),
    Utf8(utf8::Codepoints<'a>),
}
```

Take a look at how `Inspect` is implemented:

https://github.com/artichoke/artichoke/blob/0ae1fb15dee52dc736d05302804625a4d4c8819c/spinoso-string/src/enc/inspect.rs#L5-L35

This avoids the allocation to normalize everything to `IntoIter`.

spinoso-string/src/enc/ascii/mod.rs
```diff
@@ -280,6 +281,18 @@ impl AsciiString {
...
         self.inner.extend_from_slice(other);
     }
+
+    #[inline]
+    pub fn codepoints(&self) -> Codepoints {
+        let iter = self
+            .inner
+            .as_slice()
+            .bytes()
+            .map(char::from)
```
lopopolo: I'm not sure it is correct to call `char::from`. What is the behavior when the encoding is not valid?

```console
$ irb
[3.2.2] > s = "\0\xFF"
=> "\u0000\xFF"
[3.2.2] > s.codepoints
(irb):2:in `codepoints': invalid byte sequence in UTF-8 (ArgumentError)
        from (irb):2:in `<main>'
        from /usr/local/var/rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'
        from /usr/local/var/rbenv/versions/3.2.2/bin/irb:25:in `load'
        from /usr/local/var/rbenv/versions/3.2.2/bin/irb:25:in `<main>'
[3.2.2] > s.dup.b.codepoints
=> [0, 255]
[3.2.2] > s.dup.force_encoding('ascii').codepoints
=> [0, 255]
```

spinoso-string/src/enc/utf8/owned.rs
```diff
@@ -146,6 +147,16 @@ impl Utf8String {
         self.try_push_codepoint(int)
     }
 
+    #[inline]
+    pub fn codepoints(&self) -> Result<Codepoints, CodepointsError> {
```
lopopolo: this should be implemented on the UTF-8 borrowed type. It does not need to mutate the backing buffer.

spinoso-string/src/enc/utf8/owned.rs
```diff
@@ -146,6 +147,16 @@ impl Utf8String {
...
+        if let Ok(s) = self.inner.as_slice().to_str() {
```
lopopolo: this should use `self.is_valid_encoding()` which is SIMD-optimized.

spinoso-string/src/lib.rs
```diff
@@ -1919,7 +1919,7 @@ impl String {
     /// [binary encoded]: crate::Encoding::Binary
     /// [`str::from_utf8`]: core::str::from_utf8
     #[inline]
-    pub fn codepoints(&self) -> Result<Codepoints<'_>, CodepointsError> {
+    pub fn codepoints(&self) -> Result<Codepoints, CodepointsError> {
```
lopopolo: this should ideally not allocate.

lopopolo (changes requested):

choznerol:
Just found out I somehow missed the notification 😓  Will get back to this soon.

spinoso-string/src/lib.rs
choznerol: The b8729b7 version doesn't use `alloc::vec` anymore 👍

spinoso-string/src/enc/utf8/owned.rs
choznerol: I end up using a single `simdutf8::basic::from_utf8()` in b8729b7.

spinoso-string/src/enc/utf8/owned.rs
choznerol: I implemented this on `Utf8Str` instead in b8729b7.

spinoso-string/src/enc/ascii/mod.rs
choznerol: I reverted to use `u32::from` instead of `char::from`.

Also tested this for ASCII in e6be81d and asserted invalid bytes are just interpreted as bytes (e.g. `\xFF` -> `255`) instead of raising error:

https://github.com/artichoke/artichoke/blob/e6be81d1dbd798b8215f7d4031f83611b6ff9b8b/spinoso-string/src/enc/ascii/codepoints.rs#L52-L56

However, from `airb` an invalid UTF-8 code point error is still raised even with `force_encoding('ascii')`, as of e6be81d1dbd798b8215f7d4031f83611b6ff9b8b:
```console
$ cargo run --bin airb
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/airb`
artichoke 0.1.0-pre.0 (2023-10-20 revision 7021) [aarch64-apple-darwin]
[rustc 1.70.0 (90c541806 2023-05-31) on aarch64-apple-darwin]
>>> "hello".codepoints
=> [104, 101, 108, 108, 111]
>>> "hello💎".codepoints
=> [104, 101, 108, 108, 111, 128142]
>>>
>>> s = "\0\xFF"
=> "\x00\xFF"
>>> s.codepoints
	2: from (airb):1
	1: from (airb):1:in codepoints
ArgumentError (invalid byte sequence in UTF-8)
>>> s.dup.b.codepoints
=> [0, 255]
>>> s.dup.force_encoding('ascii').codepoints
	2: from (airb):1
	1: from (airb):1:in codepoints
ArgumentError (invalid byte sequence in UTF-8)
>>>
```

Since `map(u32::from)` is introduced before this PR, perhaps that's related to another issue #1726?

spinoso-string/src/enc/ascii/mod.rs
choznerol: Thanks! This was exactly what I tried to figure out but couldn't. I reimplement with this design in b8729b7.

spinoso-string/src/enc/utf8/borrowed/codepoints.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl<'a> TryFrom<&'a Utf8Str> for Codepoints<'a> {
+    type Error = CodepointsError;
+
+    #[inline]
+    fn try_from(s: &'a Utf8Str) -> Result<Self, Self::Error> {
+        match simdutf8::basic::from_utf8(s.as_bytes()) {
+            Ok(s) => Ok(Self { inner: s.chars() }),
+            Err(_) => Err(CodepointsError::invalid_utf8_codepoint()),
```
lopopolo, line 30: Can we add this in a comment as documentation?

```
[3.2.2] > s = "abc\xFFxyz"
=> "abc\xFFxyz"
[3.2.2] > s.encoding
=> #<Encoding:UTF-8>
[3.2.2] > s.codepoints
(irb):5:in `codepoints': invalid byte sequence in UTF-8 (ArgumentError)
        from (irb):5:in `<main>'
        from /usr/local/var/rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'
        from /usr/local/var/rbenv/versions/3.2.2/bin/irb:25:in `load'
        from /usr/local/var/rbenv/versions/3.2.2/bin/irb:25:in `<main>'
```

lopopolo (approved):

spinoso-string/src/enc/ascii/mod.rs
lopopolo: ahh yea, `String#force_encoding` is a stub right now, not properly implemented:

https://github.com/artichoke/artichoke/blob/f32bd63872b8b5b247c3bcb6516801767290057f/artichoke-backend/src/extn/core/string/string.rb#L519-L526

I think getting that fixed is part of @b-n's work on implementing `Encoding`

spinoso-string/src/enc/ascii/mod.rs
lopopolo: awesome! new revision looks great. I kinda love this pattern of handling the various encodings and Rust's lifetime support makes the code super easy to reason about (and elegant IMO).

spinoso-string/src/enc/utf8/borrowed/codepoints.rs
lopopolo: I pushed a followup commit with this: 131fd3dab4f2451e7393f626e088c189b61f9732

lopopolo:
So sorry it took me so long to get back to you on this @choznerol! Once CI is green this is going in. Thanks for addressing all of my feedback (and for the ping by re-requesting my review 😅)

choznerol:
> So sorry it took me so long to get back to you on this @choznerol! Once CI is green this is going in. Thanks for addressing all of my feedback (and for the ping by re-requesting my review 😅)

Not at all. Thank you for the detailed code review!

PR #2646, opened 2023-09-02
Upgrade Rust toolchain to Rust 1.72.0

lopopolo:


lopopolo:
docker nightly image upgrade is here:

- https://github.com/artichoke/docker-artichoke-nightly/pull/182

PR #2737, opened 2024-08-19
Skip fuzzing for cfg in rustdoc tests too

lopopolo:
Follow up to:

- #2736


lopopolo:
Both this PR and the previous emit warnings because latest nightly checks for un-used and undeclared cfg strings. I need to upgrade the rust toolchain in this repo to resolve

PR #2811, opened 2025-02-02
Prepare `artichoke-*` crates for Rust 2024

lopopolo:
All except `artichoke-backend`. Add missing SAFETY comments and align crates to use new set of crate-level pragmas.


lopopolo:
pulling in https://github.com/strawlab/iana-time-zone/pull/148 once it lands should get most of the builds passing.

PR #2818, opened 2025-02-03
Fix spelling errors

lopopolo:
The codebase now reports no errors with cargo-spellcheck. A lot of this change involved fixing up LLM-generated code comments.

As part of going through this, I found an old FIXME comment in artichoke-backend's state output constructors referring to a `bstr` limitation which was fixed in the 1.0.0 release, so I've included the const changes here.


scolapasta-hex/src/tests/mod.rs
```diff
@@ -61,28 +61,28 @@ fn test_literal_exhaustive() {
 
 #[test]
 fn test_escape_byte() {
-    // Expected hex escape codes for all possible u8 values
+    // Expected hex escape codes for all possible `u8` values
     #[rustfmt::skip]
-        const EXPECTED: [&str; 256] = [
```
lopopolo, line 66: this indentation mishap appears to be from a move of this test in a nested `mod tests { }` block to this top-level decl in the `tests/mod.rs` file.

PR #2819, opened 2025-02-03
Remove dep on `regex` from `spinoso-time`, enable local timezone

lopopolo:
With the [help of o3-mini-high](https://chatgpt.com/share/67a05f20-7dfc-8002-ab24-6d1b3594182c), rewrite the parser for `[+/-]HHMM` fixed timezone offsets to be a small hand-rolled parser instead of a regex + capture groups matcher.

While in here, I noticed that the `tzrs-local` feature wasn't wired up to artichoke, which meant `Time.now` always returned a GMT datetime.

With these two changes:

```console
$ cargo tree -p spinoso-time
spinoso-time v0.10.0 (/Users/lopopolo/dev/artichoke/artichoke/spinoso-time)
├── iana-time-zone v0.1.61
│   └── core-foundation-sys v0.8.7
├── strftime-ruby v1.1.0
├── tz-rs v0.7.0
└── tzdb v0.7.2
    ├── tz-rs v0.7.0
    └── tzdb_data v0.2.1
        └── tz-rs v0.7.0
```

```console
$ cargo run --bin airb -q
artichoke 0.1.0-pre.0 (2025-02-03 revision 7370) [aarch64-apple-darwin]
[rustc 1.83.0 (90b35a623 2024-11-26) on aarch64-apple-darwin]
>>> Time.now
=> 2025-02-02 22:17:54 -0800
```

cc @b-n I think we chatted about this as a TODO a long time ago.

### DIFF
--- a/spinoso-time/src/time/tzrs/offset.rs
+++ b/spinoso-time/src/time/tzrs/offset.rs
@@ -1,9 +1,7 @@
 use std::io::{self, Write as _};
 use std::slice;
 use std::str;
-use std::sync::LazyLock;
 
-use regex::Regex;
 use tz::timezone::{LocalTimeType, TimeZoneRef};
 use tzdb::time_zone::etc::GMT;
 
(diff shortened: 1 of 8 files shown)


spinoso-time/src/time/tzrs/offset.rs
```diff
@@ -392,6 +362,47 @@ impl TryFrom<i32> for Offset {
...
+        [b'+', _, _, _, _] => (1, &input[1..3], &input[3..5]),
+        // Matches `-hhmm`
+        [b'-', _, _, _, _] => (-1, &input[1..3], &input[3..5]),
+        // Matches `+hh:mm`
+        [b'+', _, _, b':', _, _] => (1, &input[1..3], &input[4..6]),
+        // Matches `-hh:mm`
+        [b'-', _, _, b':', _, _] => (-1, &input[1..3], &input[4..6]),
+        // Anything else is an error.
+        _ => return Err(TzStringError::new()),
+    };
```
b-n, line 393: This is absolutely beautiful way to solve this string parsing. I'm sure it's faster than regex, but I'm also curious how the compiler optimizes this out. (but also not motivated enough to write a benchmark :sweat_smile:)

PR #2849, opened 2025-05-03
readme: add some copy-paste joy

tshepang:


lopopolo (changes requested):
Happy to take this change but I think the language on the code blocks should be changed from `console` to `shell`

tshepang:
do you mean if you show a raw command it should be `shell`, and if it is a command with its output, it should be `console`

lopopolo:
Yes. The `$` is interpreted as a shell prompt with the `console` language.

lopopolo (approved):

